PR #130, opened 2015-01-07
added dirent.h

anshul1912:


anshul1912:
Added some more fixes to crash when language not found

PR #149, opened 2015-04-07
Prevent overwriting files

kisselef:
Terminating application if one of output filenames is same as one of input filenames. Since it looks like user has mistaken with input or output parameters, I decided that this error is fatal and there is no need to try to handle this situation in ccextractor (e.g. add something to the output name).


kisselef:
Sorry for "if"s' opening brace style :(

PR #154, opened 2015-04-16
"empty" PES packet for dvb subs streams now doesn't break the appflow

kisselef:
Since "empty" PES packets (00 00 01 BE 00 01 FF) in dvb-subtitles stream with length < 9 (7 to be exact) don't mean anything, just skipping them.


kisselef:
Seems that "empty" pes  is an encoder feature, and while there is nothing illegal, probably it is better to detect only 7-bytes PES here instead of all that have lenght < 9. At least broken pes with length < 7 could be detected. @cfsmp3, discard pull request if it matters (seems to me it does), I will implement only 7-bytes PES handling.

kisselef:
Here it is https://github.com/CCExtractor/ccextractor/pull/155

PR #155, opened 2015-04-17
handling 7-byte PES packets for dvb-subs stream

kisselef:
This handling is be better, because broken PES packets could happen too.


gfranck:
Thanks, your patch works perfectly except that it prints a lot of traces : "dvbsub_decode: incomplete, broken or empty packet"

PR #173, opened 2015-06-03
Case fix

mailumangjain:
Case fix.
Base sample : /John(lowercase)

### DIFF
--- a/src/lib_ccx/lib_ccx.h
+++ b/src/lib_ccx/lib_ccx.h
@@ -126,6 +126,7 @@ struct ccx_s_teletext_config {
 	// uint8_t se_mode : 1; // search engine compatible mode => Uses CCExtractor's write_format
 	// uint64_t utc_refvalue; // UTC referential value => Moved to ccx_decoders_common, so can be used for other decoders (608/xds) too
 	uint16_t user_page; // Page selected by user, which MIGHT be different to 'page' depending on autodetection stuff
+	int sentence_cap;
 };
 #define MAX_PID 65536
 struct lib_ccx_ctx
--- a/src/lib_ccx/params.c
+++ b/src/lib_ccx/params.c
@@ -1100,6 +1100,8 @@ void parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
 				strcmp (argv[i],"-sc")==0)
 		{
 			ccx_options.sentence_cap=1;
+			// Teletext needs to be set as well
+			tlt_config.sentence_cap = 1;
 			continue;
 		}
 		if ((strcmp (argv[i],"--capfile")==0 ||
--- a/src/lib_ccx/telxcc.c
+++ b/src/lib_ccx/telxcc.c
@@ -239,10 +239,10 @@ static uint64_t prev_show_timestamp;
 char cap_telx(char *a)
  {	static int flag=1;
 
-	if(*a == '\r')
- 		{flag = 1; }
+	if(*a == '\r' || *a == '?' || *a=='.' || *a == '!' || *a == ':')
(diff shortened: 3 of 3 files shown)


wforums:
I think that the case fixing should be optional, now it will always be applied. For regular captions we have the -sc (--sentencap) parameter. I'd propose to use this one (and add an option to the telxcc code to make use of it) to let the user decide if he wants the case formatting yes or no.

If this is not done, regression testing will need to be done.

wforums:
@mailumangjain: could you add support for the -sc parameter? You can borrow the code I have for it (see: https://github.com/wforums/ccextractor/commit/d539ee9109a8adc9df8b0c9ea1d9cd4c1e51b2ab) :smile:

mailumangjain:
Oh yes @wforums ! just give me some time, I'll do it as soon as possible and notify here ! Thanks for the help. :)

mailumangjain:
@wforums I have added -sc parameter for capitalization choice by user.

cfsmp3:
Why use a different config variable? Rather than adding a variable
in tlt_config just use the same one used in the rest of the code.

wforums:
the ccx options are not available in the tlt config (anymore), that's the reason it has a new variable in tlt_config.

wforums:
While the outdated version of @mailumangjain still might contain ccx_options, it is in our current git repo not longer the case. @anshul1912 refactored this a while ago, hence the need for the extra variable.

anshul1912:
something that is need to be done  is teletext should use encoder interface, if it use same encoder interface then there is no need of that variable, same variable can be used.
If someone is going for workaround then he dont need to add any new global variable, instead paas global variable to tlt_config

anshul1912:
I have passed cc_sub structure inside, using it telexcc no more need any local configuration for output

wforums:
@mailumangjain Would it be possible for you to close this request, update CCExtractor and make a new (unconflicting) PR? I tried a merge, but it seems as though a couple of regression files are affected by the change... http://gsocdev.ccextractor.org/~willem/users/willem/Testsuite_Report_2015-06-18_115417/Report__Teletext_2015-06-18_120038.html

mailumangjain:
ok. I was waiting for cc_sub to be passed.
Can anyone brief me about how regression testing/ @wforums testSuite works so that I can myself try and find if my solution is correct, in future ?

wforums:
Regression testing is at this moment only possible if you have an account on the development server, or if you can download the testing suite and all the samples. I'm planning to improve this, but in the meantime, I'll gladly run the test suite for you if you think you have a solution ready :)

mailumangjain:
Well, Thank you. Just being curious :)

On Thu, Jun 18, 2015 at 6:22 PM, Willem <notifications@github.com> wrote:

> Regression testing is at this moment only possible if you have an account
> on the development server, or if you can download the testing suite and all
> the samples. I'm planning to improve this, but in the meantime, I'll gladly
> run the test suite for you if you think you have a solution ready :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CCExtractor/ccextractor/pull/173#issuecomment-113146723>
> .
>



-- 
Umang Jain

anshul1912:
"I was waiting for cc_sub to be passed."
@mailumangjain  I have passed them already, git pull your repository

mailumangjain:
I added a "int sent_cap" in cc_sub struct for managing capitalization. I handled the cases if users wants capitalization.
My question is where do I set the variable value for "int sent_cap" when the user passes -sc parameter.
I am looking in params.c , but there is no cc_sub structure reference present there ? Am i doing something wrong ?

PR #180, opened 2015-06-10
flushing cc buffer every I, P slice for both fields

rkuchumov:
fixes espn.ts but affects /repository/newRepository/TestFiles/XDS/9.mpg which is already broken. (it seems that it extracts more data now http://gsocdev.ccextractor.org/~ruslan/tests/Testsuite_Report_2015-06-10_081012/Report__XDS_2015-06-10_081753.html )


wforums:
What about this? http://gsocdev.ccextractor.org/~ruslan/tests/Testsuite_Report_2015-06-10_081012/Report__MP4_2015-06-10_081446.html

Seems to break MP4 in it's entirety...

PR #192, opened 2015-06-26
Midterm: cctranslate integration + sharing service

kisselef:


src/lib_ccx/ccx_common_option.c
```diff
@@ -95,7 +95,13 @@ void init_options (struct ccx_s_options *options)
 	options->input_source=CCX_DS_FILE; // Files, stdin or network
 	options->demux_cfg.auto_stream = CCX_SM_AUTODETECT;
 	options->demux_cfg.m2ts = 0;
-
+#ifdef ENABLE_SHARING
+	options->sharing_enabled = 0;
+	options->sharing_url = NULL;
+    options->translate_enabled = 0;
```
anshul1912: need indentation here

src/lib_ccx/ccx_common_option.h
```diff
@@ -101,6 +101,15 @@ struct ccx_s_options // Options from user parameters
 	// Output structures
 	struct ccx_s_write wbout1;
 	struct ccx_s_write wbout2;
+#ifdef ENABLE_SHARING
+	//CC sharing
+	int sharing_enabled;
+	char *sharing_url;
+    //Translating
+    int translate_enabled;
```
anshul1912: ditto

src/lib_ccx/ccx_sub_entry.pb_c.h
```diff
@@ -0,0 +1,86 @@
+/* Generated by the protocol buffer compiler.  DO NOT EDIT! */
```
anshul1912: Auto generated files should be avoided to be placed in git

kisselef: @anshul1912 Of course, I agree in general, but it is not that case )

cfsmp3: Two questions

1) What does Willem's test suite say about this patch
2) Lots of changes in the code, but none in CHANGES.TXT :-) We need to take
that file seriously... I think each new version is missing new features
that for some reason we haven't documented anywhere.

kisselef: @cfsmp3,
1. It passes not all, but same tests as current master branch of ccextractor repo
2. Sure I will provide all documentation including changes. This is a "familiarization" branch for you all to try the feature. I've pulled a request only because you asked for it.

PR #211, opened 2015-08-18
support tesseract 3.04 library

anshul1912:


anshul1912:
@ccx-testerbot runtests

PR #214, opened 2015-08-20
Networking epg support

rkuchumov:
BIN data is now wrapped in protocol packets.
Added feature to send EPG events.
CCExtractor doesn't prompt password, users should use `-tcppassword` instead.
Connection protocol now requires keep-alive packets so that both sides can close connection when doesn't receive them


rkuchumov:
@ccx-testerbot runtests

PR #217, opened 2015-08-24
Add support for parsing the Service Description Table in MPEGTS 

brooss:
and using it to provide real channel names in EPG output.


brooss:
@ccx-testerbot runtests

brooss:
@ccx-testerbot runtests

PR #232, opened 2015-10-02
cea-708 fixed unhandled predefined window styles

kisselef:


kisselef:
@ccx-testerbot runtests

PR #240, opened 2015-10-21
correct Garbeled cc in multiple slice header

### LINKED ISSUE #235: Anagrammed words
CC extracting from this MPEG2-TS source (https://onedrive.live.com/redir?resid=9AE76B6175D960D3!381261&authkey=!ANmlBJmNWnzFyos&ithint=video%2cts) I get some anagrams instead of  the original source.
Which could be the cause?

Here what  I get currently:

2
00:00:02,336 --> 00:00:04,545
 It's too late for Dre.amamin   
 We're practillcathy ere.       

that should be instead:

2
00:00:02,336 --> 00:00:04,545
 It's too late for Dreamamin.   
 We're practically there.

anshul1912:
Also added some debug logs, while correcting the garbled logic
fixes #235


anshul1912:

anshul1912:

PR #292, opened 2016-03-06
Fixing warning errors for GCC while compilation

abhishek-vinjamoori:
Function definitons in output.c
Fixed in lib_ccx.h


abhishek-vinjamoori:

Issue #287, opened 2016-02-26
.79 crashes on small .bin files


Kapstew:
I've found what I believe to be bug in CCExtractorwin.exe version 0.79.   When I attempt to extract captioning from a file that’s from a 5-minute recording ccextractorwin crashes.  

Ccextractorwin version 0.77 does not crash when extracting text from this same file.  I’ve had no problems extracting text from a 30 to 60 minute recording with 0.79.

I’m using this command line:

ccextractorwin "C:\Source\WMTV_Today_6_955am.bin" -ttxt -o Output79.txt

I’ve attached the .bin (with .txt extension) file that I’ve been using to extract from and a text file from a successful output from version 0.77 that was run from this file.
[Output77.txt](https://github.com/CCExtractor/ccextractor/files/148412/Output77.txt)

[WMTV_Today_6_955am.bin.txt](https://github.com/CCExtractor/ccextractor/files/148410/WMTV_Today_6_955am.bin.txt)

abhibansal530:
This issue was due to fflush,which is fixed now. 
877ee7b2c7b4d2c660417f792257ed545839607f

canihavesomecoffee:
Can be closed once #323 is merged (after bot confirms that PR doesn't introduce new regressions). Tested the PR manually and it solves this issue.

Issue #281, opened 2016-02-20
File that makes CCExtractor crash


cfsmp3:
[1455449602_727724800_2.zip](https://github.com/CCExtractor/ccextractor/files/139080/1455449602_727724800_2.zip)
The attached file caused a really early crash.

vinayakathavale:
File did not cause a crash for me 
[output](http://pastebin.com/kj0vG2kB) on stdout 
[contents](http://pastebin.com/L33Sctqx) of the .srt file created

Arbitrer:
@vinayakathavale what Operating system , IDE do you use?

vinayakathavale:
OS: ubuntu 14.04 
Oh my bad Got the former output on installing  v0.79
I am getting a segmentation fault after installing the latest build

phanindra707:
CCextractor didnot crash in mine.It produced the subtitles
![stdout](https://cloud.githubusercontent.com/assets/9819126/13625679/1cb882be-e5e0-11e5-90f4-3d35b1e04de2.png) output for subtitles

But the problem is after compiling the ts file it is not playing in vlc before it played the video and subtitles are being generated for each and every compilation.

vinayakathavale:
@phanindra707  the issue has been fixed by the following commit https://github.com/CCExtractor/ccextractor/commit/877ee7b2c7b4d2c660417f792257ed545839607f

canihavesomecoffee:
@vinayakathavale Could you use the "Closes #281" in the PR next time? Then this issue will be closed automatically :)

vinayakathavale:
I didn't make the commit but will keep that in mind thanks :+1:

canihavesomecoffee:
Can be closed once #323 is merged. Tested the PR manually and it solves this issue.

Issue #277, opened 2016-02-07
.78 & .79 crash when processing attached bin file


DHLgh:
This issue began with version .78 and is still present in .79. Thank you!
[ProblemFile.zip](https://github.com/CCExtractor/ccextractor/files/120846/ProblemFile.zip)

barun511:
Unable to reproduce, running ccextractor returns the attached file (turned into a .txt to comply with git)
[problemfileresult.txt](https://github.com/CCExtractor/ccextractor/files/170548/problemfileresult.txt)

anshul1912:
Already solved by arbitrer, only reproducible on Windows

cfsmp3:
But is this solution good? Isn't it the one that checked if the file was
open instead of the reason for which it could be closed?

On Sat, Mar 12, 2016 at 8:13 PM, Anshul Maheshwari <notifications@github.com

canihavesomecoffee:
Can be closed once #323 is merged (after bot confirms that PR doesn't introduce new regressions). Tested the PR manually and it solves this issue.

PR #293, opened 2016-03-06
Small input files

Arbitrer:
Tickets #277 , #281 , #287 are solved by an if here.


canihavesomecoffee:

canihavesomecoffee:
Tests match the last one from Anshul, so from that point of view this PR is fine :+1:

anshul1912:
This is just removing print and adding that again, no way we should merge it.
@Arbitrer can you please close this pull request.

PR #294, opened 2016-03-06
Additional Argument added to sprintf giving the correct format specifier

abhishek-vinjamoori:
%s argument added. Alternative would be using a different function instead of sprintf and change to strlcat. Then it has to be changed everywhere.


cfsmp3:
the printf() family of functions copy all characters that are not escaped with %... so doing %s and passing the string to copy as an extra parameter is equivalent to just passing that string in the format parameter.

abhishek-vinjamoori:
But the GCC compiler throws a warning stating that there is no format specifier. So wouldn't it be correct to add it anyways to remove the warning and ensure that the nothing is left to the compiler to decide the type ?

Issue #290, opened 2016-03-06
error while  building ccextractor


vinayakathavale:
There is an error while executing make [link](http://pastebin.com/ZwGBTNYk) 
there seems to be an error while importing file "zvbi_decoder.h"
output of cmake [link](http://pastebin.com/qDXursHc)
I am using cmake version 3.3.2

anshul1912:
I think better solution would be to change makefile instead of source code, that change makes the position of file fix,

vinayakathavale:
I made [this](https://github.com/DrImpossible/ccextractor/commit/5517bd8363006799b68735f82835d3a2ccbed102) change but it does not fix the issue make still [fails](http://pastebin.com/QTjgMWQZ)

vinayakathavale:
issue resolved  [by](https://github.com/CCExtractor/ccextractor/pull/296) pull request #296

PR #295, opened 2016-03-06
Solution for #290

ioanas96:
Changed a couple of headers.
Added lines 48, 50 and 51in CMakeLists.txt, besides the default ones.


anshul1912:
Your solution looks good, but your commit messages are bad, can you please try again with new build, and create pull request of only those changes that you have done to fix the issue

ioanas96:
Yes, sure

PR #298, opened 2016-03-07
Modified headers and added CMakeLists.txt - issue #290

ioanas96:
I left the changes in the headers because on my computer ccextractor wouldn't be built without them.
I uploaded my own version of CMakeLists.txt.


anshul1912:
cant merge, commit still have modification that is not required

Issue #288, opened 2016-02-26
ccextractor segfaults when decoding CC using -sc when a blank line is encountered


jimmyoh:
Hello,

I am using ccextractor to decode Closed caption from live TV streams.
US ATSC live TV.
When I use -sc (Sentence Capitalisation)
ccextractor segfaults with code 139 
if it encounters a blank line in the closed captioning.

**decoded with -sc**
330
00:17:26,880 --> 00:17:29,113
                        Uh...   

331
00:17:29,182 --> 00:17:32,116
                    (<i>mumbling</i>)  

332
00:17:33,453 --> 00:17:35,386
                    Oh.         

333
00:17:48,101 --> 00:17:50,168

**Decoded Without -sc

**

331
00:17:29,182 --> 00:17:32,116
                    (<i>mumbling</i>)  

332
00:17:33,453 --> 00:17:35,386
                    OH.         

333
00:17:48,101 --> 00:17:50,168
        .                       

334
00:17:50,236 --> 00:17:54,705
UH, DR. MARK SLOAN              
FROM COMMUNITY GENERAL HPITAL.

anshul1912:
can you share dumped video, with which ccextractor crashes. It will help
developers to reproduce the issue.

Thanks
Anshul Maheshwari

On Sat, Feb 27, 2016 at 12:20 AM, jimmyoh <notifications@github.com> wrote:

> Hello,
>
> I am using ccextractor to decode Closed caption from live TV streams.
> US ATSC live TV.
> When I use -sc (Sentence Capitalisation)
> ccextractor segfaults with code 139
> if it encounters a blank line in the closed captioning.
>
> *decoded with -sc*
> 330
> 00:17:26,880 --> 00:17:29,113
> Uh...
>
> 331
> 00:17:29,182 --> 00:17:32,116
> (*mumbling*)
>
> 332
> 00:17:33,453 --> 00:17:35,386
> Oh.
>
> 333
> 00:17:48,101 --> 00:17:50,168
>
> **Decoded Without -sc
>
> **
>
> 331
> 00:17:29,182 --> 00:17:32,116
> (*mumbling*)
>
> 332
> 00:17:33,453 --> 00:17:35,386
> OH.
>
> 333
> 00:17:48,101 --> 00:17:50,168
> .
>
> 334
> 00:17:50,236 --> 00:17:54,705
> UH, DR. MARK SLOAN
>
> FROM COMMUNITY GENERAL HPITAL.
>
> —
[...]

PR #300, opened 2016-03-08
Fixed issue #288 : Segfault on using -sc for a blank line

Abhinav95:
The seg-fault would happen if the value of string p1 or p2 is null, which happens in a blank line. Returning a non-zero value for string_cmp2 (1 chosen arbitrarily here) if the null case happens prevents the seg-fault from happening.


cfsmp3:
This hides the problem but doesn't solve it. We shouldn't be calling any function with invalid values, and it's the responsibility of the caller to check parameters.

PR #307, opened 2016-03-11
Added --forceflush (-ff) option

### LINKED ISSUE #306: Add --forceflush (-ff) as parameter
An easy one for GSoC applicants:

We need to add a parameter --forceflush (short -ff) that if used will cause the file buffers to be flushed as soon as content is written.

Actually the functionality itself is already there 
(file ccx_encoders_common.c, check the call to fsync)

but currently it's forced. The required change implies is that only if forceflush is used fysnc is to be called.

Aesmade:
Added option for forcing a flush on the output buffer whenever content is written

### DIFF
--- a/src/lib_ccx/ccx_common_option.c
+++ b/src/lib_ccx/ccx_common_option.c
@@ -56,6 +56,7 @@ void init_options (struct ccx_s_options *options)
 	options->xmltvoutputinterval=0; // interval in seconds between writting xmltv full file output
 	options->xmltvonlycurrent=0; // 0 off 1 on
 	options->keep_output_closed = 0; // By default just keep the file open.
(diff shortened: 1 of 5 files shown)


Aesmade:

canihavesomecoffee:

@Aesmade I'll run them for you :)

anshul1912:
you are changing the default behaviour of ccextractor, now we only force the buffer if -ff option is passed. you are trying to make it optional, may I know the use case, where you find force buffer is not useful and create problem for you.

canihavesomecoffee:
@anshul1912: It was requested by @cfsmp3, because when he added it it slows down CCExtractor (or at least the bot test run) with 100%, hence his request in #306 to make it optional again

PR #309, opened 2016-03-11
Flush only when -ff is parsed

abhishek-vinjamoori:
Still looking if more files need to be changed.


abhishek-vinjamoori:
Doesn't work in all cases. Will update soon.

abhishek-vinjamoori:
Closing this pull request because it contains some unnecessary debugging print statements. Making a new pull request.

Issue #284, opened 2016-02-20
Issues with ISDB (Brazil) support


cfsmp3:
All seem easy to solve (so possibly a good chance for GSOC applicants to get points).

On the Brazilian files, the text is great (ignore the failure to show utf8 properly), but the timestamps don't increment -- and the second timestamp doesn't respond to the -delay switch:

ccextractor -datets -pn 48352 -noru -out=ttxt -unixts 0 -delay 1454292360000 -o 2016-02-01_0206_BR_Globo_Test_02.ccx2.out /tv/2016/2016-02/2016-02-01/2016-02-01_0206_BR_Globo_Test_02.mpg

csa@esfinge:/tv/2016/2016-02/2016-02-01$ cat 2016-02-01_0206_BR_Globo_Test_02.ccx2.out ﻿
20160201020600,000|19700101000000,002|ISDB|PARTICIPOU DAS PRINCIPAIS
20160201020600,000|19700101000000,002|ISDB|JOGADAS OFENSIVAS DO... OUTRO
20160201020600,000|19700101000000,002|ISDB|TOQUE, E DEPOIS ELE ENTRA PELA
20160201020600,000|19700101000000,002|ISDB|JOGADAS OFENSIVAS DO... OUTRO
20160201020600,000|19700101000000,002|ISDB|TOQUE, E DEPOIS ELE ENTRA PELA
20160201020600,000|19700101000000,002|ISDB|ESQUERDA, OLHA L� O ESPA�O QUE
20160201020600,000|19700101000000,002|ISDB|TINHA O JORGE HENRIQUE PARA
20160201020600,000|19700101000000,002|ISDB|TENTAR DAR UM TOQUE. E NEN�
20160201020600,000|19700101000000,002|ISDB|AINDA FAZ O GESTO, P�, PODIA TER
20160201020600,000|19700101000000,002|ISDB|TOCADO PARA MIM.
20160201020600,000|19700101000000,002|ISDB|>> [LUIS ROBERTO] LANCES DO
20160201020600,000|19700101000000,002|ISDB|PRIMEIRO TEMPO EM S�O JANEIRO
20160201020600,000|19700101000000,002|ISDB|PASSO AIR, A COLINA HIST�RICA,

If I include the -UCLA switch, I do get the offset on the second timestamp:

[...]

Issue #273, opened 2016-02-02
Automatic videotext Subtitle extraction fails for some HD TS files


workflowsguy:
When running CCExtractor 0.79 on those files, the output is as follows:

```
Opening file: /Volumes/Public/bearbeiten/Videos/extrahieren Untertitel/Konzerne klagen - Wir zahlen.eyetv/000000001c3bc241.mpg
Detected MP4 box with name: mfra

File seems to be a MP4

Analyzing data with GPAC (MP4 library)
Creating /Volumes/Public/bearbeiten/Videos/extrahieren Untertitel/Konzerne klagen - Wir zahlen.eyetv/Konzerne klagen - Wir zahlen.srt
opening '/Volumes/Public/bearbeiten/Videos/extrahieren Untertitel/Konzerne klagen - Wir zahlen.eyetv/000000001c3bc241.mpg': [iso file] Incomplete box en.R
failed to open


Total frames time:	  00:00:00:000  (0 frames at 29.97fps)

Done, processing time = 57 seconds
This is beta software. Report issues to carlos at ccextractor org...
0

```

When running telxcc on the file, the subtitles are correctly extracted.

It seems that autodetection for video type fails in this case. It works if the parameter "in=ts" is added to the command line. 

The MediaInfo information for this file is:
```
General
ID                                       : 1039 (0x40F)
Complete name                            : /Volumes/Public/bearbeiten/Videos/extrahieren Untertitel/Konzerne klagen - Wir zahlen.eyetv/000000001c3bc241.mpg
Format                                   : MPEG-TS
File size                                : 4.29 GiB
Duration                                 : 43mn 13s
Start time                               : UTC 2016-01-05 09:25:58
End time                                 : UTC 2016-01-05 10:09:11
Overall bit rate mode                    : Variable
Overall bit rate                         : 14.2 Mbps
Country                                  : DEU
Timezone                                 : +01:00:00

Video
ID                                       : 5411 (0x1523)
Menu ID                                  : 10376 (0x2888)
Format                                   : AVC
Format/Info                              : Advanced Video Codec
Format profile                           : High@L4
Format settings, CABAC                   : Yes
Format settings, ReFrames                : 4 frames
Codec ID                                 : 27
Duration                                 : 43mn 13s
Bit rate                                 : 12.7 Mbps
Width                                    : 1 280 pixels
Height                                   : 720 pixels
Display aspect ratio                     : 16:9
Frame rate                               : 50.000 fps
[...]

PR #314, opened 2016-03-12
#284 made end timestamp respond to delay switch also fix for #273

vinayakathavale:


abhibansal530:
But it does not solves the issue of timestamp being not incremented.

abhishek-vinjamoori:
@abhibansal530. It's just a partial fix to the whole issue.

vinayakathavale:
@canihavesomecoffee  can you run regression tests on this one 
Thanks

src/lib_ccx/stream_functions.c
```diff
@@ -94,14 +94,14 @@ void detect_stream_type (struct ccx_demuxer *ctx)
...
+		if (boxScore > 7) // will check again for mp4 with boxscore > 1 after checking for all other file types
 		{
 			// We had at least one box (or multiple) at the end to "claim" this is MP4. A single valid box at the end is doubtful...
 			ctx->stream_mode = CCX_SM_MP4;
 		}
 	}
 	if (ctx->stream_mode==CCX_SM_ELEMENTARY_OR_NOT_FOUND) // Still not found
-	{
+	{	
 		if (ctx->startbytes_avail > 188*8) // Otherwise, assume no TS
```
anshul1912: Please avoid whitespace.

src/lib_ccx/stream_functions.c
```diff
@@ -195,7 +228,7 @@ int detect_myth( struct ccx_demuxer *ctx )
 {
 	int vbi_blocks=0;
 	// VBI data? if yes, use myth loop
-	// STARTBTYTESLENGTH is 1MB, if the file is shorter we will never detect
+	// STuARTBTYTESLENGTH is 1MB, if the file is shorter we will never detect
 	// it as a mythTV file
```
anshul1912: why STuart start was looking much better ;)

anshul1912:
@vinayakathavale : there is an option -f for git push, if you already know the mistake that you have done, I would suggest better to git reset HEAD^  then update your changes then commit it.
while pushing use git push -f

src/lib_ccx/stream_functions.c
vinayakathavale: sorry will change that

canihavesomecoffee:
@vinayakathavale Will do, it escaped my attention...

canihavesomecoffee:
Doesn't look to break more than what is currently broken, so :+1:

PR #318, opened 2016-03-12
added --forceflush (-ff)

ravipatianirudh:
added **force flush**. Works with command line options -ff and --forceflush


cfsmp3:
This PR messes file permissions

old mode 100755
new mode 100644

(that's removing +x in the build script)

Please submit a PR that doesn't do that :-)





On Sat, Mar 12, 2016 at 4:41 PM, Aniruddh Ravipati <notifications@github.com

PR #319, opened 2016-03-13
Added Cyrillic support for Teletext

### LINKED ISSUE #283: Cyrilic support. Missing or broken.
When processing a Russian TV capture we're getting Western characters instead of Cyrillic. I see we need Cyrillic G0 Primary Set - Option 2 - Russian/Bulgarian -- how does that get turned on?

The first test file is available here:

http://vrnewsscape.ucla.edu/dropbox/2016-02-09_1800_RU_TVC_Test.mpg
http://vrnewsscape.ucla.edu/dropbox/2016-02-09_1800_RU_TVC_Test.txt

VLC plays the teletext just fine
cyrilic1
cyrilic2

bigharshrag:
Added Cyrillic support (Cyrillic-1, Cyrillic-2, Cyrillic-3 character sets) for teletext which had earlier been missing which resolves #283 .


abhishek-vinjamoori:
// Now we have the begining of page transmission; if there is page_buffer pending, process it
		if (ctx->page_buffer.tainted == YES)
		{
			// Convert telx to UCS-2 before processing
			for(uint8_t yt = 1; yt <= 23; ++yt)
			{
				for(uint8_t it = 0; it < 40; it++)
				{
					if (ctx->page_buffer.text[yt][it] != 0x00)
						ctx->page_buffer.text[yt][it] = telx_to_ucs2(ctx->page_buffer.text[yt][it]);
				}
			}


Can I know why this is being done before ?
It is causing some issues, for fixing other bugs.

bigharshrag:
When dealing with character sets other than the Latin G0 character set, the X/28 packet may be received after the data from the X/1 to X/25. The X/28 (or M/29) contain data regarding the default character sets. Since it was earlier being converted to UCS-2 there itself, Latin data was being written for the first time, after which it switched to the new G0 character set. This simply delays conversion from telx to UCS-2 until data about default G0 character set is received.

Could you please share what issues this is causing. Please share the related sample files as well.

Issue #270, opened 2016-01-30
MPEG2 Transport streams no longer processed


workflowsguy:
When trying to open MPEG2 Transport stream files, ccextractor now fails with "Unable to open" & error code 6.

By comparison, telxcc has no problem extracting teletext subtitles from such a file.

CCExtractor 0.78 CLI installed on OS X through Homebrew.

canihavesomecoffee:
Hello,

could you try running .80, which is the latest version released a week ago?

workflowsguy:
I do not know how to build ccextractor, so I will have to wait until the version becomes available through Homebrew on the Mac. Then I can try it out.

canihavesomecoffee:
I don't own a mac myself, but from what I see in the [mac](https://github.com/CCExtractor/ccextractor/tree/master/mac) folder of the repository, all you need to do is perform these steps: 

* install the [Apple Developer Tools](http://developer.apple.com/technologies/tools/)
* download the latest [0.79 source files](http://sourceforge.net/projects/ccextractor/files/ccextractor/0.79/ccextractor.src.0.79.zip/download) from SourceForge
* extract the zip
* open the terminal
* navigate to the mac folder of the extracted folder (e.g. ``cd ~/Downloads/ccextractor0.79/mac``)
* run the command which should compile ccextractor for you (``./build.command``)

If all goes well and there are no errors, you should have a ccextractor executable in the same folder. If an error occurs, it would be nice if you could open a new issue for this, so we can update the mac build command (as @cfsmp3 noted, none of us contributors owns a mac, so we can't test it every time we release unfortunately).

If you don't want to do this, would it be possible to make a small sample available, so we can test it on Linux with the current version (and with 0.78 as well) to confirm the issue?

workflowsguy:
Thanks for the instructions,

unfortunately it fails to build on my Mac, so I have created a new issue as per your suggestion.

workflowsguy:
I have tested this with a build of 0.79. However, it still produces the error message

```
Unable to open 


Error: Failed to determine total file size.

6

```

cfsmp3:
That means it can't find or open the input file.

workflowsguy:
Sorry, this was caused by an error in my extraction script.

However, it would be great if in the case of such an error the name of the file were listed.
I mistakenly thought that CCExtractor could not open the video file; instead the path to the srt file to be created was incorrect.
With a more verbose error message, I would have spotted my error immediately.

[...]

PR #320, opened 2016-03-14
Fix #281. Crash issue in small file under 1Mib. & Referring Issue #270. Detailed error message when opening files 

### LINKED ISSUE #281: File that makes CCExtractor crash

YorkHe:
First Commit

Because the default buffer size is set to 1024*1024. When encounters a file smaller than 1Mib, it will determined the file as done reading and close it without going through the processing loop before entering `LSEEK`  in `ccx_demuxer_getfilesize` in file `ccx_demuxer.c`, then it fails. 

The fix is just smaller the buffer size to 100K, which might be a hack but it works for most files.

=================================
Second commit.

Referring Issue #270. Detailed opening file error message.

I commit this before last PR merged, so they seems coming into one PR.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -84,8 +84,22 @@ int main(int argc, char *argv[])
 	if (ccx_options.binary_concat)
 	{
 		ctx->total_inputsize=gettotalfilessize(ctx);
-		if (ctx->total_inputsize==-1)
-			fatal (EXIT_UNABLE_TO_DETERMINE_FILE_SIZE, "Failed to determine total file size.\n");
+		if (ctx->total_inputsize < 0)
+		{
+			switch (ctx->total_inputsize)
+			{
+			case -1*ENOENT:
+				fatal(EXIT_NO_INPUT_FILES, "Failed to open file: File not Exist");
+			case -1*EACCES:
+				fatal(EXIT_READ_ERROR, "Failed to open file: Unable to access");
+			case -1*EINVAL:
+				fatal(EXIT_READ_ERROR, "Failed to open file: Invalid opening flag.");
+			case -1*EMFILE:
+				fatal(EXIT_TOO_MANY_INPUT_FILES, "Failed to open file: Too many files are open.");
+			default:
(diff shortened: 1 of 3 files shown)


cfsmp3:
We cannot reduce the buffer size, that breaks other things for sure... and
is likely to slow things down.

What does "determine as file close" mean? Where in the code does that
happen? That's what we have to fix...

YorkHe:
@cfsmp3 That's in function `buffered_read_opt` in file `file_functions.c`. The buffer reading procedure only goes for one loop so when doing the prefetch(switch_to_next_file) the read value returns 0. So eof is marked 1 and determine the file as closed.   I've just changed the predefined STARTBYTESLENGTH, which would influence the preread or stream_detect. But I think 100KB is bigger than any box or packet in video streaming so it would influence a lot. I'm now having a class and reply this by tablet. Perhaps you could run a automatic test to see if this PR is valid.

canihavesomecoffee:
@YorkHe your wish is my command ;)

cfsmp3:
I think your analysis is good but the solution is not. .bin files can have
any length (zero, or non-zero but less than 1 KB) so any solution that
consists on reducing the buffer size is not general. But there's also
another problem which is that reducing the buffer does affect performance
(and many other operations that do need a larger buffer) in the more
general case of large transport streams.

If there's an issue with the function logic let's fix that instead of
tweaking the constants so that it seems to work :-)

Also Willem, could you add a couple of .bin and.raw files to the test suite
(large and small)?

YorkHe:
@canihavesomecoffee Thanks for the testing.  It seems I failed a lot... 
@cfsmp3 You're right, I'll take a deeper look into this. Maybe there will be some refactoring jobs...

I'll close this PR and thanks for your comment,

canihavesomecoffee:
@YorkHe Not sure if it's related to this commit though. Some certainly will be, but I suspect some others are related to earlier changes (that's why I'm running it now on a previous commit to check)

PR #330, opened 2016-03-16
Teletext: replace * with @ according to X/26 packet specification

### LINKED ISSUE #249: @ in teletext-subs saved as asterisk
CCExtractor 0.77 and git-677fee4 
http://www.mediafire.com/download/q6ebvmzwe1prvi3/cce-at-sign.7z (25MB)

Teletext:
at sign

SRT:
```
2
00:00:12,840 --> 00:00:16,520
ORF 2015
untertitel*orf.at
```

isacdaavid:
fixes #249 
This is my initial solution. I think it's still possible to follow the standard's semantics more closely by actually overwriting whatever charset the current X/26 packet happens to be in.

I tested it against the video samples referenced in #249


cfsmp3:
A different solution has been merged, but yours was valid too - you get the points.
Please mention the PR in your proposal :-)

PR #334, opened 2016-03-17
Fixing issues with ISDB

Abhinav95:
#284 
The current fix I have used is setting the current timestamp of the detected subtitle line to the FTS at that point. This makes the timestamps increment according to the time the subtitle was found in the file (which is correct), but the end time is considered as the start time of the next subtitle line (which is not always correct). In any case, it is a much more accurate representation and way better than the timestamps not incrementing at all.

Also, this pull request takes care of removing the "NA" in the third column when using -UCLA.

### DIFF
--- a/src/lib_ccx/ccx_common_option.c
+++ b/src/lib_ccx/ccx_common_option.c
@@ -57,6 +57,7 @@ void init_options (struct ccx_s_options *options)
 	options->xmltvonlycurrent=0; // 0 off 1 on
 	options->keep_output_closed = 0; // By default just keep the file open.
 	options->force_flush = 0; // Don't flush whenever content is writtern.
+	options->ucla = 0; // By default, -UCLA not used
 
 	options->transcript_settings = ccx_encoders_default_transcript_settings;
 	options->millis_separator=',';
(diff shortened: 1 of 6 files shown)


Abhinav95:
@canihavesomecoffee Could you run tests on this please? :)

abhishek-vinjamoori:
@Abhinav95  Just went through your commit and I can see that you are using get_fts() method by default for isdb videos. But according to me this is wrong, because there are some videos in which PCR maybe present in such cases PTS must not be used. Otherwise doing so, will break files which do not use PTS timing.

Abhinav95:
@abhishek-vinjamoori The get_fts() method which I have added will get called only if the type of the data in the buffer in 'general_loop.c' is detected as a valid ISDB subtitle line, and the FTS will correspond to the time in the video when the line was detected.
Also, PCR timing is handled in 'ts_functions.c', and will not affect anything in this file.
The starting timestamp using get_fts() as done here will be correct.
The problem here is the ending timestamp, which my pull request just assumes to be the starting timestamp of the next detected line, which may not necessarily hold true. Do you see any way to address that?

Abhinav95:
@anshul1912 It was a leftover of some change I was trying, but is actually a trivial change.

Abhinav95:
This breaks files in which the global timestamp was already initialized.
The fix made broken samples work and working samples break.
Sending another pull request in which both work

PR #336, opened 2016-03-18
Fixing issues with ISDB

Abhinav95:
#284 
ISDB timestamps were not incrementing in some samples where the global timestamp was not initialized. In those cases, set the timestamp to the FTS.

Pull request #334 broke the output in valid files in which the global timestamp was correctly initialized. Fixed that in this one.

Also, this pull request takes care of removing the "NA" in the third column when using -UCLA.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -577,19 +577,19 @@ int write_cc_subtitle_as_transcript(struct cc_subtitle *sub, struct encoder_ctx 
 
 			if (context->transcript_settings->showStartTime)
 			{
-				char buf1[80];
+				char buf[80];
 				if (context->transcript_settings->relativeTimestamp)
 				{
-					millis_to_date(start_time + context->subs_delay, buf1, context->date_format, context->millis_separator);
-					fdprintf(context->out->fh, "%s|", buf1);
+					millis_to_date(start_time + context->subs_delay, buf, context->date_format, context->millis_separator);
+					fdprintf(context->out->fh, "%s|", buf);
 				}
 				else
 				{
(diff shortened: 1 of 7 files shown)


canihavesomecoffee:

src/lib_ccx/ccx_encoders_common.c
```diff
@@ -577,19 +577,19 @@ int write_cc_subtitle_as_transcript(struct cc_subtitle *sub, struct encoder_ctx
...
 				{
 					time_t start_time_int = (start_time + context->subs_delay) / 1000;
 					int start_time_dec = (start_time + context->subs_delay) % 1000;
 					struct tm *start_time_struct = gmtime(&start_time_int);
-					strftime(buf, sizeof(buf), "%Y%m%d%H%M%S", start_time_struct);
-					fdprintf(context->out->fh, "%s%c%03d|", buf, context->millis_separator, start_time_dec);
+					strftime(buf1, sizeof(buf1), "%Y%m%d%H%M%S", start_time_struct);
+					fdprintf(context->out->fh, "%s%c%03d|", buf1, context->millis_separator, start_time_dec);
 				}
 			}
```
anshul1912: why this change

src/lib_ccx/general_loop.c
```diff
@@ -801,12 +801,21 @@ void general_loop(struct lib_ccx_ctx *ctx)
...
+								- ctx->demux_ctx->min_global_timestamp;
+				}
+				else
+				{
+					// Temporary fix if global timestamp not inited
+					tstamp = get_fts(dec_ctx->timing, dec_ctx->current_field);
+				}
 				isdb_set_global_time(dec_ctx, tstamp);
-                        }
+            }
```
anshul1912: we use tab for indentation

Issue #257, opened 2015-12-13
Dump XDS packet when not recognized


cfsmp3:
The XDS decoder supports all packages we've ever found in actual streams.

From time to time a new one appears though - we should hexdump its contents.

PR #338, opened 2016-03-18
Fix for #257

vinayakathavale:


cfsmp3:
Hexdump should be done using this function:

void dump (LLONG mask, unsigned char *start, int l, unsigned long abs_start, unsigned clear_high_bit)

Also only the current packet needs to be dumped, not all of them - and just up to the actual length, not the maximum.

PR #348, opened 2016-03-28
Fixing problems with XDS

### LINKED ISSUE #347: XDS not working
The following parameters

-out=txt -xdsdebug "F:\ConBackup\Closed captions\CC\Samples variados\Correct\With XDS\Survivor S23E01 1080i HDTV DD5.1 MPEG2.ts" 

should produce XDS. However, in the function
void do_end_of_xds (struct cc_subtitle *sub, struct ccx_decoders_xds_context *ctx, unsigned char expected_checksum)

ctx is passed NULL in the call from 
do_end_of_xds (sub, dec_ctx->xds_ctx, lo);

Which in turn can be traced here (in ccx_decoders_common.c)
	// Init XDS buffers
	if(setting->ignore_xds == CCX_TRUE)
		ctx->xds_ctx = NULL;
	else
		ctx->xds_ctx = ccx_decoders_xds_init_library(ctx->timing);

etc

Anyway XDS needs to be processed *regardless* of whether it's going to be exported to a final text file or not, because XDS carries the program name, station name and more which can be used for other things than write this data to file. For starters the GUI use them to show the information to the user.

GSoC points: 3 (for a perfect solution in which XDS is processed but only written to file in the formats in which is makes sense, obviously not in .srt etc)

Abhinav95:
Fixes #347 
Always parsing XDS data. (ctx never initialized to null).
When using -xdsdebug, debug data always prints on screen but XDS data is written to file only in case of transcripts.

### DIFF
--- a/src/lib_ccx/ccx_decoders_common.c
+++ b/src/lib_ccx/ccx_decoders_common.c
@@ -341,16 +341,15 @@ struct lib_cc_decode* init_cc_decode (struct ccx_decoders_common_settings_t *set
 
 	ctx->anchor_seq_number = -1;
 	// Init XDS buffers
-	int xds_write_to_file;
 	if(setting->output_format==CCX_OF_TRANSCRIPT)
 	{
-		xds_write_to_file = 1;
+		setting->xds_write_to_file = 1;
 	}
 	else
 	{
-		xds_write_to_file = 0;
+		setting->xds_write_to_file = 0;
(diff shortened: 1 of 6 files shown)


canihavesomecoffee:

canihavesomecoffee:

canihavesomecoffee:
Looks good :+1:

PR #368, opened 2016-05-24
Fixes #139 - Case fixing in teletext

### LINKED ISSUE #139: Case fixing in teletext
For American closed captions CCExtractor is able to apply some case conversion rules so instead of all caps we get reasonably correct case.

We need to apply the same logic to Teletext (and well, DVB).

Abhinav95:
Applying modified 'capitalize' and 'correct_case' functions for teletext with the same logic.


Abhinav95:

Abhinav95:

PR #370, opened 2016-05-25
Fixes #353 : High memory consumption

### LINKED ISSUE #353: High Memory Consumption - 180 MB
Shelling out ccextractor instances on Linux from an app to process parallel streams.  Feeding low-bit-rate video via stdin, using stream mode, disabled input buffering, captions going into stdout.  Seems each instance occupies a fixed amount of ~180 MB of RAM, severely limiting what we can accomplish on each box.

Is it possible to configure the size of internal buffers more modestly?  Any insight into the 180 MB RAM footprint?

Thank you!
Jose

Abhinav95:
Saves around 180 MB of RAM by separating the EPG context from the general context, and allocating memory for it only when needed, i.e only when -xmltv is passed as a parameter.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -24,6 +24,7 @@ struct ccx_s_options ccx_options;
 int main(int argc, char *argv[])
 {
 	struct lib_ccx_ctx *ctx;
+	struct epg_ctx *epgctx;
 	struct lib_cc_decode *dec_ctx = NULL;
 	int ret = 0;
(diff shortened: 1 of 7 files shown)


Abhinav95:

Abhinav95:
Despite passing most tests (crashing on some) and seeing an improvement in runtimes, this is not ready to be merged with the current approach. I will work on a different approach and update the PR.

Abhinav95:

PR #373, opened 2016-05-26
Reducing Memory Consumption

### LINKED ISSUE #353: High Memory Consumption - 180 MB

Abhinav95:
Fixes #353 
EPG variables were being initialized and taking up memory or around 180 MB while never being used. Fixed this by initializing them only when necessary.

Also includes a fix to passing the wrong context to the detect_myth() function, which was surprisingly still working with the live repository.


Abhinav95:
@ccx-testerbot runtests memoryfix-new

PR #390, opened 2016-06-14
skipping redundant bytes at the end of tx3g atom

rkuchumov:


rkuchumov:

Issue #394, opened 2016-06-22
Illegal .srt being created from DVBs


cfsmp3:
Lines like this (processing 01-BBC1.EastEnders.ts):

334
00:16:49,197 --> 00:16:51,966
Jane had squeezed my hand.

Yeah. . .was it just that?

Are illegal in .srt (there cannot be empty lines in a subtitle frame in .srt)

I think we're just adding an extra \n at the end of each line. 

Also, we're not consistent with \n vs \r\n, the same .srt contains both.

cfsmp3:
The line separators in Windows must be \r\n, not just \n... I've corrected this by passing encoded_crlf to add_ocrtext2str.

In fact I believe for .srt it's always CRLF regardless of the platform 

See 
http://www.textfiles.com/uploads/kds-srt.txt

Issue #391, opened 2016-06-20
Tesseract fail cause spngpng to fail


cfsmp3:
This command:

ccextractor -out=spupng -parsePMT -parsePAT -codec dvbsub  "C:\DeletableTempStuff\UK Samples\chan52_GOOD.ts"

fails with the error "tessaract cannot find any language!" for me. I understand the issue is that I'm missing some tesseract files, but the output is spupng so I should get output anyway.

In fact, if I compile without OCR support, it works fine (as expected).

anshul1912:
need complete logs, MSG ""tessaract cannot find any language!" comes from inside tesseract, full logs would help me to get traces.

PR #398, opened 2016-07-03
corrrect Incorrect progress display

### LINKED ISSUE #397: 01-BBC1.EastEnders.ts: Incorrect progress display
In 01-BBC1.EastEnders.ts (using this file but I assume it's a general problem) the first time progress is reported it prints this:

 0%  |  202:07

That timestamp is calculated here:

	LLONG t=get_fts(dec_ctx->timing, dec_ctx->current_field);
	if (!t && ctx->demux_ctx->global_timestamp_inited)
	t=ctx->demux_ctx->global_timestamp - ctx->demux_ctx->min_global_timestamp;
	int cur_sec = (int) (t / 1000);
	activity_progress(progress, cur_sec/60, cur_sec%60);

With  ctx->demux_ctx->min_global_timestamp==0

Anyway, the timing variables are fixed by the next pass (of course something is broken here already) and the correct time start. But that the 7 to the right is left there forever because the new lines only use 2 characters for the hour.

Of course no, it's not a solution to write an extra space :-)

anshul1912:
correct Incorrect progress display
close #397

### DIFF
--- a/src/lib_ccx/dvb_subtitle_decoder.c
+++ b/src/lib_ccx/dvb_subtitle_decoder.c
@@ -444,11 +444,6 @@ void* dvbsub_init_decoder(struct dvb_config* cfg)
 
 #ifdef ENABLE_OCR
 	ctx->ocr_ctx = init_ocr(ctx->lang_index);
-        if(!ctx->ocr_ctx)
-	{
-		freep(&ctx);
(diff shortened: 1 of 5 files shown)


anshul1912:
close #391

anshul1912:
close #394

PR #402, opened 2016-07-20
DVB Color Detection

Abhinav95:
Detects and outputs word-wise color and the point of speaker changes in DVB subtitles. Outputs font tags to SRT and WebVTT.

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -25,6 +25,7 @@ struct image_copy
 	png_color *palette;
 	png_byte *alpha;
 	unsigned char *data;
+	int bgcolor;
 };
 
 static int check_trans_tn_intensity(const void *p1, const void *p2, void *arg)
@@ -75,8 +76,8 @@ static int search_language_pack(const char *dirname,const char *lang)
 void delete_ocr (void** arg)
 {
 	struct ocrCtx* ctx = *arg;
-		TessBaseAPIEnd(ctx->api);
-		TessBaseAPIDelete(ctx->api);
+	TessBaseAPIEnd(ctx->api);
+	TessBaseAPIDelete(ctx->api);
 	freep(arg);
 }
 void* init_ocr(int lang_index)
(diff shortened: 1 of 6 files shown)


Abhinav95:

Abhinav95:

Abhinav95:

src/lib_ccx/ocr.c
```diff
@@ -133,29 +135,33 @@ int ignore_alpha_at_edge(png_byte *alpha, unsigned char* indata, int w, int h, P
 	}
 	cropWindow = boxCreate(start_y, 0, (w - (start_y + ( w - end_y) )), h - 1);
 	*out = pixClipRectangle(in, cropWindow, NULL);
-	boxDestroy(&cropWindow);
+	//boxDestroy(&cropWindow);
```
NilsIrl: Why comment this out?

Doesn't this make a memory leak?

cfsmp3: On inspection it seems like it :-) It's possible that by commenting it out we hid some problem under the rug. This is for @Abhinav95 but I believe he's super busy with this PhD or something :-)

Abhinav95: I think I changed the return type from int to BOX* and returned the instance of cropWindow to use later for something else. Could be refactored and yes it does lead to a memory leak later, although that should be a one line fix.

Alternatively we could also use a different and separate function with the return type as a BOX*. Easy GSoC qualification task?

NilsIrl: > Easy GSoC qualification task?

@cfsmp3 would I be allowed to fix it?

canihavesomecoffee: Nils, you don't need to ask those questions ;) If you want to fix it, please go ahead :+1:

cfsmp3: @Abhinav95 glad to see you around! Come by slack sometime to catch up! :-)

GSOC qualification task - maybe, but not easy :-) I mean, it's "easy" for us, but this code is deep into the program, you don't trivially reach it or test it.

You don't have any friend you could send our way, do you? It's harder and harder to find C programmers on GSoC.

Abhinav95: Yes it's a bit tougher to get C devs. Everyone I know in my bubble is using python (plus the usual machine learning stack) for their bread and butter.

I'll keep an eye out for a good student if possible. GSoC has lost its appeal to the top students at my old uni ever since they reduced the stipends (which was in my 2nd GSoC year, 2017 I think).

NilsIrl: Ok I opened PR #1187

PR #418, opened 2016-09-03
Fix protobuf-c include

canihavesomecoffee:
The <> make the build fail in Release mode in Visual Studio, as the <>
makes it look for a standard header
(https://msdn.microsoft.com/query/dev14.query?appId=Dev14IDEF1&l=EN-US&k=k(C1083)&rd=true).
This change makes it compile both in Debug and Release mode.


cfsmp3:
We can't modify this file:

/* Generated by the protocol buffer compiler.  DO NOT EDIT! */
/* Generated from: ccx_sub_entry_message.proto */

Changes are overwritten each time it's generated, so we need to change the
directories in the project files I'm afraid.

Dakta:
Bumping to note that this issue affects Mac build as well. Works on Linux (tested with a Ubuntu 14 instance I had sitting around).

cfsmp3:
@Dakta This PR cannot be applied - please open a issue instead.

PR #419, opened 2016-09-03
Fix regression introduced in 1a1e973

canihavesomecoffee:
Due to an assignment instead of comparison, all types of codecs handled
after the DVD subtitles would now be processed as DVD subtitles...


Abhinav95:
Nice catch @canihavesomecoffee :D

Abhinav95:
Nice catch @canihavesomecoffee :D

Issue #432, opened 2016-10-10
Can't build Mac version


mackworth:
The github version does not compile; it complains about missing protobuf. I tried adding ` -I../src/` to the BLD_INCLUDE, and `SRC_PROTOBUF="$(find ../src/protobuf-c -name '*.c')"` to the BLD-SOURCES line, which avoids the error, but then it has the rather odd symptom of not producing an executable.

The 0.82 version downloaded from [the original ccextractor site](https://sourceforge.net/projects/ccextractor/?source=typ_redirect) does compile and does work, but it seems to be behind the github version.

cfsmp3:
Post the complete output from build so we can take a look.

mackworth:
First is output without any changes:
[ccextractor Build1.txt](https://github.com/CCExtractor/ccextractor/files/520031/ccextractor.Build1.txt)

Then I changed build.command to:
[build2.command.txt](https://github.com/CCExtractor/ccextractor/files/520032/build2.command.txt)

 and got essentially the same results without the protobuf errors:
[ccextractor build2.txt](https://github.com/CCExtractor/ccextractor/files/520037/ccextractor.build2.txt)

I should mention that many of the warning messages from clang should probably be looked into. It looks like several checks are not doing what they look like they're doing due to type issues.

cfsmp3:
Most likely you also have to add curl to the list of libraries.

mackworth:
Figured this one out; I missed an error message for `utf8proc` buried in all the warnings. I'll do a pull request for the revised build command that incorporates protobuf and utf8proc in the process.

mackworth:
fixed in PR #435

Issue #433, opened 2016-10-10
Stdin overrides --stream value


mackworth:
Using `-` to specify stdin as input file sets the stream timeout to -1, overrding the value set with the `--stream` option. The STDIN parameter processing should only change this value to -1 if it's still at the default 0). A workaround is to put `--stream 1` after the `-`, which works fine, despite the command-line syntax of `ccextractor [options] input file...`

mackworth:
Fixed in #435

PR #435, opened 2016-10-11
Fix Mac Build Process

mackworth:
Adds includes/compiles for protobuf and utf8proc. Fixes Issue #432.


mackworth:
Hmm, so git auto-added the fix for Issue #433  as well to this PR. If you want it separated, let me know.

PR #446, opened 2016-11-29
Fixed multiple spelling mistakes

ManveerBasra:


grave-w-grave:
my British English self cries at the correction of British English to American Englishes :(

ManveerBasra:
Oh, I'm sorry if I did that. I didn't mean to tbh, I just thought that was
how it was spelled. If you want me to change them back, let me know which
ones they are and I'll do it tomorrow.

cfsmp3:
This is a general thing - doesn't have a good solution, one side of the
Atlantic will be unhappy :-)
So for official acronyms like OCR we use whatever the official spelling is,
which I assume depends on who invented it.
For words that are not correctly written in either variant whoever fixes it
can use the variant they want.
Words that are correct in one variant are not changed.
Yes, this leads to having both British English and American English mixed
in the code, but if they can do it in TV shows all the time so can we,
being a TV show centric software.

anshul1912:
Can you remove British to American change from your pull request to get it merged.

PR #453, opened 2016-11-30
GCI: Clean up function names and refactor some code.

DanilaFe:
Aside from renaming hex2int --> hex_to_int and processhex --> process_hex for consistency, I also took the liberty to implement a hex_string_to_int function. It seems like hexadecimal characters are accessed from a string before being passed to hex_to_int, so it seems logical to eliminate that extra step and generalize the behavior a bit more. I did not introduce the new hex_string_to_int function into existing code to make sure it's reviewed before it's used.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -119,7 +119,7 @@ int main(int argc, char *argv[])
 
 	if (ccx_options.binary_concat)
 	{
-		ctx->total_inputsize=gettotalfilessize(ctx);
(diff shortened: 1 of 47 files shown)


DanilaFe:
Apologies for the random formatting changes. My editor seems to drop those there whenever I save.

DanilaFe:
print_mstime2bug and mstime_sprintf were similar enough to be merged into one function. I did that.

Issue #260, opened 2015-12-23
Incorrect output of number of processed frames and progress


workflowsguy:
Since Version 0.78,
ccextractor always outputs
"Total frames time:   00:00:00:000  (0 frames at 29.97fps)"

even when it successfully extracts subtitles.

Also, the progress output is no longer correct. It will output lines like this:

 88%  |  00:00
 89%  |  606:29
 90%  |  607:10
 91%  |  608:09
 92%  |  00:00
 93%  |  00:00
 94%  |  611:10
 95%  |  612:04
 96%  |  612:53
 97%  |  613:45
 98%  |  614:24
 99%  |  615:06
100%  |  00:00

I am using the command line version on OS X provided through Homebrew.

hurda:
It's also showing 29.97fps when the source-material is 50p.

anshul1912:
@hurda  can you provide sample

cfsmp3:
@workflowsguy we need a sample to test this - otherwise the ticket will be closed soon

workflowsguy:
Here is a sample:

https://www.dropbox.com/s/h53kaujpvjwdh69/Quarks%20%26%20Co.mpg?dl=0

It is a video clip in MP4 MPEG-TS format with both german teletext subtitles on page 150 and DVB-S subtitles. It was recorded with Elgato's EyeTV on OS X.
 
CCExtractor successfully extracts teletext subtitles but shows "Total frames time: 00:00:00:000 (0 frames at 29.97fps)". Also, various tools report the true fps rate of the file to be 50 fps.

Note that this sample (surprisingly) does not show the incorrect timestamp output that I reported above. If I find one, I will also post the link.

Thanks!

hurda:
> @hurda can you provide sample
You can use the sample from https://github.com/CCExtractor/ccextractor/issues/249 , (720p50 H264)

anshul1912:
This problem main reason is that these teletext subs does not of have frames but have packets (which are displayed) and no fix frame rate to calculate the total time,
so I have removed that print when its not atsc closed caption

workflowsguy:
In CCExtractor 0.79, there still are incorrect timecodes for the progress output.

```
  1%  |  727:54
  2%  |  728:23
  3%  |  729:01
  4%  |  729:34
  5%  |  730:07
  6%  |  730:44
  7%  |  731:25
  8%  |  732:11
  9%  |  732:57
 10%  |  733:37
```

[...]

PR #458, opened 2016-12-01
Added --noempty (GCI)

Izaron:
Task "Add an option not to generate output file if it's going to be empty (CCExtractor Development)"
[https://codein.withgoogle.com/dashboard/task-instances/6018877513269248/](Link)

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -404,6 +404,30 @@ int main(int argc, char *argv[])
 		curl_easy_cleanup(curl);
   	curl_global_cleanup();
 #endif
+
+	if (ccx_options.no_empty)
+	{
+		char *filename;
+		if (ccx_options.output_filename)
+		{
+			filename = ccx_options.output_filename;
+		}
+		else
+		{
+			int len;
+			char *extension = get_file_extension(ccx_options.enc_cfg.write_format);
+			len = strlen(ctx->basefilename) + 10 + strlen(extension);
+			filename = malloc(len);
+			sprintf(filename, "%s%s", ctx->basefilename, extension);
+			freep(&extension);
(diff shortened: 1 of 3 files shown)


cfsmp3:
How does this PR check if the output file is going to be empty or not? 
The goal is to open the file for writing only when actual subtitles are going to be written to disk, but you don't know that until the file is actually being processed.
What happens if you apply your patch and run ccextractor to process a file that does have subtitles?
What if you process a file that does not have subtitles?

Izaron:
> How does this PR check if the output file is going to be empty or not?

I check if the result file size is [3 bytes](https://github.com/CCExtractor/ccextractor/pull/458/files#diff-c0df1fe2367db8e8475c453bdefdc269R427) or less, I delete the file. [Why 3 bytes?](https://github.com/CCExtractor/ccextractor/issues/260#issuecomment-259093468) But in fact, we can read the file and verify that it contains not only whitespace. It will be safer.

>  The goal is to open the file for writing only when actual subtitles are going to be written to disk, but you don't know that until the file is actually being processed.

The result file is checked at the end of main(), after all the work with files has passed.

>  What happens if you apply your patch and run ccextractor to process a file that does have subtitles?

Subtitles will be successfully saved because the output file size will be more than 3 bytes.

> What if you process a file that does not have subtitles?

I tested many files. With flag --noempty: while the program is running, an empty 3-byte file is created, but after work it is removed.

cfsmp3:
Can't merge because it does create an empty file even if it deletes it later which is not the same thing - the creation itself can trigger external actions (for example a program could be monitoring if the file is created at all).
Still, it's a very good first attempt at this.

PR #460, opened 2016-12-01
GCI task: Return a non-zero return code if no subtitles were found (CCExtractor Development)

yodadude2003:
added error for when opened file is blank.


yodadude2003:
I noticed that in all of the pull requests so far, the same two tests are failing every time
Please check this out

PR #469, opened 2016-12-02
Added basic ASS/SSA encoding support

Izaron:
[https://codein.withgoogle.com/dashboard/task-instances/5110381871628288/](https://codein.withgoogle.com/dashboard/task-instances/5110381871628288/)

Added basic ASS/SSA encoding support.
I checked some files, works well. Run with args: `file -out=ssa` or `file -out=ass`.
`src/lib_ccx/ccx_encoders_ssa.c` has two empty functions, as I can't check it without example video file, on which this functions will be called. Please give me test file if you can.
`use_ass_instead_of_ssa` there is, as the format is slightly different, so it is possible to generate a little different files because of different formats .ass and .ssa, but maybe we need only one format.
Please give your opinion about the `ssa_header`, maybe there is something to change in style.

### DIFF
--- a/src/lib_ccx/ccx_common_constants.h
+++ b/src/lib_ccx/ccx_common_constants.h
@@ -170,6 +170,7 @@ enum ccx_output_format
 	CCX_OF_SIMPLE_XML = 10,
 	CCX_OF_G608 = 11,
 	CCX_OF_CURL = 12,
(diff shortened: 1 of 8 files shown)


Izaron:
Closed due to a bug, opened new [https://github.com/CCExtractor/ccextractor/pull/470](url)

PR #476, opened 2016-12-03
Added two functions for ASS/SSA encoder

Izaron:
Added functions to write subs. Other encoders were very helpful.


DanilaFe:
Amusingly, the test for this pr breaks because of the issue you fixed.

PR #479, opened 2016-12-05
Fix build on Windows.

DanilaFe:


DanilaFe:
@Izaron implemented the new SSA encoders (or started, for that matter) in #476 and #470, but likely did not attempt to compile the Windows build - I didn't either. In attempt to find my own build issues, I tried to compile the up-to-date code in a windows machine, and got a linker error:
```
unresolved external symbol _write_cc_buffer_as_ssa
unresolved external symbol _write_cc_subtitle_as_ssa
unresolved external symbol _write_stringz_as_ssa
unresolved external symbol _write_cc_bitmap_as_ssa
```
This was because Visual Studio's file list did not include the new encoders.c file. This PR addresses that.

Izaron:
Not bad, but should this vcxproj files be in gitignore, not in project?

DanilaFe:
I wouldn't assume so, because it appeared to be necessary to build on windows (to open to project in visual studio). However, I have limited experience with windows, so correct me if I'm wrong. I assume the vcxproj files are similar to the build scripts for Linux and macOS in that they contain instructions on what files to compile and how.

canihavesomecoffee:
@Izaron They are necessary for "regular" users to be able to build the project on windows. However, we can't require all contributors to use windows, so it's up to the persons who build on windows (including myself) to fix the build every now and then.

DanilaFe:
So is this PR sufficient to fix the build?

cfsmp3:
Yes. I closed it because my local copy already had the same thing so I
committed my changes.

PR #503, opened 2016-12-09
Add VC.db file to gitignore

canihavesomecoffee:
Also moves the folders to the correct place


cfsmp3:
Can't merge (conflicts)

PR #505, opened 2016-12-10
Fixed iTunes subtitle duplication

### LINKED ISSUE #427: Duplicating lines from iTunes
I have this problem periodically and none of the settings seem to help. When extracting a subtitle from an iTunes file, the resulting .srt file has duplicated lines as shown in the upload at: https://www.dropbox.com/s/hvx129ih4bu5pxb/03%20He%20Calls%20Himself%20Daxis%20%281080p%20HD%29.m4v?dl=0

Subs: https://www.dropbox.com/s/yios5l1lifxhe67/03%20He%20Calls%20Himself%20Daxis%20%281080p%20HD%29.srt?dl=0

I assumed it had something to do with roll-up captions, but those settings don't seem to change the output.

Settings used: C:\Program Files (x86)\CCExtractor\ccextractorwin.exe  --gui_mode_reports  -autoprogram  -out=srt -ru3 -bom -utf8 -trim -autodash --norollup [+input files]

Thanks for your help!

AlexBratosin2001:
Closes #427

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -85,8 +85,8 @@ int main(int argc, char *argv[])
 #ifdef WITH_LIBCURL
 	curl_global_init(CURL_GLOBAL_ALL);
  
-  	/* get a curl handle */ 
-  	curl = curl_easy_init();
+	/* get a curl handle */ 
+	curl = curl_easy_init();
 	if (!curl)
 	{
 		curl_global_cleanup(); // Must be done even on init fail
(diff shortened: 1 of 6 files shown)


cfsmp3:
Can you submit only the file related to this bug, which seems to be  src/gpacmp4/mp4.c? The other ones don't belong to this PR.

Great job here by the way. As often happens, a one liner fixes the problem.

cfsmp3:
Um, I'm looking more at the change and it doesn't make much sense? 
len is the amount of bytes in the buffer
ret is the bytes processed in process608(), which is actually the same amount of bytes passed in the 2nd parameter.
So substracting ret from len after return is correct, and also seems correct to continue processing while len>0 (meaning there's still unprocessing bytes).

cfsmp3:
Solved already, discussed in slack.

Issue #431, opened 2016-10-03
fps switching in HISTORY channel


cfsmp3:
Processing a HISTORY (writing it in uppercase because it seems to be a big deal to them) TS we get this:

 67%  |  00:16Changed fps using NAL to: 29.970030
 72%  |  00:18Changed fps using NAL to: 29.970030
repeated lots of times

A sample file:
https://drive.google.com/file/d/0BxRlGXOr_NhUNmh0NXg3S2dHeFE/view?usp=sharing

The FPS changes to that value here:

void seq_parameter_set_rbsp (struct avc_ctx *ctx, unsigned char *seqbuf, unsigned char *seqend)
{
[...]
                if (current_fps != (double)time_scale / (2 \* num_units_in_tick))
                {
                    current_fps = (double)time_scale / (2 \* num_units_in_tick); // Based on formula D-2, p. 359 of the ISO/IEC 14496-10:2012(E) spec.
                    mprint("Changed fps using NAL to: %f\n", current_fps);
                }
            }

However it's going back to double that FPS (however it's not logged) here:

void slice_header (struct lib_cc_decode *ctx, unsigned char *heabuf, unsigned char *heaend, int nal_unit_type, struct cc_subtitle *sub)
{
[...]
            // TODO - Do this right.
            // When bottom_field_flag is set the video is interlaced,
            // override current_fps.
            current_fps = framerates_values[7];

Where that TODO comment really seems to apply.

mackworth:
I'm seeing this on multiple channels on Comcast.

cfsmp3:
I've been looking into this a bit more. The issue is probably more serious than it seems.

- The variable current_fps is global, not part of the context. This alone is probably very wrong since we may be processing both interlaced and non-interlaced streams at the same time if they are multiplexed. Probably even at different FPS, even though that's likely just a theoretical case.
- The line current_fps = framerates_values[7]; which is the same as current_fps=29.97*2 is definitely wrong as it pretty much means our timing is always going to be wrong when processing European AVC streams (well, unless it's working by pure luck due to the code that actually sets the fps to the right value as mentioned in the initial post).

I'm assigning this to Anshul (since he did the context refactoring - Anshul, do you remember why current_fps was left as a global?) and myself.

cfsmp3:
Code-in task created.

grave-w-grave:
Error is still here as of 0.82. Number of times it appears is equal to the number of jump-in frames.

[...]

PR #510, opened 2016-12-10
Fixed FPS switching

Izaron:
Fixed FPS switching
[Issue #431 ](https://github.com/CCExtractor/ccextractor/issues/431)
I tested it also on another video, which also had this bug. See this value write [https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/es_functions.c#L435](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/es_functions.c#L435) I did the same and it works


canihavesomecoffee:
The results show very minor changes in timing for some samples, so I don't see a problem with this PR! 👍

PR #526, opened 2016-12-12
Fixed -unixts option

Izaron:
Issue #430 

default

This is not a bug - this function is **only** realised for Teletext files
I radically enabled it for all subtitles
Checked on the .bin file, works well

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -1080,7 +1080,8 @@ int encode_sub(struct encoder_ctx *context, struct cc_subtitle *sub)
 
 				if (utc_refvalue != UINT64_MAX)
 				{
-					data->start_time += utc_refvalue * 1000;
+					if (data->start_time != -1)
+						data->start_time += utc_refvalue * 1000;
 					data->end_time += utc_refvalue * 1000;
 				}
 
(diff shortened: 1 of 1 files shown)


cfsmp3:
This is touching a really sensitive thing :-) have you checked it doesn't break anything when -unixts is not used? What about other formats?

Izaron:
Sorry, I had to explain my actions.
Initially, the variable `utc_refvalue` is equals to `UINT64_MAX`
If we use `-unixts,` then we change this value - see line [https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/params.c#L1943](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/params.c#L1943)

In fact, the assignment is performed in another piece of code - see line [https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/telxcc.c#L1271](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/telxcc.c#L1271), but we do not break the program, because `opt->noautotimeref= 1` after `-unixts` and we will not perform the second assignment

In teletext uses its own variable to correct subtitle time - [https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/telxcc.c#L127](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/telxcc.c#L127)

I've included it for all formats

PR #534, opened 2016-12-14
Fixed linux/build script

Izaron:
Checks for the presence of the dependencies.
Tested on Ubuntu, Debian, Fedora.
First in the variable output of the command recorded, then checked for errors.
I covered all possible dependencies.

### DIFF
--- a/linux/build
+++ b/linux/build
@@ -14,7 +14,7 @@ BLD_LINKER="-lm -zmuldefs -l tesseract -l lept -lcurl"
 
 ./pre-build.sh
 
-out=$((gcc $BLD_FLAGS $BLD_INCLUDE -o ccextractor $BLD_SOURCES $BLD_LINKER) 2>&1)
+out=$((LC_ALL=C gcc $BLD_FLAGS $BLD_INCLUDE -o ccextractor $BLD_SOURCES $BLD_LINKER) 2>&1)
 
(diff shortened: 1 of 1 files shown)


Izaron:
Added LC_ALL=C
It works in virtual boxes
http://unix.stackexchange.com/questions/87745/what-does-lc-all-c-do
http://stackoverflow.com/questions/869309/selecting-message-language-in-gcc-and-g
(And other proofs what LC_ALL works)

linux/build
```diff
@@ -13,4 +13,31 @@ BLD_SOURCES="../src/ccextractor.c $SRC_CCX $SRC_GPAC $SRC_ZLIB $SRC_ZVBI $SRC_LI
...
+    echo "Error: please install tesseract development library (tesseract-ocr-dev for Debian/Ubuntu)";
+    exit 1
+fi
+
+if [[ $out == *"allheaders.h: No such file or directory"* ]]
+then
+    echo "Error: please install leptonica development library (libleptonica-dev for Debian/Ubuntu)";
+    exit 1
+fi
+
```
canihavesomecoffee: What if something else fails? User will have no output, as it's all captured in $out...

Izaron:
Thanks to @canihavesomecoffee for help!
I tested it in Debian, Ubuntu, Fedora, openSUSE, everything works well!

PR #527, opened 2016-12-12
return a non-zero return code if no subtitles were found

anukul:
WIP
i'm making this pull request right now to discuss what would be the best way to go about the changes.

PR #535, opened 2016-12-15
return a non-zero return code if no subtitles were found

anukul:
Previous PR: https://github.com/CCExtractor/ccextractor/pull/527

### DIFF
--- a/src/lib_ccx/general_loop.c
+++ b/src/lib_ccx/general_loop.c
@@ -523,13 +523,14 @@ void process_hex (struct lib_ccx_ctx *ctx, char *filename)
 }
 #endif
 // Raw file process
-void raw_loop (struct lib_ccx_ctx *ctx)
+int raw_loop (struct lib_ccx_ctx *ctx)
 {
 	LLONG ret;
 	struct demuxer_data *data = NULL;
 	struct cc_subtitle *dec_sub = NULL;
 	struct encoder_ctx *enc_ctx = update_encoder_list(ctx);
 	struct lib_cc_decode *dec_ctx = NULL;
+	int caps = 0;
 
 	dec_ctx = update_decoder_list(ctx);
 	dec_sub = &dec_ctx->dec_sub;
@@ -549,6 +550,7 @@ void raw_loop (struct lib_ccx_ctx *ctx)
 		ret = process_raw(dec_ctx, dec_sub, data->buffer, data->len);
 		if (dec_sub->got_output)
 		{
+			caps = 1;
 			encode_sub(enc_ctx, dec_sub);
 			dec_sub->got_output = 0;
 		}
@@ -559,6 +561,7 @@ void raw_loop (struct lib_ccx_ctx *ctx)
 
 
 	} while (data->len);
+	return caps;
 }
 
 /* Process inbuf bytes in buffer holding raw caption data (three byte packets, the first being the field).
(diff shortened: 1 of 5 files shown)


src/lib_ccx/general_loop.c
```diff
@@ -975,7 +977,6 @@ int general_loop(struct lib_ccx_ctx *ctx)
 		if (ccx_options.send_to_srv)
 			net_check_conn();
 	}
-	caps = dec_ctx->saw_caption_block;
```
anukul: This failed some tests

anukul:
I'm working on it and the sample platform CI is a big help. If possible, please don't close it just yet.

cfsmp3:
A problem here is that if there's no captions fatal() is called. But fatal terminates rather quickly. For example if there's more than 1 input file the program will stop without processing all of them. However the correct behaviour would be to report success if there was any subtitle generated and fail if there were no subtitles at all.

So possibly having a flag somewhere is better than trying to do it only with function return values.

anukul:
@cfsmp3 Could you check now?
I guess its passing the tests. It says "Fail" because of the unexpected non-zero return code.

canihavesomecoffee:
There's definitely something still off.

Take a look at the teletext samples. I'm sure that most (if not all of them) have output. Yet the return code is 10 all the time.

anukul:
Ok the last commit improves some of it.
Does the word "Pass" at the end not denote that it works fine on that particular test?
In some, it says Pass for each individual test but Fail for the category as a whole.

image

canihavesomecoffee:
A category fails if:

- A test has a non-0 exit code
- A test has a different output
- A test has a file that has been ignored

The result however that's listed per sample is simply determined by the different output. Hence the difference between the "pass" for tests vs "fail" for category.

In the case of your screenshot (with the NoCC category), the category is marked as failed because there was >= 1 test with non-zero exit code, but the tests mark as passed because the output of them (empty file expected) is "correct".

anukul:
What other problems exist now?

cfsmp3:
@anukul try processing two input files, the first one without subtitles and the second one with them and you'll see. You cannot fatal() on the condition of no subtitles. For starters because that's not fatal at all.

anukul:
Okay. I found my mistake there.
I've tried to fix it (in file ccextractor.c). https://github.com/anukul/ccextractor/commit/73af3c677fd1b867a55822bc777e5472fbfe8536

Also, while processing multiple input files, I found some weird behavior.

Let inp.bin denote a file with captions
and nocaps.bin denote a file without captions

Processing inp.bin generates a subtitle file of 45 lines.
Process nocaps.bin generates an empty subtitle file.
Processing inp.bin, nocaps.bin generates a subtitle file of 45 lines.
Processing nocaps.bin, inp.bin generates a subtitle file of 5 lines.

Here are their outputs: https://www.dropbox.com/sh/5sgvvgtagbavfzl/AAABVYNSLvd0si6ldac7hVUba?dl=0

If you'll reopen the PR, I'll try to see if my last commit breaks anything in the tests on the sample platform.

cfsmp3:
It's always the same thing. You set ret to 0 at the start of the loop, then change it to whatever the calls to general_loop() returns. So if you have 2 files and the first one has subtitles but the second one doesn't have subtitles then ret is going to have the value of the second pass not taking into account that the first file had subtitles.

anukul:
Not really.
I do this instead:
image

PR #540, opened 2016-12-15
Propose the use of clang-format.

DanilaFe:
Clang-format is an automatic code formatting tool.

It would greatly ease the creation of consistent code: it follows exactly the rules in the .clang-format file. Style changes would be as easily applied as editing the .clang-format file and re-running the clang-format command:
```Bash
clang-format src/lib_ccx/*.c src/lib_ccx/*.h src/ccextractor.c -i
```

The file that I'm proposing is a draft that can be easily customized by the maintainers to fit the code style they wanted for the project: I attempted to configure it such that it was similar to what the style appeared to be before.

`clang-format` is definitely a command line tool that works on macOS and Ubuntu, and to the best of my knowledge, it's also available as a Visual Studio plugin on Windows.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,6 @@
+UseTab: Always
+IndentWidth: 8
+BreakBeforeBraces: Allman
+AllowShortIfStatementsOnASingleLine: false
+IndentCaseLabels: false
+ColumnLimit: 80
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,4 @@
 UseTab: Always
-IndentWidth: 8
 BreakBeforeBraces: Allman
 AllowShortIfStatementsOnASingleLine: false
 IndentCaseLabels: false


cfsmp3:
Ah but maintainers don't want to customize themselves, we'd like to see the result of several options and then pick one :-)

DanilaFe:
That's possible too. I didn't run the tool because changing formatting will likely cause pointless conflicts.

DanilaFe:
How can I demonstrate the result of clang-format, other than by committing formatted files?

cfsmp3:
Yo can upload here a .zip file with contains the different versions of one
file you find significant :-)

DanilaFe:
[Here](https://drive.google.com/open?id=0B773-D8RjVQyOVNLOGRENjJQTEE) is a link to a zip file containing three different .clang-format files and the resulting ccextractor.c file.

The best-match style was written to try to match the current code style of the source code.
The linux and vs styles imitate the styles of the Linux Kernel and Visual Studio's formatting, respectively.

cfsmp3:
I'm not going to merge because I'm quite torn here...  I guess working on Python lately has made me switch my preference from 8 spaces to 4 for indents (which means no tab, which is not really C style).

But I liked seeing how this work, so it's something to revisit possibly during summer of code, after all the team has spent a few weeks on C :-)

PR #545, opened 2016-12-16
Added configurable curl post url

Izaron:
To compile with WITH_LIBCURL add in linux/build "-DWITH_LIBCURL" in BLD_FLAGS and "-l curl" in BLD_LINKER.

After code research, I added support for configurable cURL post URL. Already implemented options is "-out=curl" and "-curlposturl [URL]".
It was quite simple, because I used a global variable `ccx_options`.

But there are problem areas. My suggestions:

1. How much reasonable this line in code architecture? [Line](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/params.c#L2246). The same variable belongs to two objects, one of which belongs to another.
2. It may be worthwhile to make the default curl post url? Instead of [line](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/params.c#L2212). Maybe we need to look at the most frequent URL in tutorials about local servers.
3. In one of my example videos after ``` Done, processing time = 44 seconds This is beta software. Report issues to carlos at ccextractor org... ```
I get `*** Error in './ccextractor': free(): invalid pointer: 0x00007f49893e9c18 ***` and backtrace. And it happens in current master. I can explore it if you want. Sample video - [Link](https://drive.google.com/file/d/0B0DIrRkpdn12MmZBckw0N3cwNnc/view)

Please, write what you think about it, thanks!

### DIFF
--- a/src/lib_ccx/ccx_common_option.h
+++ b/src/lib_ccx/ccx_common_option.h
@@ -46,7 +46,9 @@ struct encoder_cfg
 	int trim_subs; // "    Remove spaces at sides?    "
 	int sentence_cap ; // FIX CASE? = Fix case?
 	int splitbysentence; // Split text into complete sentences and prorate time?
+#ifdef WITH_LIBCURL
 	char *curlposturl; // If out=curl, where do we send the data to?
+#endif
 
 	int with_semaphore; // Write a .sem file on file open and delete it on close?
 	/* Credit stuff */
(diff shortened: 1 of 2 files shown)


cfsmp3:
"I get *** Error in './ccextractor': free(): invalid pointer: 0x00007f49893e9c18 *** and backtrace. And it happens in current master. I can explore it if you want."

Yes, please explore and fix :-)

PR #547, opened 2016-12-16
Fixed -udp ignoring interface

Izaron:
Use netstat for checking: `sudo netstat -ntupl | grep ccextractor`
We must use our address instead of INADDR_ANY

### DIFF
--- a/src/lib_ccx/networking.c
+++ b/src/lib_ccx/networking.c
@@ -976,12 +976,7 @@ int start_upd_srv(const char *addr_str, unsigned port)
 	struct sockaddr_in servaddr;
 	servaddr.sin_family = AF_INET;
 	servaddr.sin_port = htons(port);
-#ifndef _WIN32
-	if (IN_MULTICAST(addr))
-		servaddr.sin_addr.s_addr = htonl(addr);
-	else
-#endif
-		servaddr.sin_addr.s_addr = htonl(INADDR_ANY);
+	servaddr.sin_addr.s_addr = htonl(addr);
 
 	if (bind(sockfd, (struct sockaddr *)&servaddr, sizeof(servaddr)) != 0)
 	{


cfsmp3:
This is going to break things I think... the check for multicast is necessary. Not the same thing to listen to a multicast stream than to a unicast one. I remember I had to fight that a bit.

Izaron:
But if we will use `INADDR_ANY` instead of `addr`, then in next line we will bind all interfaces - [line](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/networking.c#L986) And will get 0 0.0.0.0:7701 instead of 127.0.0.1:7701
And with "-udp 127.0.0.1:7701" we will always get that `addr`, for which `IN_MULTICAST(addr)` equals to false - [line](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/networking.c#L946). How we can deal with it if code for this is hidden in other library?

cfsmp3:
I remember this stuff used to work, then stopped working, then back again. I'm looking into past emails about this. One of them says:

In file_functions.cpp
                servaddr.sin_addr.s_addr=htonl(INADDR_ANY);
should be
                servaddr.sin_addr.s_addr=htonl(IN_MULTICAST(udpaddr) ? htonl(udpaddr) : INADDR_ANY);

(don't mind the file name - this is a very old email, before networking.c existed).

Another email says:

In line 147 of file_functions.cpp the code reads
servaddr.sin_addr.s_addr=htonl(IN_MULTICAST(udpaddr) ? htonl(udpaddr) : INADDR_ANY);

It should read
servaddr.sin_addr.s_addr=htonl(IN_MULTICAST(udpaddr) ?       udpaddr  : INADDR_ANY);

And other emails (mine) say that in Windows it doesn't work unless... but in linux unless... :-) 
Are you able to test all 4 variants (multicast and unicast, Windows and Linux) ?

Izaron:
Ok, I think I can do it tomorrow after some theory about casting (diffs between uni and multi) and sockets code, and testing :)

PR #548, opened 2016-12-17
Make code more legible by aligning columns, comments, etc.

ngxingyu:


cfsmp3:
Comments not aligned for example in src/lib_ccx/asf_constants.h  (just see diff in github), some changes such as removing indentation in comments for structure members doesn't make sense...  when comments are in a separate line that line should be properly aligned with the variable they refer to, not go back to the start of the line.

PR #551, opened 2016-12-18
Make code more legible

ngxingyu:


cfsmp3:
Alignment not correct at least in some files, such as ccx_common_option.h.

PR #560, opened 2016-12-21
GCI Task(Make code more legible by aligning columns, comments, etc )(JayGupta2003)

JayGupta2003:
Found a minor mistake in this file.


cfsmp3:
What does this PR do? On first inspection (in github) it rewrites the whole file but changes nothing visually.

DanilaFe:
It seems to change the indentation of a function declaration on line 48, but that's all I can see. If my memory isn't failing me, the rewrite might be caused by a change in file permissions (something like chmod) that might run accidentally on a different OS which has different file permissions by default.

PR #566, opened 2016-12-23
Added basic support of Tesseract 4.0.0

Izaron:
I've added support for built Tesseract 4.0.0 in Windows - http://www.ccextractor.org/doku.php?id=public:general:tess_build
It was necessary only to update the code in one place - in hardsubx_classifier.c, so what works well both on Linux and Windows
Also I updated old zlip and libpng (dependencies of tesseract) to avoid errors

**Known issues:**
1. Built executables requires .dll files with names starts with "pvt.cppan.demo." (different for Debug and Release). I have not figured out yet how to fix it. [Download DLL](https://drive.google.com/open?id=0B2fcX80_rHuOMFJTYnlMWEZVRW8) Also, while it is not fixed, I do not copy DLL files after build in folders

I'll fix this issues in later commits


Izaron:
Disabled info messages as it been earlier

PR #567, opened 2016-12-23
Add --noempty option not to generate output file if it's going to be empty (GCI)

gonzaloUran:


src/gpacmp4/mp4.c
```diff
@@ -330,6 +330,14 @@ int processmp4 (struct lib_ccx_ctx *ctx,struct ccx_s_mp4Cfg *cfg, char *file)
 
 	mprint("mp4: found %u tracks: %u avc and %u cc\n", track_count, avc_track_count, cc_track_count);
 
+	if(ccx_options.noempty)
+	{
+		if(cc_track_count)
+			enc_ctx = update_encoder_list(ctx);
+		else
+			exit(0);
```
cfsmp3: We cannot do this exit(0)... no explanation to the user at all, plus what if the user was processing more than one .mp4 and the next goes does have captions?

cfsmp3:
Reopen if you fix.

PR #571, opened 2016-12-25
Fix SubStation Alpha subtitles in bitmap

Izaron:
When I created the decoder, I could not check function write_cc_bitmap_as_ssa with ENABLE_OCR, and then I fixed one bug - we should change newline symbols in SSA to "\\N"

### DIFF
--- a/src/lib_ccx/ccx_encoders_ssa.c
+++ b/src/lib_ccx/ccx_encoders_ssa.c
@@ -111,6 +111,17 @@ int write_cc_bitmap_as_ssa(struct cc_subtitle *sub, struct encoder_ctx *context)
 	str = paraof_ocrtext(sub, context->encoded_crlf, context->encoded_crlf_length);
 	if (str)
 	{
+		// SSA format - change "\r\n" to "\N"
+		len = strlen(str);
+		for (int i = 0; i < len - 1; i++)
+		{
+			if (str[i] == '\r' && str[i + 1] == '\n')
+			{
+				str[i] = '\\';
+				str[i + 1] = 'N';
+				i++;
+			}
+		}
 		if (context->prev_start != -1 || !(sub->flags & SUB_EOD_MARKER))
 		{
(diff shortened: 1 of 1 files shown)


cfsmp3:
Why \N instead of \n?

Izaron:
Because we have to write SSA/ASS subs in one line per one subtitle and instead of symbol '\n' we have to write two symbols '\' and 'N', as you can see in the sample .ssa files in Internet

cfsmp3:
Don't know... SSA specs (
https://www.matroska.org/technical/specs/subtitles/ssa.html):

*\n*                              New line (carriage return)
                                 \n is ignored by SSA if “smart-wrapping”
is enabled



*eg. This is the first line\nand this is the second
\N                              *New line (carriage return). This is used
by SSA instead of \n if

                                 “smart-wrapping” is enabled.


Do we have smart-wrapping? :-)

Izaron:
Unfortunately we anyway have to write every single "Dialogue" in single line
```
Field 10:	Text
Subtitle Text. This is the actual text which will be displayed as a subtitle onscreen. Everything after the 9th comma is treated as the subtitle text, so it can include commas.

			The text can include \n codes which is a line break, and can include Style Override control codes, which appear between braces { }.

```

We can write text field as
```
Hello\nWorld
```
Or with **\N** but not as:
```
Hello
World
```

So it must be
```
Dialogue: 0,00:00:18.56,00:00:23.64,Default,,0000,0000,0000,,But this seemingly infinite\Ncomplexity
```
(Or with \n as two symbols) But must not be
```
Dialogue: 0,00:00:18.56,00:00:23.64,Default,,0000,0000,0000,,But this seemingly infinite
complexity
```

Izaron:
Although I would prefer to replace the one character '\n' instead of two '\r\n', but I do not know yet how to insert a new character in char* in one line (because '\\' and 'N' are two characters).

PR #574, opened 2016-12-26
Added build script for .deb

Izaron:
default
(sorry for non-english)

This command build the .deb file in folder package_creators/build, and you can install by:
```
sudo dpkg -i %built_filename%.deb
sudo apt-get -y -f install
```
(Last command for dependencies)
And then you can run `ccextractor` in console in any folder of your system.

### DIFF
--- a/linux/description-pak
+++ b/linux/description-pak
@@ -1,0 +1,1 @@
+Best open source tool for a subtitled world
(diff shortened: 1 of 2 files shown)


cfsmp3:
I get checkinstall not found - so some error checking is needed. 
Also I'm reading the checkinstall docs and it seems to require that we have done a make install prior to calling it... (if I understood things right).  But this is not checked either is, it?

Izaron:
@cfsmp3 Just forgot to say this: We must firstly go to folder linux and run ./build, and then go to  package_creators and run debian.sh with sudo

Izaron:
Added warning if there is no installed 'checkinstall'. In fact, we do not need to run the ./build, because the program does it for us (the main thing that we have to create linux/objs folder instead of get error)

PR #579, opened 2016-12-29
Fixed memory leak for DVB subs.

AlexBratosin2001:


Izaron:
If you will just write "x = NULL", this does not mean that the object is freed.
Watch how I catch leaks, at this PR - https://github.com/CCExtractor/ccextractor/pull/478

AlexBratosin2001:
Read again. I'm setting the pointer to NULL in the initialization of the decoder. This is to avoid a crash when I don't actually process a file with DVB subtitles because in the dinit_libraries, I wouldn't want to free a non-initialized pointer. Check again what I'm doing.

Izaron:
Understand your actions after conferencion in Slack, I'm sorry, everything is very clean.

PR #591, opened 2016-12-31
Fixed SAMI unnecessary empty subtitle when extracting DVB subs.

AlexBratosin2001:
Basic continuity check.


canihavesomecoffee:
Could you double-check on sample 225 (https://sampleplatform.ccextractor.org/regression/test/225/view) ?

canihavesomecoffee:
Looks fine to me.

PR #599, opened 2017-01-03
C608 to WebVTT translation

Izaron:
Work in progress - translating CEA-608 subs to WebVTT

**Done:**

- Support of the raw old format without colors
- Added colors and fonts with segments (consecutive characters of the same font or color)

**Todo:**

- Replace current "webvtt" to "webvtt-raw" and "webvtt-full" to "webvtt" (???)
- Add XDS [metadata](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#metadata-xds) what **already** computed by CCExtractor and written on screen; and add it in file
- True positioning as in document
- Anything else missing from the [document](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html)

@cfsmp3 Write your suggestions

### DIFF
--- a/src/lib_ccx/ccx_common_option.h
+++ b/src/lib_ccx/ccx_common_option.h
@@ -143,6 +143,7 @@ struct ccx_s_options // Options from user parameters
 	unsigned send_to_srv;
 	enum ccx_output_format write_format;                // 0=Raw, 1=srt, 2=SMI
 	int use_ass_instead_of_ssa;
+	int use_webvtt_styling;
 	LLONG debug_mask;                                   // dbg_print will use this mask to print or ignore different types
 	LLONG debug_mask_on_debug;                          // If we're using temp_debug to enable/disable debug "live", this is the mask when temp_debug=1
 	/* Networking */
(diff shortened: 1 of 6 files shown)


cfsmp3:
Aren't we missing the actual styling here? CSS or something :-)

Izaron:
Don't sure that there was CSS style need for CEA-608 (any hint???), by "Styling" I mean support of text color and fonts in output file in proper format. CSS styling is the post-CCExtractor work :)

cfsmp3:
This what the document says:

"When a WebVTT file is created from a CEA-608 or CEA-708 caption data
stream using the conversion methods proposed in this document, it is
suggested to use the following header for the WebVTT file, including
an identification of the type of data contained:

WEBVTT
Style:
@import(cea608.css)
##
Kind: captions
Language: en-US
 "

The reason of course is that by adding such style the player can
render the subtitles so they look exactly as they look on a TV
(assuming everything else explained in the document is done, such as
positioning, etc).

Issue #452, opened 2016-11-30
Ripping captions from WTV files ignores all time stamps and commercials


Silver-Streak:
I have a HTPC I built running Windows 8Windows Media Center. I use MCEBuddy to clean up my recorded WTV files, ripping captions to SRT for my wife, who is deaf and wants to be able to watch the videos throughout the house, and creating edl (I store for later to try and tweak appropriately). 

I noticed that all of a sudden all of my subtitles are WAY out of sync. Like, minutes. Doing some digging, it happened in every file I've ripped CC from since I upgrade MCEBuddy. Digging into that, I saw they updated their version of CC extractor. The previous version of MCEBuddy I had installed (2.3.5) had CCExtractor 0.66,, where as the newer version of MCEBuddy had CCExtractor 0.81.

Looking at output files, it looks like what is happening is that the newly output captioned file act like the show starts at 0:00:00, which is not true, due to commercial/channel logos.  Older captured files (from 0.66, I believe) include the commercials themselves, and seem to be fully in sync. The new files don't appear to have captions at all for the commercials that have them, and timestamps are off by minutes in some cases. 

I've eliminated MCEBuddy from the equation, and have been able to reproduce this behavior on a separate machine with 0.81 vs 0.66. I've attached what each version spits out. Any insight into what I'm doing wrong, or what could have changed, is greatly appreciated.

Created previously
[062116 Arrow_S03E18 - Copysrt.txt](https://github.com/CCExtractor/ccextractor/files/620795/062116.Arrow_S03E18.-.Copysrt.txt)

Created and recreated yesterday
[112816 The Flash (2014)_S03E05 - Copysrt.txt](https://github.com/CCExtractor/ccextractor/files/620796/112816.The.Flash.2014._S03E05.-.Copysrt.txt)

For reference, the first line in the video file this is ripped from doesn't start until 1 minute, 41 seconds.

cfsmp3:
Sounds like you are not using the subtitle file on the same video file it
was extracted from? The subtitles are in sync with the source file, but if
you remove clips from that file and generate a new one with missing parts
of course the subtitles are going to be totally off sync.

Silver-Streak:
To clarify, these files are being used with the exact same files they're ripped from. Previously, I had been automatically converting these files to MKV and leaving the subtitles external files. To provide additional clarity, these are being recorded via ATSC over the air broadcasts.

[...]

PR #605, opened 2017-01-06
Fix for bad WTV timings (#452)

brooss:
This quick fix should correct the change introduced in dd8923b2e4b4af08b3322811af4f8d7aad953e0e that broke (some?) WTV timings. Reported in #452.


canihavesomecoffee:
Surprisingly we don't have a WTV test that suffered from that issue... Think I'm going to add the sample that was provided in the issue asap to the platform then.

PR is fine to me, CI doesn't indicate anything else that's broken except for the already failing ones.

PR #606, opened 2017-01-06
A pointer was freed twice in file

sidgairo18:
Double Free
Same pointer freed twice gf_list_del(cfg->pictureParameterSets) in lines 550 and 559


canihavesomecoffee:
It might be best to report this to the authors of GPACMP4 too...

PR #623, opened 2017-01-10
Fixes in CEA-708

### LINKED ISSUE #477: Here's a quandary: 608 captions in Korean (redux)
Since the last issue [https://github.com/CCExtractor/ccextractor/issues/409](url) was closed (I've been a bit busy lately) the comment asked for the TS file and not the BIN file. I'll attach Dropbox links to the TS file as well as the BIN file. I don't believe the problem is with the extractor, but with the 608 decoder, as legible caption data was extracted, but not converted to SRT.

BIN: [https://www.dropbox.com/s/g72aaw3txzmto8a/DancingQueen.bin?dl=0](url)
TS: [https://www.dropbox.com/s/xvq99nyhj4rymlp/DancingQueen.ts?dl=0](url)

Izaron:
#Getting started on improving
Closes #477 - result file [DancingQueen.p2.svc01.srt ](https://gist.github.com/Izaron/44c030eae8c6c1049ae6d3e6c6d0dd32)
There are 2 other issues in the list with Korean subtitles, but I do not know whether it is possible to close them
Now correctly displayed characters like ® 
There still characters missing in some lines in English subtitles

### DIFF
--- a/src/lib_ccx/ccx_decoders_708.c
+++ b/src/lib_ccx/ccx_decoders_708.c
@@ -1068,8 +1068,8 @@ void dtvcc_handle_SPA_SetPenAttributes(ccx_dtvcc_service_decoder *decoder, unsig
 	int text_tag  = (data[1] >> 4) & 0xf;
 	int font_tag  = (data[2]     ) & 0x7;
 	int edge_type = (data[2] >> 3) & 0x7;
-	int underline = (data[2] >> 4) & 0x1;
-	int italic    = (data[2] >> 5) & 0x1;
+	int underline = (data[2] >> 6) & 0x1;
+	int italic    = (data[2] >> 7) & 0x1;
 
 	ccx_common_logging.debug_ftn(CCX_DMT_708, "       Pen size: [%d]     Offset: [%d]  Text tag: [%d]   Font tag: [%d]\n",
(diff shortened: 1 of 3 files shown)


cfsmp3:
for (int i = last + 1; i < CCX_DECODER_608_SCREEN_WIDTH; i++)

That doesn't seem right at all... how can be use a 608 constant for 708? They're totally different things.

Izaron:
Both 608 and 708 have at max 32 symbols in a row, CCX_DECODER_608_SCREEN_WIDTH == 32

cfsmp3:
Nope :-) 

From Wikipedia:
column count, 6 bits. This is the number of columns of text, assuming the STANDARD font size, the window will hold. The range is 0-31 for 4:3 streams, and 0-41 for 16:9 streams. NOTE: In practice a decoder must add one to the number to get the intended effect. i.e. 0 -> 1, 1 -> 2, etc.

Izaron:
- Fixed column count for whitespace filling (all changes only in SRT yet)
- Fixed italics and underline - there was bug with shifting, see in https://en.wikipedia.org/wiki/CEA-708 **SetPenAttributes (0x90 + 2 bytes)**

PR #637, opened 2017-01-12
make -ignoreptsjumps and -dvbcolor default

gonzaloUran:


src/lib_ccx/params.c
```diff
@@ -990,6 +990,10 @@ int atoi_hex (char *s)
 
 int parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
 {
+	//default arguments
+	opt->dvbcolor = 1;
+	opt->ignore_pts_jumps = 1;
+
```
saurabhshri: Maybe a better place to make this change : https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_common_option.c#L63

gonzaloUran: Yes, that's more appropriate.

Issue #648, opened 2017-01-15
Memory Leaks found.


sidgairo18:
1. Leaked_storage: Variable `el` going out of scope leaks the storage it points to. File src/lib_ccx/ccx_encoders_webvtt.c and function `ccx_encoders_webvtt.c()`

2. Leaked_storage: Variable `parsebuf` going out of scope leaks the storage it points to. File src/lib_ccx/general_loop.c and function `rcwt_loop()`

PR #649, opened 2017-01-15
Fix defects

sidgairo18:
Refer issue #648


src/lib_ccx/ccx_encoders_webvtt.c
```diff
@@ -182,7 +182,8 @@ int write_stringz_as_webvtt(char *string, struct encoder_ctx *context, LLONG ms_
 		}
 		written = write(context->out->fh, context->encoded_crlf, context->encoded_crlf_length);
 		if (written != context->encoded_crlf_length)
-		{
+		{   
+            		free(el);
```
maxkoryukov: As I know, the CCExtractor uses its own `freep` as a wrapper for `free`

sidgairo18:
In this case free() does the job.

maxkoryukov:
@sidgairo18 , sorry, my mistake

PR #658, opened 2017-01-21
Fix defects

sidgairo18:


src/lib_ccx/ccx_encoders_common.c
```diff
@@ -1276,6 +1276,9 @@ int encode_sub(struct encoder_ctx *context, struct cc_subtitle *sub)
 		freep(&sub->data);
 	if (wrote_something && context->force_flush)
 		fsync(context->out->fh); // Don't buffer
+    
+    	free(context);
```
cfsmp3: That free() is definitely not correct. The context is needed there. change_name doesn't allocate a new context, it returns the same one which is used globally.

Issue #657, opened 2017-01-21
Memory leaks !


sidgairo18:
1. Leaked_storage: Variable unescaped going out of scope leaks the storage it points to in file src/lib_ccx/ccx_encoders_webvtt.c , function `write_stringz_as_webvtt()`
2. leaked_storage: Returning without freeing `context` leaks the storage that it points to in file src/lib_ccx/ccx_encoders_common.c  function `encode_sub()`
3. leaked_storage: Variable `epg` going out of scope leaks the storage it points to in the file /src/lib_ccx/networking.c function `net_send_epg()`
4. leaked_storage: Variable `ctx` going out of scope leaks the storage it points to in the file src/lib_ccx/lib_ccx.c `function init_libraries()`

MatejMecka:
Valgrind tells me those ones are fixed but there are other ones that Valgrind reports for me.

```
==24760==    by 0x40A53D: ocr_bitmap (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x40B4DA: ocr_rect (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x45F377: write_dvb_sub (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x45F66F: dvbsub_handle_display_segment (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x45FA29: dvbsub_decode (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x43DFAE: process_data (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x43EF42: general_loop (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x4071D3: api_start (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x407F61: main (in /root/GoogleCodein/ccextractor/linux/ccextractor)
```

```
==24760==    by 0x40B7D9: paraof_ocrtext (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x4267B1: write_cc_bitmap_as_srt (in /root/GoogleCodein/ccextractor/linux/ccextractor)
==24760==    by 0x41833C: encode_sub (in /root/GoogleCodein/ccextractor/linux/ccextractor)
[...]

PR #666, opened 2017-01-24
Fix defects

sidgairo18:
Issue #657 fix


cfsmp3:
This is horrible:

+			free(el);
+            		free(begin);
+                    	free(unescaped);

It's not your fault, but if we're going to fix the function let's fix it correctly, with a single place in which things are freed(). Instead of checking if write failed and if so free and return, let's check if it succeeded and if so do the next write, etc, and at the end free everything that was allocated.

Otherwise we have code that is technically correct (no leaks) but it's totally unreadable.

PR #673, opened 2017-01-28
Fixed huge memory usage caused by OCR inits.

AlexBratosin2001:
- Fixed memory usage and slow speed caused by too many OCR initializations.
(Now there's only one OCR init per program)


canihavesomecoffee:
Either your branch is running behind the current CCX master branch, or you've broken timing on Teletext sample 78. Mind verifying this manually?

PR #682, opened 2017-02-10
Change constant to expression for readability

barun511:
Also removes the comment, since it was no longer necessary after the change.

### DIFF
--- a/src/lib_ccx/ccx_encoders_webvtt.c
+++ b/src/lib_ccx/ccx_encoders_webvtt.c
@@ -210,7 +210,7 @@ int write_webvtt_header(struct encoder_ctx *context)
 	if (ccx_options.webvtt_create_css)
 	{
 		char *basefilename = get_basename(context->first_input_file);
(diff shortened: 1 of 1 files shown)


barun511:
I thought that replacing 4 with strlen(".css") makes more sense than adding 4 and explaining why 4 is being used as a comment.

saurabhshri:
This is 4 times costlier than the previous way. I see no problem in using a
constant. Readability is same.

CCExtractor is a core component in huge number of services where it is
needed to run several times. The _minor_ change as this becomes quite
visible in that scenario. We should try to make it faster and more
efficient. :)

On 10-Feb-2017 11:08 AM, "Barun Parruck" <notifications@github.com> wrote:

I thought that replacing 4 with strlen(".css") makes more sense than adding
4 and explaining why 4 is being used as a comment.

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/CCExtractor/ccextractor/pull/682#issuecomment-278862255>,
or mute the thread

barun511:
Fair enough! Thanks for the feedback :)

PR #685, opened 2017-02-12
Fix defects

sidgairo18:
memory fixes in src/lib_ccx and src/gpacmp4


src/gpacmp4/stbl_write.c
```diff
@@ -143,6 +143,7 @@ GF_Err stbl_AddDTS(GF_SampleTableBox *stbl, u64 DTS, u32 *sampleNumber, u32 Last
 	if (stts->nb_entries + 2 >= stts->alloc_size) {
 		stts->alloc_size += 2;
 		stts->entries = gf_realloc(stts->entries, sizeof(GF_SttsEntry)*stts->alloc_size);
+		gf_free(DTSs);
```
cfsmp3: Doesn't seem correct to do it there. We're not exiting the function at that point and the free happens later.

Issue #705, opened 2017-03-03
Make some improvements to the Matroska decoder


Izaron:
Some not so hard tasks to become familiar with the CCExtractor code.
Matroska supports now SRT and ASS/SSA subtitle formats
The main files is matroska.c and matroska.h

1. Make an unlimited count of the tracks (subtitle files) and sentences (one subtitle block). Currently, they are constants as `MATROSKA_MAX_TRACKS` and `MATROSKA_MAX_SENTENCES`. You should **realloc**ate memory when it's required and free it after.

2. Make some "defence from the fool". Even the official Matroska specifications are warning that someone may do `copy /b` or `cat a b > c` for "merging" videos. If there are more than one Segment block (normal videos contains the only one block), you can get some error (locate it yourself and fix for getting correct subtitles for all the Segment blocks)

3. Change generated names - remove ".mkv" (and other possible extensions) from "[name].mkv_[lang]_[index].[extension]", for example "video.mkv_fre_2.srt" -> "video_fre_2.srt".

4. Do not allow to have any newline characters at the start of the sentences (only newline, extra spaces at the start and at the end are OK). For example
```
3
00:01:41,720 --> 00:01:43,670



Thank you, brother.
```
Is bad because this sentence is dissapears in the VLC player, and
```
3
00:01:41,720 --> 00:01:43,670
Thank you, brother.
```
Is good. Same thing with ASS/SSA subtitles, but they have the special character for newline.

Specs https://www.matroska.org/technical/specs/index.html (although it is not useful for the tasks)

**UPD:** Please keep all of this in a single pull request (with multiple commits of course) for getting full points.

Izaron:
GSoC: This issue gives 3 qualification points.

Also 1 additional qualification point for creating new option "-mkvlang <lang>" (see options.c) for extracting from only the language user want to.

kapilkd13:
Hi @Izaron  I would like to work on this issue

Izaron:
Hi @kapilkd13 you're welcome! It is good. Also, you do not required to indicate that you are working on issue. First coder who will send good PR will get this points.

Izaron:
Samples:
https://drive.google.com/file/d/0B2fcX80_rHuOOEM3RjhOTXlqbEk/view?usp=sharing (30mb)
https://drive.google.com/file/d/0B2fcX80_rHuOZnU3ZFU3VmY2dTQ/view?usp=sharing (~1gb)
[...]

PR #709, opened 2017-03-05
 	modified subtitle file name genrated by matroska(mkv) - remove 'mkv'

kapilkd13:
In regard with point 3 https://github.com/CCExtractor/ccextractor/issues/705

### DIFF
--- a/src/lib_ccx/matroska.c
+++ b/src/lib_ccx/matroska.c
@@ -1,8 +1,7 @@
 #include "lib_ccx.h"
 #include "matroska.h"
 #include <limits.h>
-#include <string.h>
-#define MAX_FILE_NAME_SIZE 200
+
 void skip_bytes(FILE* file, ULLONG n) {
     FSEEK(file, n, SEEK_CUR);
 }
@@ -34,11 +33,13 @@ UBYTE mkv_read_byte(FILE* file) {
 
 ULLONG read_vint_length(FILE* file) {
     UBYTE ch = mkv_read_byte(file);
-#ifdef _WIN32
-    int cnt = 8 - __lzcnt16(ch);
-#else
-    int cnt = __builtin_clz(ch) - 24 + 1;
-#endif
+    int cnt = 0;
+    for (int i = 7; i >= 0; i--) {
+        if ((ch & (1 << i)) != 0) {
+            cnt = 8 - i;
+            break;
+        }
+    }
     ch ^= (1 << (8 - cnt));
     ULLONG ret = ch;
     for (int i = 1; i < cnt; i++) {
@@ -126,7 +127,7 @@ void parse_ebml(FILE* file) {
                 MATROSKA_SWITCH_BREAK(code, code_len);
             default:
                 if (code_len == MATROSKA_MAX_ID_LENGTH) {
-                    mprint(MATROSKA_ERROR "Unknown element 0x%x at position %lld, skipping EBML block\n", code,
+                    mprint(MATROSKA_ERROR "Unknown element 0x%x at position %I64d, skipping EBML block\n", code,
                            get_current_byte(file) - MATROSKA_MAX_ID_LENGTH);
                     set_bytes(file, pos + len);
                     return;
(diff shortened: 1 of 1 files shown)


src/lib_ccx/matroska.c
```diff
@@ -738,14 +738,35 @@ void parse_segment(struct matroska_ctx* mkv_ctx)
...
+    if (lastdot != NULL)
+        *lastdot = '\0';
+    return returnname;
+}
+
 char* generate_filename_from_track(struct matroska_ctx* mkv_ctx, struct matroska_sub_track* track)
-{
-    char* buf = malloc(sizeof(char) * 200);
+{ 
+  char* filename = malloc(sizeof(char)*(MAX_FILE_NAME_SIZE-30));
```
Izaron: What this magical numbers 30 and 31 are for?

src/lib_ccx/matroska.c
```diff
@@ -738,14 +738,35 @@ void parse_segment(struct matroska_ctx* mkv_ctx)
     }
 }
 
+char *filename_without_ext(char* filename) {
+    char *returnname;
+    char *lastdot;
+    if (filename == NULL)
+         return filename;
+    if ((returnname = malloc (strlen (filename) + 1)) == NULL)
```
Izaron: If we want X chars, we should alloc (sizeof(char)*X), not just (X). This is good practice, because sizeof(char) not always equals to 1

src/lib_ccx/matroska.c
```diff
@@ -738,14 +738,35 @@ void parse_segment(struct matroska_ctx* mkv_ctx)
...
+    char* buf = malloc(sizeof(char) * MAX_FILE_NAME_SIZE);
     if (track->lang_index == 0)
-        sprintf(buf, "%s_%s.%s", mkv_ctx->filename, track->lang, matroska_track_text_subtitle_id_extensions[track->codec_id]);
+        sprintf(buf, "%s_%s.%s", filename_non_ext, track->lang, matroska_track_text_subtitle_id_extensions[track->codec_id]);
     else
-        sprintf(buf, "%s_%s_%llu.%s", mkv_ctx->filename, track->lang, track->lang_index,
+        sprintf(buf, "%s_%s_%llu.%s", filename_without_ext, track->lang, track->lang_index,
                 matroska_track_text_subtitle_id_extensions[track->codec_id]);
+free(filename);
+free(filename_non_ext);
```
Izaron: Would be good if all the code was formatted with correct spacing. Lines 758-761 too.

src/lib_ccx/matroska.c
```diff
@@ -738,14 +738,35 @@ void parse_segment(struct matroska_ctx* mkv_ctx)
...
+
 char* generate_filename_from_track(struct matroska_ctx* mkv_ctx, struct matroska_sub_track* track)
-{
-    char* buf = malloc(sizeof(char) * 200);
+{ 
+  char* filename = malloc(sizeof(char)*(MAX_FILE_NAME_SIZE-30));
+  char * filename_non_ext;
+  filename[MAX_FILE_NAME_SIZE-31] = '\0';
+  strncpy(filename, mkv_ctx->filename,(MAX_FILE_NAME_SIZE-31));
+  filename_non_ext=filename_without_ext(filename);
```
Izaron: Also, why do you use additional variable `filename`? Why not passing mkv_ctx->filename to the function?

src/lib_ccx/matroska.c
kapilkd13: Yes, It is a big mistake. Would have caused errors among some system. Thanks for pointing

src/lib_ccx/matroska.c
kapilkd13: Actually I kept maximum size to 200, this 30 is for any additional format addition like eng_srt or any other language format. Ths means effectively 170 characters from original file name and 30 chars to define our srt extension type
Originally it was 200 chars so I kept it to 200 chars. Do you want any change.

src/lib_ccx/matroska.c
kapilkd13: I would do all above changes in my next commit. :)

PR #710, opened 2017-03-06
Modified formatting and did some changes- imp to merge

kapilkd13:
In previous commit i forgot to add a file matroska.h which affects the compilation.
Merge this request. Until it is merged you can add `#define MAX_FILE_NAME_SIZE 200` in your matroska.h
I am really sorry.

### DIFF
--- a/src/lib_ccx/matroska.c
+++ b/src/lib_ccx/matroska.c
@@ -2,6 +2,7 @@
 #include "matroska.h"
 #include <limits.h>
 #include<string.h>
+
 void skip_bytes(FILE* file, ULLONG n) {
     FSEEK(file, n, SEEK_CUR);
 }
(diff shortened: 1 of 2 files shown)


cfsmp3:
Code needs to be reasonably self-explaining, however:

returnname[MAX_FILE_NAME_SIZE-30]='\0';

I have no idea where that 30 comes from.

Izaron:
Yes, technically this could be correct, but not in terms of code cleanliness (invalid spacing, unused include, this strange '\0' at the 170-th char, etc.). At least we can generate filename without extension once and for all, but not for all tracks every time.
Also @kapilkd13 I found bug in your code anyway, when it's multiple tracks with same lang, so not technically correct. But I thought that you will check all 3 examples from the issue page, it is not a huge waste of time.
```
Output file: /home/izaron/Downloads/Vikings S01 (2013) BDRip-AVC/vikings_rus.ass
Output file: UH��H�� H�}�H�}�_rus_1.ass
Output file: UH��H�� H�}�H�}�_rus_2.srt
Output file: UH��H�� H�}�H�}�_rus_3.srt
Output file: /home/izaron/Downloads/Vikings S01 (2013) BDRip-AVC/vikings_eng.srt
Output file: UH��H�� H�}�H�}�_eng_1.srt
```

Issue #678, opened 2017-02-06
GSoC qualification: Make CCExtractor more linux standard compliant


cfsmp3:
This task gives you enough points to qualify (but remember that your proposal still needs to be good, etc).

CCExtractor is written in pure C, it uses external hard-linked libraries and it is designed to be portable to different platforms. Because of that we have multiple building scripts which somebody should maintain. The task is to create

    Autoconf scripts
    Man pages
    .tgz, .deb, .rpm, .tar.xz (archlinux) installation packages

Izaron:
We just need to build package from built binary from linux/build-static.sh
.deb - just rewrite outdated https://github.com/CCExtractor/ccextractor/blob/master/package_creators/debian.sh remove REQUIRES and so on
.rpm - http://meinit.nl/creating-rpm-some-binary, but we can also build .rmp from **Debian/Ubuntu**

Izaron:
Would you like to have official packages for the most popular linux distros?
I mean, if everybody will be able to launch Linix distro, write in terminal line as `sudo apt-get install ccextractor` and get binary? :)

Izaron:
https://www.linux.com/learn/easy-package-creation-checkinstall
Can be helpful

alexandrumc:
I'm starting to do it.

pranavsid98:
I am interested in this issue. I would like to work on this.

Pranav

techfreakworm:
I am working on autoconf scripts for ccextractor and I have a slight problem. In Makefile.am (with non-recursive approach) I have set, "AM_LDFLAGS = -lm -zmuldefs -ltesseract -llept". But still the build crashes with "/usr/bin/ld: src/gpacmp4/av_parsers.o: undefined reference to symbol 'trunc@@GLIBC_2.2.5'
/lib/x86_64-linux-gnu/libm.so.6: error adding symbols: DSO missing from command line"
Any idea where I might be wrong?

xvkdev:
To make it clear - we don't want to pregenerated packages, man, etc: We want scripts that generate them automatically, otherwise we'd been stuck with old versions really quickly.

cfsmp3:
To make it clear - we don't want to pregenerated packages, man, etc: We want scripts that generate them automatically, otherwise we'd been stuck with old versions really quickly.

fandango-:
[...]

PR #715, opened 2017-03-13
[IMPROVEMENT] Add Autoconf scripts to make CCExtractor more linux standard compliant

fandango-:

---

This PR add autoconf scripts (mainly configure.ac and Makefile.am) to allow the standard configure / make / make install process for building and installing CCExtractor on Linux systems.

Makefile.am is mostly derived from the existing Makefile (removed in this PR as it is no longer required) with target renaming as appropriate.

All the lib checking (tesseract, leptonica, etc.) and option parsing (--enable-ocr, --enable-ffmpeg, etc.) is now moved to the configure step after the addition of relevant macros in configure.ac

The other files added (Makefile.in, config.h.in, etc.) allow the building and installation to take place on systems where autotools is not installed.

pre-build.sh was modified to use relative paths with respect to the directory in which the shell script exists and not the current build directory.

Tested on
Kernel		: Linux 4.4.48-1-MANJARO (x86_64)
C Library		: GNU C Library version 2.24 (stable)
Default C Compiler		: GNU C Compiler version 6.3.1 20170109 (GCC) 
Distribution		: Manjaro Linux

Part 1 of 3 to resolve Issue #678


fandango-:
Thanks for the review. Okay, I've removed the autogenerated scripts. Also, I've added a bootstrap script to generate the configure file using the autotools commands.

So, new flow for building and installing would be:
$ ./bootstrap
$ ./configure
$ make
$ sudo make install

rkuchumov:
seems that package generated with `make dist` doesn't work. And don't forget to update .gitignore

fandango-:
Yes, thanks for noting the issue. 'make dist' now works - it generates a .tar.gz file with all the required files. However, 'make distcheck' will fail as it expects configure in the topmost directory. Currently, the tar archive generated has 2 subfolders with the ccextractor-0.85 folder - linux and src. The configure is actually in the linux directory. So, to test it out:
$ make dist
$ tar -xvf tar -xvf ccextractor-0.85.tar.gz
$ cd ccextractor-0.85/linux/
$ ./configure
$ make
$ sudo make install

Also, the .gitignore was updated.

rkuchumov:
when I do 
```
./bootstrap
./configure --prefix=/home/rk/.local/bin/
make install -j4
```
it still tries to copy executable to `/usr/bin/` instead of my local folder. `--exec-copy` didn't help as well.

cfsmp3:
Closing because an alternative PR has been merged, you still get GSoC points for the work.

PR #716, opened 2017-03-13
[IMPROVEMENT]GSoC qualification: Make CCExtractor more linux standard compliant #678

techfreakworm:
Please prefix your pull request with one of the following: **[FEATURE]** **[FIX]** **[IMPROVEMENT]**.

**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist. 
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [x] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [ ] I am an active contributor to CCExtractor.

---

Successfully wrote Autoconf scripts without changing any other file or script. Prior build mechanisms still work and README.md is updated accordingly.

(I admit pull request exists but  pre-existing files were modified in that patch and this one doesn't.)
Autoconf tools are required for this because if scripts are to be modified they need to have Autoconf tools installed.

PR #717, opened 2017-03-14
[IMPROVEMENT]Make CCExtractor more linux standard compliant #678

techfreakworm:

---

-Added autoconf scripts such that they would be able to generate a tarball using `make dist` and create a script in ccextractor/package_creators for the same. 
-No pre-existing files changed.
-Pre-existing build mechanisms stay intact.
-Updated README.md accordingly.
-Wrote man pages in ccextractor/linux.
-wrote tarball.sh for *tar.gz release package creation in ccextractor/linux.
Resolves many issues stated in #678


techfreakworm:
Working on the remaining problems of the current issue. Will be creating new pull request for those problems soon.

techfreakworm:
I've almost completed the task for the issue. There's a slight problem though, since I don't have any linux distro installed to test .rpm or arch linux packages, maybe someone else would be interested in writing the package creation scripts for those linux enivornments. And I want to ask how many points would be deducted for not completing the package creation task (.rpm and .tar.xz packages) ?

rkuchumov:
The problems I've mentioned in #716 seems to stay in this PR as well.

Man page seems fine, but:

- Somehow debug options (`-debugdvbsub` etc) got into the "File name related options" section
- There's no `-trimfR` option, just `-trim`
- In the description part it's better to list main available features 
- (Being picky :) ) Please check GNU standard for displaying alternatives. I'm not sure but I think they are separated by commas, but you have just spaces (e.g. `-ff --forceflush`) or slashes (e.g. `-ru1 / ru2 / ru3`). it's better to be consistent.

You can use Linux Containers (e.g. Docker) to have several distros on top of a single kernel for testing packages. Or just use virtual machine. Anyway this task was supposed for summer, so it's better for you to finish what you've started and make it good enough for merging.

techfreakworm:
@rkuchumov A polite ping, though adding the autoconf scripts to ccextractor/linux would prevent clogging the root dir, there's a slight problem though. `make dist` doesn't work (doesn't include the files required) if `Makefile` isn't in root dir and thus taking away the functionality to create tarballs. And also, I can't seem to find a way to automatically generate man pages in `groff` could you provide with some pointer? Update on script: Compiled fine on my pc, seems like `allheaders.h` belongs to `tesseract` and should not be included in ccextractor.c if tesseract is not installed. I'm working to improve `configure.ac` for the same.

rkuchumov:
That's seems reasonable indeed. GNU standard is also forcing us to use their directories structure, but we don't :)

Actually I can't think of an easy way of converting help screen to a man page. Maybe converting in a different direction would work. You can write your ideas in proposal :)

cfsmp3:
For the manpage, I think the correct solution is to add a -man to ccextractor, like this:

ccextractor -man >/usr/local/man/man1/ccextractor.1
gzip -f /usr/local/man/man1/ccextractor.1
mandb -q 

Where obviously the only work is in adding that -man parameter which would just display the current help screen with a few extra lines for the man formatting.

techfreakworm:
@rkuchumov @cfsmp3 I updated the scripts and moved them to `ccextractor/linux`. To compile:
$cd linux
$./autogen.sh
$./configure
$make
Since, `make dist` still doesn't work if run directly.However, I've made `ccextractor/package_creators/tarball.sh` as a workaround. 
$cd package_creators
$./tarball.sh
The script makes tarball using GNU directory structure standards, I thought it would be good if the tarball maintains GNU standards since tarball will only be used on LINUX operating system.
Update: Working on `man` pages and other package creation scripts, will commit changes related to the same soon.

rkuchumov:
That's interesting workaround, but what if I want to specify some arguments for compilation? I think 
you should pass script's args to `./configure` call and document it somewhere.

when I do `./configure --enable-ocr` it doesn't fail, but `make` yields an error because I don't have `allheaders.h`.  ./configure should fail earlier telling that I don't have something installed.

don't forget to update `.gitignore` file because `git status` shows a lot of auto-generated files. we don't want them in version control system.

techfreakworm:
@rkuchumov Updated `configure.ac` so that `make` doesn't fail, also updated `.gitignore`.

rkuchumov:
Ah, indeed, my bad.

`./configure --enable-ocr` now prints:
```
./configure: line 413: test: required: integer expression expected
configure: error: to
./configure: line 302: return: required: numeric argument required
./configure: line 312: exit: required: numeric argument required
```
It doesn't seem ok :)

techfreakworm:
@rkuchumov Corrected the mistake. Please check it out. :)

rkuchumov:
I tried your scripts in clean ubuntu system and `./configure` yields:  
```
checking for getLeptonicaVersion in -llept... no
./configure: line 4179: syntax error near unexpected token `lept,'
./configure: line 4179: `  HAS_LEPT=1 && PKG_CHECK_MODULES(lept, lept)'
```

Also you should set execution permissions for autogen.sh

techfreakworm:
I tried to compile it again on a clean install of Ubuntu 16.04 LTS with no traces of `tesseract` or `leptonica`.
Worked fine for me. Did you `apt-get purge tesseract-ocr` ? because it doesn't remove tesseract and leptonica completely so I compiled on a clean install again and it compiled fine. Here are logs:
[logs](https://paste.ubuntu.com/24221277/) for `./configure` on a clean ubuntu without `tesseract-ocr`
[logs](https://paste.ubuntu.com/24221294/) for `./configure` on a clean ubuntu with `tesseract-ocr`:

rkuchumov:
No, I used Docker's Ubuntu container for that with only autotools and gcc installed.
```
docker run -v `pwd`:/mnt -t -i ubuntu /bin/bash
apt update && apt install gcc autotools
...
```
I can't provide logs right now.

techfreakworm:
I can't figure out what the problem is, I checked it again on clean install of Manjaro linux on VMware but it compiled fine on that too. Perhaps the Docker's Ubuntu container doesn't have `pkg-config` utility installed?

rkuchumov:
I installed `pkg-config` and it still shows the same error

saurabhshri:
If absence of `pkg-config` causes this, then, in that case, maybe adding something like below in `configure.ac` should be helpful.

```
AC_MSG_CHECKING([pkg-config m4 macros])
if test m4_ifdef([PKG_CHECK_MODULES], [yes], [no]) == yes; then
	AC_MSG_RESULT([yes]);
else
	AC_MSG_RESULT([no]);
	AC_MSG_ERROR([
pkg-config is required.])
fi
```

techfreakworm:
I made the changes suggested by @saurabhshri , although `autoconf` checks for `pkg-config` (as in logs) on its own along with its version, I though extra check with failure error won't hurt. Thank You!
@rkuchumov , I checked the scripts on friend's machine, installed VMware, installed Manjaro in it and compiled again. It worked fine. Did the same with Ubuntu 16.04 LTS and worked fine with that too.

rkuchumov:
the problem was indeed with `pkg-config`. Seems that some temporary files didn't clean up (or whatever) when I ran ./configure the second time after installing `pkg-config`. It's fine now.

But  `./configure --enable-hardsubx` gives:
```
./configure: line 413: test: required: integer expression expected
configure: error: to
./configure: line 302: return: required: numeric argument required
./configure: line 312: exit: required: numeric argument required
```
I've seen this already :)

techfreakworm:
@rkuchumov , I don't know what's with this text giving such errors in configuration. Anyways, I corrected it, sorry for inconvenience for such trivial mistakes. 
What are the permission that you want me to fix on `autogen.sh` ?. They work fine on every linux I tested, other than the fact that it requires `bash` command in Manjaro linux to run it.

rkuchumov:
just for execution (`u+x`) as when I fetch your branch `autogen.sh` and `tarball.sh` are not executable. maybe it's a problem with git, idk :)

techfreakworm:
@rkuchumov , please check if it works correctly now. 
Also I wanted to ask if I can do rest of the work (manpages and remaining package creation scripts) after 3rd April? (That is student application deadline), since I need to research about the Idea and concentrate on my proposal too. Would that be ok?

rkuchumov:
It's ok now and it can be merged. Thanks for you work.

yes, sure. This task is for the summer but you can work on it before gsoc. And don't forget about the others as this one won't take much time.

techfreakworm:
@rkuchumov Thanks. Yes I'll finish the remaining work as soon as I am done with the proposal.
And also, are 7 GSoC points to be earned before submitting the proposal? Or finishing the work after proposal would count too? And if they are to be earned before application deadline then how many points will be deducted if I do the remaining work after 3rd April?

rkuchumov:
@techfreakworm they can be earned any time before we start selecting proposals. It officially starts on 3rd April. but it's one month.
Also, I forgot to tell you to update `docs/CHANGES.TXT` :)

techfreakworm:
@rkuchumov  yes, I'll update them soon😀

Issue #699, opened 2017-02-26
Few lines missing from output of .bin file [sample attached]


saurabhshri:
While working on another issue, I created a .bin file from a sample (https://sampleplatform.ccextractor.org/sample/67). 

When extracting subs from .bin file, the ending few lines are missing, while they are perfectly extracted from original file.

Original :

```
12
00:00:52,140 --> 00:00:54,279
Mogu, your bag!

13
00:00:54,380 --> 00:00:56,679
<font color="#00ffff">(Strains)</font>

14
00:00:56,780 --> 00:00:58,979
<font color="#00ffff">It's strong!</font>

```

From .bin of above sample (http://gsocdev3.ccextractor.org/~saurabhshri/67.bin) : 

```
12
00:00:52,140 --> 00:00:54,279
Mogu, your bag!

```

I created bin using : `ccextractor -ts -pn 578 -out=bin -o 67.bin 67` where 67 was the name of sample.

EDIT : Corrected the .bin file. Previous file had incorrect timings.

cfsmp3:
2 GSoC points.

saurabhshri:
Closing through #725 . The issue appears to be resolved with attached file.
Note : Timing is off by few milliseconds (but they may be for good, maybe someone can check on it).

PR #725, opened 2017-03-20
Issue 699 solved

Amey-Jain:

- [*] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [*] I have checked that another pull request for this purpose does not exist.
- [*] I have considered, and confirmed that this submission will be valuable to others.
- [*] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [*] I give this submission freely, and claim no ownership to its content.

- [*] I absolutely love CCExtractor, but have not contributed previously.

---
Due to blank returns from various conditions in process_telx_packet() (even after page_buffer containing data), data was not being written to output file. Also after file buffer is completed, data needs to be dumped to file.

### DIFF
--- a/src/lib_ccx/general_loop.c
+++ b/src/lib_ccx/general_loop.c
@@ -1102,7 +1102,7 @@ int rcwt_loop(struct lib_ccx_ctx *ctx)
 	int caps = 0;
 	LLONG result;
 	struct encoder_ctx *enc_ctx = update_encoder_list(ctx);
-
+	struct Teletext *telctx;
 	// As BUFSIZE is a macro this is just a reminder
 	if (BUFSIZE < (3*0xFFFF + 10))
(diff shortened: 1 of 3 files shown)


AlexBratosin2001:
The teletext processing originates from telxcc, an open source teletext parser written by Petr. As far as I can see your PR fixes the missing lines in the .bin output BUT it creates a 20 ms timing mismatch, which is quite huge comparing it to previous correct outputs. In my opinion, the fix for the multiple lines shouldn't affect timing at all. Moreover, the original output is just fine, so I don't think it has to do with any telxcc processing. Be sure not to dump data prematurely.

cfsmp3:
@Amey-Jain Please join us on slack :-)

canihavesomecoffee:
#699

PR #728, opened 2017-03-30
[FEATURE] Arch Linux installation script (create .pkg.tar.xz archive or install directly)

fandango-:

---

Add installation script for Arch Linux based distributions.

Steps to test this out (on an Arch Linux based distribution):
$ cd package_creators/
$ ./arch.sh

There is an option to do 2 things (install directly or create the .pkg.tar.xz archive for distribution purposes without installing on host machine).

Tested on
Kernel : Linux 4.4.48-1-MANJARO (x86_64)
C Library : GNU C Library version 2.24 (stable)
Default C Compiler : GNU C Compiler version 6.3.1 20170109 (GCC)
Distribution : Manjaro Linux

Resolves sub-part of Issue #678

### DIFF
--- a/package_creators/PKGBUILD
+++ b/package_creators/PKGBUILD
@@ -1,0 +1,22 @@
+pkgname=ccextractor
+pkgver=0.85
+pkgrel=1
+pkgdesc="A closed captions and teletext subtitles extractor for video streams."
+arch=('i686' 'x86_64')
+url="http://www.ccextractor.org"
+license=('GPL')
+depends=('gcc-libs' 'tesseract')
+source=(
+   $pkgname-$pkgver.tar.gz
+)
+
+build() {
+  cd "$srcdir/$pkgname-$pkgver"
+  ./configure --prefix="$pkgdir/usr/local" --CC=gcc
+  make -j4
+}
+
+package() {
+  cd "$srcdir/$pkgname-$pkgver"
+  make install
+}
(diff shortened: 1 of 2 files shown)


rkuchumov:
1.  I've got `./arch.sh:read:8: -p: no coprocess`. (I fixed it by changing #!/bin/sh to bash, but you should check it)
2.  There may be no `sudo` in the system
3.  When ./autogen.sh or ./tarball.sh fails arch.sh still asks me to install the package.
4.  When I answer anything but `n` or `N` to the prompt, it still install the package. But default behavior everywhere in this case is to exit

fandango-:
Thanks for the review.
1. Changed to use bash as the shell
2. Use 'su' instead of 'sudo' now
3. Fixed; now if a command fails, the script won't continue
4. I was just emulating the default behaviour of pacman / makepkg - Y is the default there. However, I've now changed the prompt within arch.sh to have N as the default.

rkuchumov:
> Use 'su' instead of 'sudo' now

But when the system has sudo, you should use it :) Usually scripts try `sudo` at first, (which asks user password), and if they fail to find it, they use `su` (which asks root password). You should either implement this logic or just remove installation step and let user do it.

Don't forget to add `.gitignore`, to update `docs/CHANGES.TXT` and to add instructions to `README.md`

fandango-:
Done - 'sudo' is used if available; fallback to 'su'. Also, updated the mentioned files.

rkuchumov:
Ok, thanks.
It can be merged.

PR #732, opened 2017-04-07
[FEATURE] Scripts to generate CCExtractor Man-page .

techfreakworm:

---

Added scripts to generate CCExtractor man-page using `ccextractor --help`. Also updated other scripts accordingly to include `ccextractor.man` while installing or creating packages.
Fixes sub part of issue #678 

**EDIT**
Removed the scripts, added `--man` parameter to CCExtractor itself. 
Squashed the commits. 
Though it incurs a small size dependency `help2man` tool but it generates fairly good manpage as compared to writing output of `print_usage` directly to file and then changing the formatting of text (Adding sections like NAME, SYNOPSIS, TITLE) manually.
Though the task can be done without the tool but the text formatting using C would be more lines of code to maintain whenever the contents of `print_usage()` are changed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,7 +61,6 @@ windows/enc_temp_folder/*
 src/cmake-build-debug/
 src/.idea
 
-
 #Autotools
 linux/config.h
 linux/config.log
@@ -95,3 +94,6 @@ src/zvbi/.dirstamp
 
 # Arch
 package_creators/*.pkg.tar.xz
+
+# Manual Pager
+manpage/*.1
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -3,7 +3,7 @@
 - Fix: Prevent the OCR being initialized more than once (happened on multiprogram and
   PAT changes)
 - New: Added build/installation script for .pkg.tar.xz (Arch Linux).
-  
+- New: Added Manpage generation script for Linux and Mac.
 
 0.85b (2017-1-26)
 -----------------
--- a/linux/Makefile.am
+++ b/linux/Makefile.am
@@ -310,7 +310,14 @@ ccextractor_LDADD += $(TESS_LIB)
 ccextractor_LDADD += $(LEPT_LIB)
 endif
 
-EXTRA_DIST = ../src/gpacmp4/gpac/sync_layer.h ../src/lib_ccx/ccfont2.xbm ../src/utf8proc/utf8proc_data.c
+EXTRA_DIST = ../src/gpacmp4/gpac/sync_layer.h ../src/lib_ccx/ccfont2.xbm ../src/utf8proc/utf8proc_data.c ../manpage/mangen.sh
 
+install-exec-hook:
+	${SHELL} ../manpage/mangen.sh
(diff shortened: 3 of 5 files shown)


cfsmp3:
I don't like having a static manpage. It should be generated by CCExtractor itself by changing the behavior of print_usage() if a parameter -man page is passed. 

Having a help screen and a man page that need to be separately maintained makes no sense - it's just not needed, less than 20 lines of code would be enough to have a proper solution.

techfreakworm:
@cfsmp3 ,thanks for review. I'll improvise.

techfreakworm:
@cfsmp3 @rkuchumov Please check if the patch is good enough to merge. Thanks :)

rkuchumov:
Why didn't you create a separate script for calling help2man? I don't think users would like to run ccextractor as a root for installing man page. It's package manager duties. The code you've written is also platform-specific, doesn't has all required checks and introduces some magic numbers.

I suggest you to create a script for it which would be called at the stage of package assembly. And when user install package the generated man page would be copied to the appropriate location. 

Probably Carlos meant that you should implement functionality of help2man inside ccextractor. If help2man yields decent manpages, then there is no need for that.

cfsmp3:
I think CCExtractor should generate (but not install) the .man files, *or*
at least it should be possible to generate the .man file from CCExtractor's
output. What I don't want to do is maintain the help in two separate
places, so any solution that imply that we export the file from
CCExtractor, then edit that file to generate the man is no solution.

However, a script that takes the help from CCExtractor and without any kind
of additional human intervention generates the .man file is OK (but not
great, compared to having a -man option).

techfreakworm:
@cfsmp3 @rkuchumov thanks, I'll improvise.

techfreakworm:
@rkuchumov Created a seperate script to create manpage. Though the output can't be intervened much to organise the content beautifully, so as to turn down on maintenance cost. I used stream editor `sed` to do some minimal editing. Which I don't think would be required to maintain in future.
1. Added a section **SYNOPSIS** to focus on `ccextractor` usage.
2. Added focus on 'Filename related options:' by making it bold.
The manpage generated from script won't require extra maintenance, would just make a manpage from the content of `print_usage()` in best way possible.

rkuchumov:
1) You are copying man pages to a fixed destination `mv ../manpage/ccextractor.1 /usr/local/man/man1` . So installing to a different `--prefix` is not possible
2) Is it possible to generate man page during tarball assembly stage? So end-user won't have to install `help2man` to install ccextractor as it doesn't really depend on it.
3) Even more, when I install from tarball (extracted to `/tmp`) `mangen.sh` yields `manpage/mangen.sh: line 14: cd: ../linux: No such file or directory`
4)  "Lists" doesn't look like lists:
image
5) No line feeds after option names: 
image
6) This doesn't look good: 
image
6) ... and you can notice a lot more, if you compare help screen to man page side by side 
7) I don't like directory `manpage` (as well as `package_creators`) in project's root because it's linux-specific

techfreakworm:
Thank you for the review. I'll Improvise.

cfsmp3:
Closing this due to inactivity, please resubmit when ready.

PR #744, opened 2017-06-27
Refactoring the source code and extension to python module.

Diptanshu8:

- [x ] I have checked that another pull request for this purpose does not exist.
- [ x] I have considered, and confirmed that this submission will be valuable to others.
- [ x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [ x] I give this submission freely, and claim no ownership to its content.

---
This pull request mainly comprises of two major parts:

- Refactoring of the source code of `ccextractor.c` and breaking the code into smaller function.
- Scripts to generate the python module which would be able to handle a single input file at a single go(so far).

Note: to generate the python module, just run the `build_library` executable.


canihavesomecoffee:
@Diptanshu8 Could you please fix the errors that cause *all* tests to fail? All tests exit with exit code 1...

linux/build
```diff
@@ -14,7 +14,7 @@ BLD_LINKER="-lm -zmuldefs -l tesseract -l lept"
 
 ./pre-build.sh
 
-out=$((LC_ALL=C gcc $BLD_FLAGS $BLD_INCLUDE -o ccextractor $BLD_SOURCES $BLD_LINKER) 2>&1)
+out=$((LC_ALL=C gcc $BLD_FLAGS $BLD_INCLUDE -o ccextractor $BLD_SOURCES $BLD_LINKER)2>&1)
```
canihavesomecoffee: Could it be possible that this vanished space is causing the errors?

src/ccextractor.c
```diff
@@ -457,3 +422,80 @@ int main(int argc, char *argv[])
...
+}
+
+void checking_configuration_file(struct ccx_s_options api_options){
+    parse_configuration(&api_options);
+}   
+
+int compile_params(struct ccx_s_options *api_options,int argc){
+    
+    api_options->myarguments = realloc(api_options->myarguments, (api_options->argument_count+1) * sizeof *api_options->myarguments);
+      api_options->myarguments[api_options->argument_count] = malloc(strlen("./ccextractor")+1);
```
canihavesomecoffee: What happens if you do not call the binary from the folder, but with a full path? e.g. calling it with `/home/foo/ccextractor/linux/ccextractor` instead of `./ccextractor`?

canihavesomecoffee:
I checked the tests, they look pretty fine. There's 2 samples (both linux & windows) which have timing changes apparently, two tests on windows are broken because the platform can't handle it (spupng & a dictionary file), and one file on windows seems to find no subtitles, whereas on linux it does.

You could possibly check out the first two, the others are platform-caused.

cfsmp3:
There a binary file here (.so). Needs to be removed before the PR can be merged.

Diptanshu8:
The `.so` file has been removed @cfsmp3

PR #802, opened 2017-10-31
[FEATURE] Support for Source-Specific Multicast

squarebracket:

---

This is a simple change to allow for binding to source-specific multicast UDP feeds.

I've modified the GUI to allow for enough characters to bind to a SSM address, but it doesn't seem to work, even for simple UDP feeds. Can anyone confirm that the GUI works with UDP?

Also, I tried to validate that it works with Windows, but was unable. I see the packets in wireshark, but all of `ccextractor`, `vlc`, and `mpv` do not seem to get anything. If someone with more Windows-fu can validate that would be great.

### DIFF
--- a/src/lib_ccx/ccx_common_option.h
+++ b/src/lib_ccx/ccx_common_option.h
@@ -157,7 +157,7 @@ struct ccx_s_options // Options from user parameters
 	LLONG debug_mask;                                   // dbg_print will use this mask to print or ignore different types
 	LLONG debug_mask_on_debug;                          // If we're using temp_debug to enable/disable debug "live", this is the mask when temp_debug=1
 	/* Networking */
-    char *udpsrc;
+	char *udpsrc;
 	char *udpaddr;
 	unsigned udpport;                                   // Non-zero => Listen for UDP packets on this port, no files.
 	char *tcpport;
--- a/src/lib_ccx/ccx_common_option.h
+++ b/src/lib_ccx/ccx_common_option.h
@@ -157,6 +157,7 @@ struct ccx_s_options // Options from user parameters
 	LLONG debug_mask;                                   // dbg_print will use this mask to print or ignore different types
 	LLONG debug_mask_on_debug;                          // If we're using temp_debug to enable/disable debug "live", this is the mask when temp_debug=1
 	/* Networking */
+    char *udpsrc;
 	char *udpaddr;
 	unsigned udpport;                                   // Non-zero => Listen for UDP packets on this port, no files.
 	char *tcpport;
--- a/src/lib_ccx/networking.c
+++ b/src/lib_ccx/networking.c
@@ -1018,21 +1018,21 @@ int start_upd_srv(const char *src_str, const char *addr_str, unsigned port)
 	}
 
 	if (IN_MULTICAST(addr)) {
-        int setsockopt_return = 0;
-        if (src_str != NULL) {
-            struct ip_mreq_source multicast_req;
-            multicast_req.imr_sourceaddr.s_addr = htonl(src);
-            multicast_req.imr_multiaddr.s_addr = htonl(addr);
-            multicast_req.imr_interface.s_addr = htonl(INADDR_ANY);
-            setsockopt_return = setsockopt(sockfd, IPPROTO_IP, IP_ADD_SOURCE_MEMBERSHIP, (char *)&multicast_req, sizeof(multicast_req));
-        }
-        else
-        {
-            struct ip_mreq multicast_req;
-            multicast_req.imr_multiaddr.s_addr = htonl(addr);
-            multicast_req.imr_interface.s_addr = htonl(INADDR_ANY);
-            setsockopt_return = setsockopt(sockfd, IPPROTO_IP, IP_ADD_MEMBERSHIP, (char *)&multicast_req, sizeof(multicast_req));
-        }
+		int setsockopt_return = 0;
+		if (src_str != NULL) {
+			struct ip_mreq_source multicast_req;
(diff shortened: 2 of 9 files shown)


src/lib_ccx/ccx_common_option.h
```diff
@@ -157,6 +157,7 @@ struct ccx_s_options // Options from user parameters
 	LLONG debug_mask;                                   // dbg_print will use this mask to print or ignore different types
 	LLONG debug_mask_on_debug;                          // If we're using temp_debug to enable/disable debug "live", this is the mask when temp_debug=1
 	/* Networking */
+    char *udpsrc;
```
anshul1912: check surrounding indentation

src/lib_ccx/params_dump.c
```diff
@@ -18,7 +18,9 @@ void params_dump(struct lib_ccx_ctx *ctx)
 			mprint ("stdin");
 			break;
 		case CCX_DS_NETWORK:
-			if (ccx_options.udpaddr == NULL)
+            if (ccx_options.udpsrc != NULL)
+                mprint ("Network, %s@%s:%d", ccx_options.udpsrc, ccx_options.udpaddr, ccx_options.udpport);
```
anshul1912: udpaddr could be null, first check for null or in if add condition

src/lib_ccx/params.c
```diff
@@ -2220,8 +2224,21 @@ int parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
 		/* Network stuff */
 		if (strcmp (argv[i],"-udp")==0 && i<argc-1)
 		{
+            char *at = strchr(argv[i + 1], '@');
```
anshul1912: Look for surrounding indentation

src/lib_ccx/networking.c
```diff
@@ -1000,10 +1018,23 @@ int start_upd_srv(const char *addr_str, unsigned port)
 	}
 
 	if (IN_MULTICAST(addr)) {
-		struct ip_mreq group;
-		group.imr_multiaddr.s_addr = htonl(addr);
-		group.imr_interface.s_addr = htonl(INADDR_ANY);
-		if (setsockopt(sockfd, IPPROTO_IP, IP_ADD_MEMBERSHIP, (char *)&group, sizeof(group)) < 0)
+        int setsockopt_return = 0;
```
anshul1912: Look for surrounding indentation

anshul1912:
One of the build error I see is "/usr/bin/ld: cannot find -lz"

saurabhshri:
@anshul1912 You probably are using cmake to build. Earlier we were using zlib installed in the environment, hence that flag was required. Now we are using the supplied ones so that flag probably is obsolete. Could you try building again after removing this line : https://github.com/CCExtractor/ccextractor/blob/master/src/CMakeLists.txt#L86 ?

src/lib_ccx/params_dump.c
squarebracket: I'm not sure I follow. Is `params_dump.c` used without `params.c` being called? Unless I messed up [the parameter parsing](https://github.com/squarebracket/ccextractor/blob/1dd4201496e6eb1e829de4ca164fb0c4c272295d/src/lib_ccx/params.c#L2225), if you have a `udpsrc` you _must_ have a `udpaddr` -- even if it's just `\0` -- so you won't end up in a condition where `udpaddr` is null if `udpsrc` is not also null.

Of course, if there's a path to `params_dump.c` which doesn't first use `params.c`, then yeah it might be null, but I'm assuming that's not the case.

squarebracket:
Agh, tab spacing :( will fix

squarebracket:
whitespace issues are fixed

src/lib_ccx/params_dump.c
```diff
@@ -18,7 +18,9 @@ void params_dump(struct lib_ccx_ctx *ctx)
 			mprint ("stdin");
 			break;
 		case CCX_DS_NETWORK:
-			if (ccx_options.udpaddr == NULL)
+			if (ccx_options.udpsrc != NULL)
+				mprint ("Network, %s@%s:%d", ccx_options.udpsrc, ccx_options.udpaddr, ccx_options.udpport);
```
anshul1912: check for ccx_options.udpaddr to be null is still missing.
I would suggest to put (ccx_options.udpsrc != NULL) in else if that would make sure if (ccx_options.udpaddr == NULL) has failed

squarebracket: Please see my reply to your previous comment, which was hidden due to white space changes.

squarebracket: Please see my reply to your previous comment, which was hidden due to white space changes.

anshul1912:
@squarebracket : Sorry for incovinence, I arrpove your changes.
Windows and linux difference of outputs test have not passed, can you please sync up with Willem and run them manually

canihavesomecoffee:
@squarebracket If you'd make one small change (e.g. add your improvement to the changelog) and push that, the tests should be run again, and then we can see how it's impacting the program 👍

squarebracket:
Should I just add it as `Unreleased`?

canihavesomecoffee:
Just add it on the next version number or so :)

squarebracket:
build errors are due to the python api stuff. I'd already preparing a PR to fix the build system to not break from the python stuff, would you like me to submit that first?

canihavesomecoffee:
That would be awesome 👍

cfsmp3:
Please reopen (or better, send a new PR) when the build errors are fixed. Other than that the changes seem reasonable.

squarebracket:
Build errors are unrelated to changes.

PR #803, opened 2017-11-01
Update dict_mr_robot.txt

shucon:

---

{pull request content here}


cfsmp3:
Should probably have Darlene and Alderson in separate lines (same in that Elliot one), otherwise Darlene by itself wouldn't be uppercased which is not what we want.
Feel free to correct and resubmit.

shucon:
@cfsmp3 why all other characters have both ,their first and last name?
For example the same thing goes with Elliot because he's not referred as Elliot Alderson every time in the tv-series.

PR #844, opened 2017-12-25
[Fix] Fix DVB bug (Multiple-line subtitle; Missing last line)

### LINKED ISSUE #840: [BUG] unable to extract multiple lines from DVB-sub using OCR
CCExtractor version (using the --version parameter preferably) : **0.85** _and_ **Git commit 18584259447a145a1b8c1cae6733223393dfb4f1**

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

I found https://github.com/CCExtractor/ccextractor/issues/392 which seems like the same issue, but it was closed in mid-2016, and I was able to reproduce the issue on a much newer version than that (including a Git build from today.) Therefore I think it's OK to file this.

Fortunately, the latest Git version seems to have addressed another bug I was going to file. Although the "something messy" instrumentation in you added would have to be filtered out, disabled, or sent to stderr before I could use this version.

**My familiarity with the project is as follows (check one, eg [X]):**

**Necessary information**
- Is this a regression (did it work before)? [x] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [ ] Windows - [x] Linux - [ ] Mac
64-bit CentOS 7.3
- What where the used arguments? `ccextractor -stdout -quiet -nofc -nodvbcolor  ccextractor_bugs_allcaps_29fps_leftjustify.m2ts`

**Video links**

https://s3-us-west-2.amazonaws.com/ccextractor-dvbsub-bugreports/ccextractor_bugs_allcaps_29fps_leftjustify.m2ts

I think I have set the permissions correctly on this S3 bucket, let me know if you can't download it.

The DVB-sub captions in this video display perfectly in VLC (tested verison 2.2.4.) There is a burned in timecode on the video to help you judge the timing, and it seems dead-on to me in VLC.

**Additional information**

There are three problems with ccextractor's output from this file:

- Timing is slightly off (really not a huge deal, it's only off by 10-30 ms)
- The last caption (the twenty-fifth caption) is missing from the output completely
- Only the first line of text from each caption is shown (this is the biggest problem)

When I use spupng output and run the tesseract command line program (version 3.04 tested) on the PNG images, the text is detected just fine, so I don't think it's a limitation of tesseract itself.

If you're curious about the bug that was fixed with the latest Git version, it's that several captions were missing altogether when extracted from this file:  
https://s3-us-west-2.amazonaws.com/ccextractor-dvbsub-bugreports/big_buck_bunny_eac3_4.m2ts  
even though they played just fine in VLC. This bug is fixed now, but you could use the video in your regression tests if you want.

harrynull:

---
Fix #840 
- Fix the bug that only the first line of text from each caption is shown.
- Fix the bug that the last caption is missing.
- The shifting time isn't really a problem because it is indeed shifting a little bit in the video and the conversion from fps->1000 will also cause some precision problems.

### DIFF
--- a/src/lib_ccx/dvb_subtitle_decoder.c
+++ b/src/lib_ccx/dvb_subtitle_decoder.c
@@ -1468,8 +1468,8 @@ static int write_dvb_sub(struct lib_cc_decode *dec_ctx, struct cc_subtitle *sub)
 
 	ctx = (DVBSubContext *) dec_ctx->private_data;
 
-        display_def = ctx->display_definition;
-        sub->type = CC_BITMAP;
+    display_def = ctx->display_definition;
+    sub->type = CC_BITMAP;
 	sub->lang_index = ctx->lang_index;
 
 	if (display_def)
@@ -1551,6 +1551,9 @@ static int write_dvb_sub(struct lib_cc_decode *dec_ctx, struct cc_subtitle *sub)
 				rect->ocr_text = ocr_str;
 			else
 				rect->ocr_text = NULL;
+            if (ccx_options.dvb_debug_traces_to_stdout) {
+                mprint("\nOCR Result: %s\n", rect->ocr_text ? rect->ocr_text : "NULL");
+            }
 		}
 		else
 		{
@@ -1645,10 +1648,12 @@ int dvbsub_decode(struct encoder_ctx *enc_ctx, struct lib_cc_decode *dec_ctx, co
 						segment_length);
 				break;
 			case DVBSUB_DISPLAY_SEGMENT: //when we get a display segment, we save the current page
+
 				if (enc_ctx->write_previous) //this condition is used for the first subtitle - write_previous will be 0 first so we don't encode a non-existing previous sub
 				{
 					sub->prev->end_time = (dec_ctx->timing->current_pts - dec_ctx->timing->min_pts) / (MPEG_CLOCK_FREQ / 1000); //we set the end time of the previous sub the current pts
-					encode_sub(enc_ctx->prev, sub->prev); //we encode it
+                    
+                    encode_sub(enc_ctx->prev, sub->prev); //we encode it
 					enc_ctx->srt_counter = enc_ctx->prev->srt_counter; //for dvb subs we need to update the current srt counter because we always encode the previous subtitle (and the counter is increased for the previous context)
 					enc_ctx->prev_start = enc_ctx->prev->prev_start;
 					sub->prev->got_output = 0;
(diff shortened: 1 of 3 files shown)


src/lib_ccx/ccx_encoders_srt.c
```diff
@@ -83,6 +83,7 @@ int write_stringz_as_srt(char *string, struct encoder_ctx *context, LLONG ms_sta
 
 int write_cc_bitmap_as_srt(struct cc_subtitle *sub, struct encoder_ctx *context)
 {
+    mprint("\nNow subtitle %x is really written into file\n", sub);
```
cfsmp3: Debug messages should use the debug function, not mprint()...

src/lib_ccx/dvb_subtitle_decoder.c
```diff
@@ -1551,6 +1551,9 @@ static int write_dvb_sub(struct lib_cc_decode *dec_ctx, struct cc_subtitle *sub)
 				rect->ocr_text = ocr_str;
 			else
 				rect->ocr_text = NULL;
+            if (ccx_options.dvb_debug_traces_to_stdout) {
+                mprint("\nOCR Result: %s\n", rect->ocr_text ? rect->ocr_text : "NULL");
```
cfsmp3: Same thing... plus debug options are controlled with a mask, not with independent variables.

src/lib_ccx/general_loop.c
```diff
@@ -887,6 +887,7 @@ int general_loop(struct lib_ccx_ctx *ctx)
     }
 
 	end_of_file = 0;
+    struct encoder_ctx *enc_ctx = NULL;
```
cfsmp3: Incorrect identation (unless github is not displaying it right)

src/lib_ccx/ccx_encoders_srt.c
harrynull: Sorry, it's supposed to be temporary but I forgot to remove them.

src/lib_ccx/dvb_subtitle_decoder.c
harrynull: I did not add the option. It is controlled by parameter `-debugdvbsub` and is used many places in the code already.

src/lib_ccx/general_loop.c
```diff
@@ -1135,6 +1135,14 @@ int general_loop(struct lib_ccx_ctx *ctx)
 		if (ccx_options.send_to_srv)
 			net_check_conn();
 	}
+
+    // Process the last subtitle
```
anshul1912: not a good way to fix it, should remove encode_sub from dvb_sub_decoder

src/lib_ccx/dvb_subtitle_decoder.c
```diff
@@ -419,7 +419,7 @@ static void delete_regions(DVBSubContext *ctx)
  *                       it could be -1 if not found in PMT.
  * @param ancillary_id ancillary-page_id found in Subtitle descriptors
  *                     associated with     subtitle stream in the    PMT.
- *                       it could be -1 if not found in PMT.
+ *                       it could be -1 if not found in PMT.
```
anshul1912: why added extra space above

PR #847, opened 2017-12-25
[FIX] Checks for text before newlines on DVB subtitles

gray-v:

---

Resolves the multiple lines issue of https://github.com/CCExtractor/ccextractor/issues/840, basically does the same thing that https://github.com/CCExtractor/ccextractor/pull/844 does but without raising potential issues with newlines as pointed out in the codein task.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,7 +1,3 @@
-0.87
------------------
-- New: Support for source-specific multicast.
-
 0.86
 -----------------
 - New: Added a histogram in one-minute increments of the number of lines in a subtitle.
@@ -22,6 +18,7 @@
 - New: Added -nospupngocr (don't OCR bitmaps when generating spupng, faster)
 - New: Add support for file split on keyframe (-segmentonkeyonly)
 - New: Added WebVTT output from Matroska.
+- New: Support for source-specific multicast.
 - Fix: Prevent the OCR being initialized more than once (happened on multiprogram and
   PAT changes)
 - Fix: Makefiles, build scripts, etc... everything updated and corrected for all
@@ -34,7 +31,6 @@
 - Fix:  When NAL decoding fails, don't dump the whole decoded thing, limit to 160 bytes.
 - Fix: Modify Autoconf scripts to generate tarball for mac from `/package_creators/tarball.sh` 
   and include GUI files in tarball
-
 
 
 0.85b (2017-1-26)


harrynull:
Did you try the program without -nodvbcolor? because dvb color will also break new line. Also, did you consider UTF8 characters when you check for ASCII characters?
Another suggestion is, although I am not sure if it will happen, what if the line only contains 0/1 character?

gray-v:
Yeah ideally I'd like to just check the entire line by scanning backwards to the previous \n or start of message and then checking if there is at least 1 letter of some sort in it but I couldn't quite get it to work because I can't find what marks the actual start of the subtitle line, as I got an infinite loop where it just rolled back and back and back forever.

gray-v:
after discussion on slack wait for #844 to be merged, then I'll update this and put it on top of that

src/lib_ccx/ocr.c
```diff
@@ -296,15 +298,6 @@ char* ocr_bitmap(void* arg, png_color *palette,png_byte *alpha, unsigned char* i
 				int x1, y1, x2, y2;
 				if (!TessPageIteratorBoundingBox((TessPageIterator *)ri, level, &x1, &y1, &x2, &y2))
 					continue;
-				// printf("word: '%s';  \tconf: %.2f; BoundingBox: %d,%d,%d,%d;",word, conf, x1, y1, x2, y2);
```
cfsmp3: Not sure we want to completely remove this. Seems like code that whoever is doing maintenance/debugging may want to enable from time to time.

src/lib_ccx/ocr.c
```diff
@@ -715,8 +707,8 @@ static int quantize_map(png_byte *alpha, png_color *palette,
 			int index = iot[mcit[mxi]];
 			alpha[iot[i]] = alpha[index];
 			palette[iot[i]].red = palette[index].red;
-			palette[iot[i]].blue = palette[index].blue;
-			palette[iot[i]].green = palette[index].green;
+palette[iot[i]].blue = palette[index].blue;
```
cfsmp3: This is breaking indent.

src/lib_ccx/ocr.c
```diff
@@ -728,51 +720,52 @@ static int quantize_map(png_byte *alpha, png_color *palette,
 			i, palette[i].red, palette[i].green, palette[i].blue, alpha[i]);
 	}
 #endif
-	end: freep(&histogram);
+end: freep(&histogram);
```
cfsmp3: Breaking indent.

src/lib_ccx/ocr.c
```diff
@@ -810,14 +803,32 @@ int compare_rect_by_ypos(const void*p1, const void *p2, void*arg)
 
 void add_ocrtext2str(char *dest, char *src, const char *crlf, unsigned crlf_length)
 {
+	char *line_scan;
+	BOOL should_break = FALSE;
```
cfsmp3: I'm not a big fan of BOOL - in C bools are int, 0 is FALSE and everything else is TRUE...

src/lib_ccx/ocr.c
gray-v: Fair point, but printf literally doesn't even work here. I'll change it to a mprint()

PR #876, opened 2018-01-05
[FEATURE] Add FreeType2

harrynull:
Please prefix your pull request with one of the following: **[FEATURE]** **[FIX]** **[IMPROVEMENT]**.

**In raising this pull request, I confirm the following (please check boxes):**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that another pull request for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [X] I am an active contributor to CCExtractor.

---

This pull requests adds a new dependency - FreeType2
It also includes the LICENSE file of FreeType2 (docs/freetype.TXT) (GPL2)

PR #877, opened 2018-01-05
[FEATURE] FreeType-based text renderer (-out=spupng with teletext/EIA608)

harrynull:

---

**ABSTRACT**
The pull request implements a FreeType-based text renderer, which can be used to generate the bitmaps for spupng in the case of teletext/EIA608.

**CHANGES**
- Remove "ccx_encoders_spupng.h" (The old implementation of it) and "spupng_encoder.h" (encoders don't usually have header files)
- Rename "spupng_encoder.c" to "ccx_encoders_spupng.c" (For consistency).
- Add new implementation based on FreeType to "ccx_encoders_spupng.c".
- Add a lot comments/document to the code.

**KNOWN ISSUES**
**Note that the issues will also affect EIA-608 which may not have these issues before** (But the new implementation is generally much cleaner, easier to maintain and more extensible and customizable than the old one).
- <del> Font needs to be manually set.</del> **FIXED** (New option -font added to specify font)
- No color: <font> tag is rendered as-is.
- XML position may be not correct.
- <del>Characters like "$" will be cut.</del> **FIXED**
- <del>Accented letters don't render. (could be encoding problem)</del> **FIXED**

**NOTE**
Please merge this PR after #876 , or it won't compile.

**EXAMPLES**
**Note that the preview is not latest. Please refer to comments for the latest preview.**
sub0001
Example of a good output.

sub0060
Example of tag rendered as-is.

<del>
sub0000
(original generated png file)
sub0000
(transparency is replaced by black)
Example of "$" being cut.
</del>

**FIXED**


saurabhshri:
This is looking good. 🙂  Okay, so couple of questions and points : 

1. Could you please also attach the generate XML file?

2. Does the program generate the png with transparent background or black?

This question is important because these are directly displayed as it is by video players at the position defined by the XML file. For example : 

image

So, if they are of transparent bg, they might not be visible properly, and if the complete rectangular frame is black : they might cover full regions.

This is how usual SPUPNG png looks : 

sub0020

Notice how only the text has filled bg?

2. The rendering of tags, is it the limitation of library or it can be implemented later? Because they can definitely have style in them.

sub0003

3. I did not understand "Example of "$" being cut.". I can see the `$` symbol, what does being cut mean?

4. **"Note that the issues will also affect EIA-608 which may not have these issues before"** could you please point to specifics about what might get broken if this is added. It'll be an important criteria as we'd not like avoid breaking stuff as much as possible.

harrynull:
> Could you please also attach the generate XML file?

I just noticed that the generated xml has incorrect offset values. I will upload them as soon as I fix them.

> Does the program generate the png with transparent background or black?

The program generates png with transparent background. But it won't be hard to change it to black.
I can also clip the png so that it only contains the text.
So which is the one that we want?
transparent background + fixed width
Like sub0000
OR
text with black background + fixed width + transparent background where no text?
Like sub0060
(The old implementation)

> The rendering of tags, is it the limitation of library or it can be implemented later? Because they can definitely have style in them.

FreeType can support colors, styles (bold/italic, by using different .tts), as long as you can parse these styles from tags.

> I did not understand "Example of "$" being cut.". I can see the $ symbol, what does being cut mean?

Although it may be not that obvious, "$" is actually cut a little bit, depending on what font you use.
I've changed a new font which can demonstrate the issue more clearly.
sub0000

> "Note that the issues will also affect EIA-608 which may not have these issues before" could you please point to specifics about what might get broken if this is added. It'll be an important criteria as we'd not like avoid breaking stuff as much as possible.

All the issues that I mentioned.
- Font needs to be manually set.

This one is reasonable I think, since we don't know what font the user wants to use, and we can't specify a default font since we don't know if the user has it.

- Characters like "$" will be cut.

Not really hard to fix. I will see if I can fix it.

- Accented letters don't render. (could be encoding problem)

Same.

- No color: tags are rendered as-is.

It needs parse of tags. I will see how the old code did it.

saurabhshri:
> I just noticed that the generated xml has incorrect offset values. I will upload them as soon as I fix them.

Sure!

> So which is the one that we want?

I think the PNGs like the following would be perfect. They can then be placed anywhere on the screen (with the help of respective XML file).

sub0001

sub0002

sub0003

sub0004

>FreeType can support colors, styles (bold/italic, by using different .tts), as long as you can parse these styles from tags.

If you're not including those changes in this PR, do remember to raise an issue once this is merged so that we don't forget it later on. 😉

harrynull:
A sample of generated xml is attached.
Note that the positioning may be not correct. The offset values are constant now. To get it work, we maybe need to know the size of the video.
I cannot make a video with SPUPNG subtitles because spumux somewhy didn't work for me so I can't test it.
[COMEDY.zip](https://github.com/CCExtractor/ccextractor/files/1609808/COMEDY.zip)

saurabhshri:
Please use relative path in XML instead of full path : 

`/mnt/d/Tests/EIA608/COMEDY.d/sub0000.png` --> `COMEDY.d/sub0000.png`

Also, could you also show how the new pngs look? Attach them here.

harrynull:
The images in the PR has updated to the new pngs.
Relative path issue fixed (though it's not me who introduced the issue)

[teletext.zip](https://github.com/CCExtractor/ccextractor/files/1611156/teletext.zip)

src/CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ include_directories ("${PROJECT_SOURCE_DIR}/extractors/")
 include_directories ("${PROJECT_SOURCE_DIR}/wrappers/")
 include_directories ("${PROJECT_SOURCE_DIR}/libpng/")
 include_directories ("${PROJECT_SOURCE_DIR}/zlib/")
+include_directories ("${PROJECT_SOURCE_DIR}/freetype/include")
```
saurabhshri: Please add trailing slash. It doesn't include all files without it.

src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -1,411 +1,912 @@
...
+/* Closed Caption character cell dimensions */
+#define CCW 16
+#define CCH 26 /* line doubled */
+
+// TODO: To make them customizable
+#ifdef _WIN32
+#define FONT_PATH "C:\\Windows\\Fonts\\calibri.ttf"
+#else
+#define FONT_PATH "/usr/share/fonts/truetype/noto/NotoSans-Regular.ttf"
+#endif
```
saurabhshri: Missing font for macOS. Results in

saurabhshri:
@harrynull Also tt'd would have been better if you commented instead of changing description because, first, it tampers the history and comments seem senseless, and second it's easier to follow things. From next time always comment.

src/lib_ccx/ccx_encoders_spupng.c
harrynull: The font problem is actually one of the issue. I didn't fix it because I think it's part of customization. But yeah, I will fix it now, by adding an option.

harrynull:
@saurabhshri 
Text justification does not break words in newlines. Words are break simply because the given space is not enough to contain them so I break them to a new line so that no subtitles will be lost.
To solve it, you can either make the font smaller, or make the canvas bigger, by changing the value of `FONT_SIZE` and `CANVAS_WIDTH`.
The pixelated output issue is not expected. Could you tell me what font and `FONT_SIZE` and `CANVAS_WIDTH` values you are using?

harrynull:
Nevermind, it's a bug related to my black background algorithm. I've fixed it.

harrynull:
sub0000
The latest preview.

src/lib_ccx/params.c
```diff
@@ -624,8 +624,9 @@ void print_usage (void)
 	mprint ("                       ISO-639-2 form (like \"fre\" for french) or a language\n");
 	mprint ("                       code followed by a dash and a country code for specialities\n");
 	mprint ("                       in languages (like \"fre-ca\" for Canadian French).\n");
-	mprint ("          -nospupngocr When processing DVB don't use the OCR to write the text as");
+	mprint ("          -nospupngocr When processing DVB don't use the OCR to write the text as\n");
 	mprint ("                       comments in the XML file.\n");
+	mprint ("                -font: Specify the font that is used when generating png files\n");
```
saurabhshri: Change it to `Specify the full path of the font that is to be used when generating SPUPNG files`.

saurabhshri:
@cfsmp3 Once that parameter description is edited, it's good for merge, thanks @harrynull - nice job! :)

PR #878, opened 2018-01-05
[FEATURE] Added -man parameter to generate manual page for ccextractor using help2man

mahalwal:

---
I have added the -man parameter which generates the man page using help2man (dependency).
help2man uses the output from --help and --version of CCExtractor's executable.
This pull request supports man pages only for Linux users. I will add support for mac as well as other users after someone provides their feedback on this pull.


src/lib_ccx/params.c
```diff
@@ -1114,6 +1113,15 @@ int parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
...
+		{
+#ifdef __linux__
+			system("bash manscript.sh");
+			mprint("Finished executing man script\n");
+#endif
+			return EXIT_WITH_HELP;
+		}
 		if (strcmp (argv[i], "-")==0 || strcmp(argv[i], "-stdin") == 0)
 		{
 #ifdef WIN32
```
mahalwal: Minimum two spaces were required after every parameter so that help2man is able to detect different options.

src/lib_ccx/params.c
```diff
@@ -1114,6 +1113,15 @@ int parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
...
 		}
+
+		if(!strcmp(argv[i], "-man"))	//Generate manual page
+		{
+#ifdef __linux__
+			system("bash manscript.sh");
+			mprint("Finished executing man script\n");
+#endif
+			return EXIT_WITH_HELP;
+		}
```
cfsmp3: This is a really really weird solution - if ccextractor (the binary) cannot generate the man page directly, then the correct way to generate it would be for manscript.sh to call ccextractor -help to get the output, and then do the transformation from the script, generate the man page, and possibly install it (if a install parameter is present).

Otherwise you're forcing the user to have manscript.sh around. Where would be this file be? ccextractor usually goes in /usr/local/bin or /usr/bin, but a file called manscript.sh doesn't really go there...

src/lib_ccx/params.c
```diff
@@ -295,28 +295,28 @@ void print_usage (void)
...
-	mprint ("                       -stdout will redirect all messages to stderr (error).\n");
-	mprint ("           -pesheader: Dump the PES Header to stdout (console). This is\n");
-	mprint ("                       used for debugging purposes to see the contents\n");
-	mprint ("                       of each PES packet header.\n");
-	mprint ("         -debugdvbsub: Write the DVB subtitle debug traces to console.\n");
-	mprint ("      -ignoreptsjumps: Ignore PTS jumps (default).\n");
-	mprint ("         -fixptsjumps: fix pts jumps. Use this parameter if you\n");
-	mprint ("                       experience timeline resets/jumps in the output.\n");
-	mprint ("               -stdin: Reads input from stdin (console) instead of file.\n");
+	mprint ("            inputfile:  file(s) to process\n");
```
cfsmp3: Why all those extra spaces in the help? Are they needed for the man?
Did you check that every line is still below 80 chars?

linux/manscript.sh
```diff
@@ -0,0 +1,23 @@
...
+    echo "help2man found."
+else
+    echo "help2man does not exist."
+    echo "Please install help2man. (for Ubuntu use: sudo apt-get install help2man)"
+    exit 1
+fi
+
+exechelp2man="help2man --section 7 --include=replace.h2m ./ccextractor > ccextractor.7" 
+echo $exechelp2man
+eval $exechelp2man
```
canihavesomecoffee: Why are you using eval? There are most likely nicer ways available than this.

linux/manscript.sh
```diff
@@ -0,0 +1,23 @@
...
+
+mkdir -p /usr/local/man/man7
+sudo cp ccextractor.7 /usr/local/man/man7/
```
canihavesomecoffee: Assuming a user has sudo available doesn't seem to wise. It'd be better to request the script to be executed with sudo or as root instead, and have a check for it.

src/lib_ccx/activity.c
```diff
@@ -135,6 +135,6 @@ void activity_header (void)
 		credits_shown = 1;
 		mprint ("CCExtractor %s, Carlos Fernandez Sanz, Volker Quetschke.\n", VERSION);
 		mprint ("Teletext portions taken from Petr Kutalek's telxcc\n");
-		mprint ("--------------------------------------------------------------------------\n");
+		// mprint ("--------------------------------------------------------------------------\n");
```
canihavesomecoffee: Why is this line commented out?

src/lib_ccx/params.c
canihavesomecoffee: I agree with this. It seems a weird solution, and it makes quite a lot of assumptions.

src/lib_ccx/params.c
mahalwal: Atleast 2 spaces are required after every param for help2man to differentiate b/w the parameter and the the decription.

src/lib_ccx/activity.c
mahalwal: For proper formatting in the man pages. This was not looking good

linux/manscript.sh
mahalwal: I'll changes this

mahalwal:
I am closing this PR and I'll send another one with the requested changes.
Since, currently help2man requires the binary, the workaround will be to use it with `make` which will only use the output from source file and not the binary.
Also the help2man says this 
> This usage allows a manual page to be generated by the maintainer and included in the distribution without requiring the end-user to have help2man installed.
So is there a way we can update the roff file generated by help2man everytime someone updates the output from --help or --version?
@canihavesomecoffee @cfsmp3

PR #895, opened 2018-01-11
[FIX] Fix Travis CI not executing `linux/build`

harrynull:

---

During installing the dependencies, one of the command forgets to exit the directory:
```
if [[ $TRAVIS_OS_NAME == 'linux' ]]; then wget https://github.com/DanBloomberg/leptonica/releases/download/1.74.4/leptonica-1.74.4.tar.gz && tar xvf leptonica-1.74.4.tar.gz; cd leptonica-1.74.4; ./configure && make && sudo make install; fi
```
It makes `cd linux; ./build; cd ..; fi` fail to execute.

I also changed `install` to `before_install`, as suggested by travis:
> In a before_install step, you can install additional dependencies required by your project such as Ubuntu packages or custom services.

It can make the verbose dependencies installation collapsed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ compiler:
 - gcc
 - clang
 
-install:
+before_install:
 - if [[ $TRAVIS_OS_NAME == 'osx' ]]; then brew install pkg-config autoconf automake libtool tesseract leptonica; fi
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -10,17 +10,16 @@ compiler:
 - gcc
 - clang
 
-
-install:
+before_install:
```
canihavesomecoffee: This should not be changed:

> before_install:
> execute all of the commands which need to be executed before installing dependencies

vs 

> install
> install all of the dependencies you need here

We are clearly installing dependencies here.

PR #906, opened 2018-01-20
[IMPROVEMENT] Update .gitignore

thealphadollar:

---
Updated .gitignore file with the files to be ignored made by CMake installation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,13 @@ src/**/.deps
 src/**/.dirstamp
 mac/ccextractorGUI
 linux/ccextractorGUI
+linux/CMakeCache.txt
+linux/CMakeFiles/
+linux/cmake_install.cmake
+linux/install_manifest.txt
+linux/lib_ccx/
+mac/lib_ccx/
+mac/install_manifest.txt
+mac/cmake_install.cmake
+mac/CMakeFiles/
+mac/CMakeCache.txt
--- a/docs/ccextractor.cnf.sample
+++ b/docs/ccextractor.cnf.sample
@@ -1,7 +1,7 @@
 #######################################################
 # Version 0.01
 #
-# To enable required option please uncommnent option
+# To enable required option please uncomment option
 #
 
 


.gitignore
```diff
@@ -120,3 +120,8 @@ src/**/.deps
 src/**/.dirstamp
 mac/ccextractorGUI
 linux/ccextractorGUI
+linux/CMakeFiles/
+linux/CMakeCache.txt
+linux/cmake_install.cmake
```
MatejMecka: If you are adding CMake Ignore files you should probably add for mac too.

MatejMecka: Isn't it better to Add it as a separate PR?

docs/COMPILATION.MD
```diff
@@ -26,6 +26,25 @@ sudo apt-get install -y libleptonica-dev
 
 2. Compiling
 
+**[Recommended] Using CMake With OCR**
```
MatejMecka: Why Bold? Couldn't you leave it as a header

.gitignore
TheFaR8: It is depend on him to whether add them in this PR or make a new PR

docs/COMPILATION.MD
```diff
@@ -26,6 +26,53 @@ sudo apt-get install -y libleptonica-dev
 
 2. Compiling
 
+**[Recommended] Using the build script with OCR**
+
+
```
TheFaR8: Remove this line please

docs/COMPILATION.MD
```diff
@@ -26,6 +26,53 @@ sudo apt-get install -y libleptonica-dev
...
+# test your build
+./ccextractor
+```
+
+**[Recommended] Standard linux compilation through Autoconf scripts with OCR**
+
+```
+sudo apt-get install autoconf      #Dependency to generate configuration script
+cd ccextractor/linux
```
MatejMecka: Why seperate OCR and Non OCR? Why not something like this?

```Using build.command script:

cd ccextractor/mac
./build.command OCR

# test your build
./ccextractor

If you don't want the OCR capabilities, then you don't need to configure the tesseract and leptonica packages, and build it with just

cd ccextractor/mac
./build.command

# test your build
./ccextractor
```

thealphadollar: @MatejMecka I don't think that makes much of a difference. In the linux one, it is for multiple options and hence I have included it once instead of including it in each of the options. I feel that's a better way, and that way we can also use the recommended tag. What do you say?

MatejMecka: I don't know about the Recommended one :/ Ask someone else for that. But i think it would be nicer like the method I mentioned. Because Right now you have OCR,OCR,OCR and then Normal Compilation, Normal Compilation, Normal Compilation. Normal Compilation then OCR sounds much nicer. So yours should be Normal Compilation, OCR, Normal Compilation, OCR like that.

TheFaR8: I think you can get a same method for first part (like navigate to ccextractor/linux, etc), you can wrap them up into one thing

docs/COMPILATION.MD
```diff
@@ -26,6 +26,53 @@ sudo apt-get install -y libleptonica-dev
...
+make
+
+#Test your build
+./ccextractor
+
+#To install it for use from terminal anywhere
+sudo make install
+```
+
+If you don't want OCR capabilities, please use the below commands.
```
MatejMecka: Why not CMake again? CMake isn't only used for OCR it's another method to compile CCExtractor.

docs/COMPILATION.MD
```diff
@@ -26,6 +26,53 @@ sudo apt-get install -y libleptonica-dev
 
 2. Compiling
 
+**[Recommended] Using the build script with OCR**
```
MatejMecka: No need to add [Recommended] cause not all input files and users will need it.

MatejMecka:
CMake files are ok. If you want you can further improve it by adding autogen files too

thealphadollar:
@MatejMecka I compiled using autoconfig and no new files were generated in my local repo.

Issue #879, opened 2018-01-07
[Bug] Python Bindings don't compile with build_library


MatejMecka:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

CCExtractor version (using the --version parameter preferably) : **0.85**

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X]):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [x] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [ ] Windows - [X] Linux - [] Mac
- What where the used arguments? `N/A // python api_testing.py <pathtovideofile>`

**Video links**

Please make the affected input file available for us (no screenshots, those don't help!). Public links to Dropbox, Google Drive, etc, are all fine. If it is not possible to make it available publicly, send us a private invitation (both Dropbox and Google Drive allow that). In this case we will download the file and upload it to the private developer repository.

Do *not* upload your file to any location that will require us to sign up or endure a wait list, slow downloads, etc. If your upload expires make sure you keep it active somehow (replace links if needed). Keep in mind that while we go over all tickets some may take a few days, and it's important we have the file available when we actually need it.

**Additional information**

`ImportError: No module named _ccextractor`

Steps to reproduce:
1. Install bindings via pip
2. Clone repo from github 
3. Go into the API directory and do `./build_api`
4. Run `api_testing.py`
5. Error

[...]

PR #916, opened 2018-01-24
[Fix] fix the python extension brokens (#879)

lennonwoo:

---

### ABSTRACT

The pull request fix the python extension brokens, now we can use python bindings to deal with the subtitle extracted from video again.

### CHANGES

- We now only use PYTHON_API to preprocessing code for pythoon bingding.
- Remove the swig auto-generated files which are ccextractor.py and ccextractor_wrap.c
- Remove signal_python_api and python_subs_array
- Delete extractor directory and move the function in extractor to ccx_encoders_python.c

### MY THOUGHTS

I want to talk about why I remove signal_python_api and python_subs_array. 
First, signal_python_api, I notice that this flag variable is used to determine whether we should continue to execute the original ccextractor processing logic or not. But the problem is, do we really need to process original logic when we are using python to deal more detail subtitle things? We could do this thing just use the compiled binary file.
Second, the original code introduce python_subs_array as global variable to deal with the time judgement, this become boresome when we have to free the arrays in the end of api_start fucntion.

### KNOWN ISSUES

- Currently, the updated scripts only compiled with python2.7 as well as python test scripts
- I introduce custom asprintf function, I have no idea whether it's right thing to do or not.

### USAGE

Use `./build_library` in api directory to build the pythoon binding, then use `python2 api_testing.py options...` to test

I have test build script under Linux, the output example in python is as follow

> srt_counter-203                                                                                                                                                                                
start_time-00:09:51,158  end_time-00:09:54,526                                                                                                                                                 
text[203]:                                                                                                                                                                                     
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR
text[203]:                                    
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:                                
color[203]:99999999999999999999999999999999
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:           AND THAT WAY WE COULD
color[203]:99999999999000000000000000000000
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
text[203]:          KEEP EACH OTHER WARM?"
color[203]:99999999990000000000000000000000
font[203]:RRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRR 
***END OF FRAME***


MatejMecka:
Python2 is going to no longer be maintained after 2020. So I think the scripts should be made to work with Python3

src/lib_ccx/ccx_encoders_python.c
```diff
@@ -3,16 +3,70 @@
...
+	int ret;
+	va_list ap;
+
+	va_start(ap, fmt);
+	ret = vasprintf(strp, fmt, ap);
+	va_end(ap);
+
+	if (ret == -1) {
+		printf("Error: Some problem with asprintf return value in extractor.c\nExiting.");
+		exit(500);
```
cfsmp3: Should be CCX_COMMON_EXIT_BUG_BUG (a #define in ccx_common_common_h) rather than a magic number.
But do we need to define this function in the first place?

src/lib_ccx/ccx_encoders_python.c
```diff
@@ -3,16 +3,70 @@
 #include "ccx_encoders_common.h"
 #include "ccx_encoders_helpers.h"
 
-#ifdef ENABLE_PYTHON
+#ifdef PYTHON_API
+void Asprintf(char **strp, const char *fmt, ...) {
```
cfsmp3: Uppercase?

src/lib_ccx/ccx_encoders_python.c
lennonwoo: Yes, CCX_COMMON_EXIT_BUG_BUG should be better. In the first place? do you mean put the error check handle in the raw asprintf function which locate in utilify.c then we move Asprintf func?

lennonwoo:
I'm sorry I don't have test it under Windows environment before PR. I will test it later, but need to be familiar with Windows building flow for a while.

lennonwoo:
@MatejMecka I agree with you, but currently, the build scripts can't figure out user's python version, only choose -l python2.7 to link specificly.

cfsmp3:
@lennonwoo I've edited that minor thing with the exit code and fixed an issue with the Windows build. It does build at least, possibly Windows+Python is broken (was before) so if linux works again that's already an improvement.

Issue #922, opened 2018-01-28
[BUG] DVB Teletext subtitle incomplete


mkver:
CCExtractor version (using the --version parameter preferably) : 0.86 (Git commit 5fa83394a0c0300244eb5707b30d344d7687bb2c)

**In raising this issue, I confirm the following (please check boxes, eg [X] - and delete unchecked ones):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**

- [x] I absolutely love CCExtractor, but have not contributed previously.

**Necessary information**
- Is this a regression (did it work before)? [x] NO | [ ] YES
- What platform did you use? [x] Windows - [ ] Linux - [ ] Mac
- What were the used arguments? `ccextractorwin.exe  --gui_mode_reports  -autoprogram  -out=srt -bom -utf8 -tpage 150`

[Link to the file](https://www.dropbox.com/s/0o2ncppc0hq8ljt/DVB-Teletext%20incomplete.ts?dl=0)
**Additional information**

When I extract the teletext from page 150 of the sample file (that actually contains another subtitle at page 799, but that is irrelevant), the subtitles are incomplete:
```
1
00:00:00,200 --> 00:00:04,559
<font color="#00ffff">Sie können ihn nicht von der Schule</font>
<font color="#00ffff">schmeißen! Schulpflicht!</font>

2
00:00:05,080 --> 00:00:09,599
Für diese Klassenstufe ist er nicht
<font color="#00ffff">Stufen Sie ihn zurück!</font>
```
The second subtitle is incomplete. Here is what the libzvbi teletext decoder inside ffmpeg produces:
```
1
00:00:00,240 --> 00:00:04,640
[...]

PR #926, opened 2018-02-08
[FIX] DVB Teletext subtitle incomplete #922

BPYap:

---
Attempt to solve issue #922 by replacing 0xB and 0xA in the middle of row with space character

Generated srt:
fixed

Playing the generated srt on windows 10 Movies & TV app:
geeignet

### DIFF
--- a/src/lib_ccx/telxcc.c
+++ b/src/lib_ccx/telxcc.c
@@ -637,25 +637,39 @@ void process_page(struct TeletextCtx *ctx, teletext_page_t *page, struct cc_subt
         // anchors for string trimming purpose
         uint8_t col_start = 40;
         uint8_t col_stop = 40;
-        uint8_t last_replacement_index = 0;
 
+        uint8_t last_replacement_index = 0;
+        int box_open = 0;
         for (int8_t col = 0; col < 40; col++)
         {
-            if (col_start == 40 && page->text[row][col] == 0xb)
+            // replace all 0/B and 0/A characters with 0/20, as specified in ETS 300 706:
+            // Unless operating in "Hold Mosaics" mode, each character space occupied by a
+            // spacing attribute is displayed as a SPACE
+            if (page->text[row][col] == 0xb) // open the box
             {
-                col_start = col;
-                line_count++;
+                if (col_start == 40)
+                {
+                    col_start = col;
+                    line_count++;
+                }
+                else
+                {
+                    page->text[row][col] = 0x20;
+                }
+                box_open = 1;
             }
-            else if (page->text[row][col] == 0xb || page->text[row][col] == 0xa)
             {
-                // replace all 0/B and 0/A characters with 0/20, as specified in ETS 300 706:
-                // 'Unless operating in "Hold Mosaics" mode, each character space occupied by a
-                // spacing attribute is displayed as a SPACE
                 page->text[row][col] = 0x20;
                 last_replacement_index = col; // remember the last index of replacement
+                box_open = 0;
+            }
+            // characters between 0xA and 0xB shouldn't be displayed
+            else if (!box_open && col_start < 40)
+            {
+                page->text[row][col] = 0x20;
             }
         }
-        // change the value of last position to 0xa to indicate end of subtitle
+        // put 0xa back to last_replacement_index
         if (last_replacement_index)
         {
             page->text[row][last_replacement_index] = 0xa;
--- a/src/lib_ccx/telxcc.c
+++ b/src/lib_ccx/telxcc.c
@@ -634,43 +634,25 @@ void process_page(struct TeletextCtx *ctx, teletext_page_t *page, struct cc_subt
 	// process data
 	for (uint8_t row = 1; row < 25; row++)
 	{
-		// anchors for string trimming purpose
-		uint8_t col_start = 40;
-		uint8_t col_stop = 40;
+        // anchors for string trimming purpose
+        uint8_t col_start = 40;
+        uint8_t col_stop = 40;
 
-		for (int8_t col = 39; col >= 0; col--)
-		{
-			if (page->text[row][col] == 0xb)
-			{   
-                // check for double 0/B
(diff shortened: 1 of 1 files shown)


mkver:
I looked at your code and it seems that characters between the end of a box and the start of a new box (which are not intended to be displayed) nevertheless end up in the output file. In the case of my samples, all these characters are spaces so that this is fine with them, but there might be others.
How about the following code (disclaimer: I am no programmer and have no means to check whether the following code even compiles etc.):
```
                int box_open = 0; //indicates whether there is currently a box open.
                // anchors for string trimming purpose
		uint8_t col_start = 40;
		uint8_t col_stop = 39;
                
                for (int8_t col = 0; col <= 39; col++)
                {
                        if (page->text[row][col] == 0xb)
                        {
                               box_open = 1;
                               page->text[row][col] = 0x20;
                        }

                        if (page->text[row][col] == 0xa) 
                        {
                                box_open = 0;
                                page->text[row][col] = 0x20;
                        }

                       if (box_open && (col_start == 40) && (page->text[row][col] > 0x20))
		       {
				col_start = col;
				line_count++;
				break;
		       }
                       
                        if (!box_open && (col_start < 40) && (page->text[row][col] > 0x20))
		                page->text[row][col] = 0x20; //no character outside a box should be displayed
               }

		// line is empty
		if (col_start > 39)
			continue;
                
		for (uint8_t col = 39; col >= col; col--)
                {
                        if (page->text[row][col] > 0x20)
                        {
                                  col_stop = col;
                                  break;
                        }
                 }
```

Some remarks:

- The above is intended to replace the lines 637-667 of the current version (without BPYap's patch).
- Just like the current version, simple 0xb suffice as a start code. It would be possible to change this, but since no one complained about thte current behaviour, I thought it best not to change it in this regard.
- All characters in between two boxes end up as spaces.
- Currently, if a line contains a 0xb, but no character >0x20 to the right of it, the line_count is incremented, but no subtitle is emitted for this row. The if-statement at lines 676-689 might introduce unwanted linebreaks in this situation. The above changes this behaviour: The line is only treated as nonempty if there is a character >0x20 inside a box. I don't know if this actually happens in the wild.
- Colour information is not processed/touched above.

BPYap:
@mkver , thanks for pointing out the issue that characters between 0xA and the next 0xB shouldn't be displayed. My third patch failed to address this situation and I think this is the main reason that caused regression test to fail just now. I will try incorporate your idea of checking and setting characters appearing between 0xA and the next 0xB to 0x20.

BPYap:
@mkver @cfsmp3  I guess the issue is finally resolved, please have a look at my latest commit. 😆 
For the failed Windows CI tests, I have compared the results with the other 2 pending PRs. It seems that all PRs failed for these regression IDs: 2, 19, 21, 238, 239, 182, 214, 73, 75, 78 82.

mkver:
I read your commit and am wondering about something: Imagine you have a situation like the following: 0xB 0xB something 0xA (potentially something) 0xB 0xB something different than 0xA (i.e. the last box is not closed by a 0xa, but by the end of the line). The last_replacement_index will then point to the place of 0xA and at [line 678](https://github.com/BPYap/ccextractor/blob/93859297c1c3a01023e5372f15b3e68ad45e8a43/src/lib_ccx/telxcc.c#L678) there will again be a 0xA in there. And the [break in the next loop](https://github.com/BPYap/ccextractor/blob/93859297c1c3a01023e5372f15b3e68ad45e8a43/src/lib_ccx/telxcc.c#L692) implies that col_stop will contain the position of the last character >0x20 before the 0xA. This in turn means that nothing of the last box will end up in output.    
Apart from that I don't really see any need to use this last_replacement_index at all: All characters > 0x20 outside of boxes are replaced with 0x20 anyway so that the [first if statement in the second loop](https://github.com/BPYap/ccextractor/blob/93859297c1c3a01023e5372f15b3e68ad45e8a43/src/lib_ccx/telxcc.c#L685) is can only be entered if the current position is the one of a non-space character inside a box.

BPYap:
Thank you again @mkver, for your patience and pointing out potential flaws in the code.
Here's what I changed in the latest commit:
- removed last_replacement_index related statements
- remove  _if (page->text[row][col] == 0xa) break;_  statement from second loop (since all 0xA characters are replaced with 0x20 in the first loop)
- defined box_open as **uint8_t** type, and use _YES / NO_ for its value

cfsmp3:
There's some errors according to the CI... @canihavesomecoffee can you check if this is caused by this specific PR?

thealphadollar:
@cfsmp3 and @canihavesomecoffee Just to add in some information (maybe this could help), the same issues were present in my PR as well and I had changed only the python modules in the api folder. So, I do not think that these warnings and errors are specific to this PR :)

cfsmp3:
@mkver Are you able to test this PR with a number of files and see how it behaves?

mkver:
If someone can make a build for me, yes.

mkver:
Yes. Sorry that I forgot to mention it.

cfsmp3:
@mkver Here you are:
https://drive.google.com/open?id=1xxBwXeJlms_ogc5LhlOsqZP1ONW6T7xh

Just replace the existing binaries with those.

mkver:
Is the BPYap's patch really included in your build? It says that its version is "Git commit: 550d3207adf19cb7035bce9cf7309f6e8d583d14" and the issue that should be fixed by #930 is indeed fixed, but the output of the samples I provided for #922 is unchanged.

cfsmp3:
@mkver you're right - I did everything except actually merge it. Sorry about that.
I've uploaded correct versions to the same location.

mkver:
The new files don't have any missing text any more, i.e. the patch is good. I'll close #922. (There is just one line where the builds with the fix for #930 contain one space where libzvbi has two. Strange, but I can live with that. Maybe it contains a control character other than a color code or 0xA and 0xB.)

Issue #929, opened 2018-02-13
Terrible OCR results with Channel 5 (UK)


cfsmp3:
(current master, pre 0.87)

This file (but well, all of channel 5)
https://drive.google.com/open?id=1Etq-pv5G3jGqVhhRl7cNrfuw4gaKkLoV

Produces terrible results in the OCR, even though the bitmaps seem normal. What's going on?

```
833
00:59:26,021 --> 00:59:29,340
(CriminaLs den't.just. ’cemmit
I one type of offence. I

834
00:59:29,341 --> 00:59:31,700
Just C05 they SitQLe same petrQL
- that day‘ dQeSn't mean -

835
00:59:31,701 --> 00:59:34,740
m
pietes Of pLLa nt,
```

cfsmp3:
GSOC qualification: 5 points

harrynull:
The problem is in `quantize_map(alpha, palette, rect->data[0], size, 3, rect->nb_colors);` (ocr.c:773).
Comment this line out can solve the problem.

With `quantize_map`:
![rawdo1](https://user-images.githubusercontent.com/7413706/36134478-084affe8-10c0-11e8-8e24-32602612171f.png)

Without `quantize_map`:
![raw](https://user-images.githubusercontent.com/7413706/36134414-aa4d0bd4-10bf-11e8-9dca-c2776b2b6b8b.png)

By the way, for reference of my own and maybe someone who wants to look into OCR, I suggest to add  `save_spupng("debug.png", indata, w, h, palette, alpha, 16);` somewhere at the beginning of the `ocr_bitmap` so that it's easier to know what the raw image looks like without any further process. I don't want to open a PR to add just a line for debugging though.

Abhinav95:
This is pretty interesting. The quantize_map() function in itself was important (from my discussions with @anshul1912 ) in order to improve the DVB results. What the function does essentially is to kind of 'binarize' the input image into text and non text regions and ignore the gradient grayscale values at the edge of the text and non text regions. With these particular set of subtitles, it seems like the binarization process is leading to some unwanted noisy artifacts around the text regions, which is throwing off the OCR results. This could probably be solved by an additional filtering step to remove the 'salt noise' present in the current images.

thealphadollar:
@Abhinav95 I think I can look in this issue if you don't mind :)

Abhinav95:
@thealphadollar Go right ahead :)

cfsmp3:
Things we could do:

[...]

PR #932, opened 2018-02-21
[IMPROVEMENT] Modify -quant 0 option

thealphadollar:

---
This pull request aims at simplifying and making the -quant 0 parameter faster. It does so by reducing the number of distinct color shades available in the palette of the PNG image under processing. 

An example would be, say a palette color is (248, 187, 027). The new algorithm reduces this image to (224, 160, 0). This is one quantised value amongst the (8*8*8) which can be formed under this algorithm. Since the color palette is reduced and our rect bitmap structures point to palette for their pixel color value; in a way this algorithm decreases the color value to the nearest multiple of 32 for R,G,B; effectively quantising it without much reduction in the actual image visibility. 

As can be seen in the below screenshot, this method improves the time taken (92 seconds vs 100 seconds) and also gives a better result than no quantisation. 

With 8-bit color reduction

As is marked in the below screenshot (with no quant algorithm at all), a dialogue is read as "Are you off the deck". This is a mistake which does not happen and is read as "Are you off the clock" with the improved algorithm. The video is provided along with the timestamp to check for authenticity.

without color reduction

video screenshot

At last below is a diff between the two files to see the other error corrections that the method provides, (for example it correctly reads "I" (capital "i") which were read as "|")

difference between the two extracted subtitles

For video file, refer issue #929 

Minor Addition: Added below line which can be uncommented to output debug.png from ocr_bitmap() function
`save_spupng("debug.png", indata, w, h, palette, alpha, 16);`

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -191,6 +191,9 @@ BOX* ignore_alpha_at_edge(png_byte *alpha, unsigned char* indata, int w, int h, 
 
 char* ocr_bitmap(void* arg, png_color *palette,png_byte *alpha, unsigned char* indata,int w, int h, struct image_copy *copy)
 {
+	// uncomment the below lines to output raw image as debug.png iteratively
+	//save_spupng("debug.png", indata, w, h, palette, alpha, 16);
+
 	PIX	*pix = NULL;
 	PIX	*cpix = NULL;
 	PIX *color_pix = NULL;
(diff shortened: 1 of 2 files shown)


cfsmp3:
- case 0 means "pass the image unchanged to tessearact", so it cannot do any kind of processing.
- You can add that new code to case 2 (but test that it actually helps in some cases, otherwise no point in adding it)
- Remember that the help text needs to be updated as well.

thealphadollar:
The proposed algorithm improves speed and gives slightly better results as is shown in the PR content.

Updated help and made it case 2 :)

PR #948, opened 2018-02-28
[FIX] Implicit declaration of these functions throws warning during build

krushanbauva:

- [ x] I have checked that another pull request for this purpose does not exist.

- [ x] I have used CCExtractor just a couple of times.
- [ x] I absolutely love CCExtractor, but have not contributed previously.

---

Should I declare these functions in a header file and include the header file rather than doing it this way?

P.S.: I used the build script on my linux machine which throws warning

### DIFF
--- a/src/gpacmp4/isom_write.c
+++ b/src/gpacmp4/isom_write.c
@@ -40,6 +40,8 @@ GF_Err CanAccessMovie(GF_ISOFile *movie, u32 Mode)
 }
 
 s32 gf_lang_find(const char *);
+const char *gf_lang_get_3cc(u32);
+
 
 static GF_Err unpack_track(GF_TrackBox *trak)
 {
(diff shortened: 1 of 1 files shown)


src/gpacmp4/isom_write.c
```diff
@@ -39,6 +39,10 @@ GF_Err CanAccessMovie(GF_ISOFile *movie, u32 Mode)
 	return GF_OK;
 }
 
+s32 gf_lang_find(const char *);
```
MatejMecka: `s32` ?

cfsmp3:
I'm merging because I'm tired of seeing those warnings myself, even though I'm not a fan of changing libgpac directly. 
It would be a good idea to also send this minor thing to the libgpac maintainers.

krushanbauva:
Also, I'm planning to go through all those warnings one by one and fix them! :)

krushanbauva:
@cfsmp3 This issue doesn't arise in gpac library. The function is already declared in a header file `gpac/include/gpac/iso639.h` which is included in the source code.

PR #950, opened 2018-03-03
[FIX] Fix segfault in add_cc_sub_text and initialize to NULL in init_encoder 

saurabhshah0410:

---

This commit adds some checks to avoid segmentation faults.

* In `add_cc_sub_text()`, strdup will cause a segfault if we duplicate an
  empty string.

* In `init_encoder()`, initialize pointer fields to NULL to avoid random
  addressing so we can avoid illegal memory accessing and segfaults in
  other places.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -1038,6 +1038,7 @@ struct encoder_ctx *init_encoder(struct encoder_cfg *opt)
 	ctx->segment_last_key_frame = 0;
 	ctx->nospupngocr = opt->nospupngocr;
 
+	ctx->prev = NULL;
 	return ctx;
 }
 
--- a/src/lib_ccx/ccx_common_common.c
+++ b/src/lib_ccx/ccx_common_common.c
@@ -70,6 +70,8 @@ void freep(void *arg)
 int add_cc_sub_text(struct cc_subtitle *sub, char *str, LLONG start_time,
 		LLONG end_time, char *info, char *mode, enum ccx_encoding_type e_type)
 {
+	if (str == NULL || strlen(str) == 0)
+		return 0;
 	if (sub->nb_data)
 	{
 		for(;sub->next;sub = sub->next);


src/lib_ccx/ccx_encoders_common.c
```diff
@@ -924,7 +924,8 @@ void dinit_encoder(struct encoder_ctx **arg, LLONG current_fts)
 		write_subtitle_file_footer(ctx, ctx->out + i);
 	}
 
-	free_encoder_context(ctx->prev);
+	if(!ccx_options.hardsubx) // Don't free ctx->prev in case of hardsubx to avoid segfault
+		free_encoder_context(ctx->prev);
```
cfsmp3: Why is that call to free_encoder_context segfaulting? That's what we should fix, not just not do the call because it segfaults... is it because ctx->prev is null, because it ends in a double-free, or what?

This seems like hiding the problem, not fixing it.

saurabhshah0410:
@cfsmp3 Yes, you are right. I didn't go deep enough to investigate the cause of the segfault before your comment.

So the `prev` field of an `encoder_ctx` is used to process dvb subtitles. They are not at all used in the code that involves processing hard subtitles. And when an `encoder_ctx` is initialized in `init_encoder` function, the `prev` field is not set to `NULL` and hence, it gets some random value(address). So, `ctx->prev` != `NULL` and  `free_encoder_context` tries to free the fields of `ctx->prev`, which do not exist, and hence a segfault. So this fix is wrong and the actual fix is very simple. Please check my latest fix.

src/lib_ccx/ccx_encoders_common.c
saurabhshah0410: @cfsmp3 Yes, you are right. I didn't go deep enough to investigate the cause of the segfault before your comment.

So the prev field of an encoder_ctx is used to process dvb subtitles. They are not at all used in the code that involves processing hard subtitles. And when an encoder_ctx is initialized in init_encoder function, the prev field is not set to NULL and hence, it gets some random value(address). So, ctx->prev != NULL and free_encoder_context tries to free the fields of ctx->prev, which do not exist, and hence a segfault. So this fix is wrong and the actual fix is very simple. Please check my latest fix.

PR #963, opened 2018-03-12
[FIX] Fix possible segfaults in hardsubx_classifier.c due to strdup

### LINKED ISSUE #928: [BUG] Error while using "-hardsubx"

CCExtractor version (using the --version parameter preferably) : **0.87**

- [X] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [X] I have checked that the issue I'm posting isn't already reported.
- [X] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have used the latest available version of CCExtractor to verify this issue exists.

**Necessary information**
- Is this a regression (did it work before)? [ ] NO | [ X] YES - *please specify the last known working version*
- What platform did you use? [ ] Windows - [ X] Linux - [ ] Mac
- What were the used arguments? `-hardsubx -subcolor white -detect_italics -whiteness_thresh 90 -conf_thresh 60`

**Video links (replace text below with your links) **

https://drive.google.com/open?id=0B_61ywKPmI0TeU5rOTlCMWxKbW8

**Additional information**

The argument "-hardsubx" is giving an error when asked to process the burned in subtitles of the attached video file. Below are the screenshots of the error.

screenshot from 2018-02-11 18-47-06

screenshot from 2018-02-11 18-47-14

PS: Make sure you set an alert in GitHub so you get notifications about your ticket. We may need to ask questions and we do everything inside GitHub's system.

saurabhshah0410:

---

`strdup` will give a segmentation fault if the argument passed to it is
`NULL`. `TessResultIteratorGetUTF8Text` returns a `char*` which can be `NULL`
and we should not call `strdup` directly over it. Once we check if the
value returned is not `NULL`, then we can call `strdup`.

This also fixes #928.

### DIFF
--- a/src/lib_ccx/hardsubx_classifier.c
+++ b/src/lib_ccx/hardsubx_classifier.c
@@ -48,9 +48,12 @@ char *get_ocr_text_wordwise(struct lib_hardsubx_ctx *ctx, PIX *image)
 	{
 		do
 		{
-			char* word = strdup(TessResultIteratorGetUTF8Text(it, level));
-			if(word==NULL || strlen(word)==0)
+			char* word;
+			char* ts_word = TessResultIteratorGetUTF8Text(it, level);
+			if(ts_word==NULL || strlen(ts_word)==0)
 				continue;
+			word = strdup(ts_word);
+			TessDeleteText(ts_word);
 			if(text_out == NULL)
 			{
 				if(ctx->detect_italics)
@@ -136,9 +139,12 @@ char *get_ocr_text_letterwise(struct lib_hardsubx_ctx *ctx, PIX *image)
 	{
 		do
 		{
-			char* letter = strdup(TessResultIteratorGetUTF8Text(it, level));
-			if(letter==NULL || strlen(letter)==0)
+			char* letter;
+			char* ts_letter = TessResultIteratorGetUTF8Text(it, level);
+			if(ts_letter==NULL || strlen(ts_letter)==0)
 				continue;
+			letter = strdup(ts_letter);
+			TessDeleteText(ts_letter);
 			if(text_out==NULL)
 			{
 				text_out = strdup(letter);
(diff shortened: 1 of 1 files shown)


src/lib_ccx/hardsubx_classifier.c
```diff
@@ -48,9 +48,10 @@ char *get_ocr_text_wordwise(struct lib_hardsubx_ctx *ctx, PIX *image)
 	{
 		do
 		{
-			char* word = strdup(TessResultIteratorGetUTF8Text(it, level));
+			char* word = TessResultIteratorGetUTF8Text(it, level);
 			if(word==NULL || strlen(word)==0)
 				continue;
+			word = strdup(word);
```
cfsmp3: This doesn't seem correct to me (but the previous one isn't either).

word = strdup(word);

We're losing the original pointer - assuming TessResultIteratorGetUTF8Text() allocates something (I assume it does, but I don't know for sure) we now have a leak.

Using strdup() seems OK to me IF we're going to be messing with the data later, but in this case we should immediately free the tesseract resource after making a copy, since we no longer need it.

Check this out:
http://tess4j.sourceforge.net/docs/docs-1.2/net/sourceforge/tess4j/TessAPI.html#TessBaseAPIGetIterator(net.sourceforge.tess4j.TessAPI.TessBaseAPI)


    TessBaseAPIGetIterator

    TessAPI.TessResultIterator TessBaseAPIGetIterator(TessAPI.TessBaseAPI handle)

    Get a reading-order iterator to the results of LayoutAnalysis and/or Recognize. ****. WARNING! This class points to data held within the TessBaseAPI class, and therefore can only be used while the TessBaseAPI class still exists and has not been subjected to a call of Init, SetImage, Recognize, Clear, End, DetectOS, or anything else that changes the internal PAGE_RES.

saurabhshah0410: Oh yes, I didn't realize that I was creating a memory leak. So something like:
```
char *word_ = TessResultIteratorGetUTF8Text(it, level);
if(word_ == NULL || !strlen(word_))
    continue;
char *word = strdup(word_);
free(word_);
```
This should work fine right?

cfsmp3: No... you can't use free() to unallocate resources returned by TessResultIteratorGetUTF8Text() - take a look at the link I sent. Most likely (but check - I just browsed the API) it's TessResultIteratorDelete.

Issue #961, opened 2018-03-09
Request: Add support to extract telemetry data (stored as subtitles) from recordings


cfsmp3:
[This file](https://drive.google.com/file/d/1MyABc8cByEWcW9wQkl_Y_JIUYGC-Kd_5/view?usp=sharing) is a MP4 recorded from a drone (specifically: "Zenmuse camera on a DJI Inspire 2 Drone"). The telemetry data is saved as subtitles and in fact VLC can display it.

CCExtractor doesn't find it - possible it's stored in a MP4 box we still don't handle, or something like that. We do show a number of "unknown box type", even though we're using a fairly new libGPAC.

**GSoC qualification: 6 points**

cfsmp3:
After some digging - this is just EXIF data. Not sure CCExtractor is the right tool to actually extract it (since it has nothing to do with subtitles after all), since exiftool does it perfectly already. We could however detect its presence and display some notification.

Anyway this tool processes the sample file just fine and reports everything:
https://www.sno.phy.queensu.ca/~phil/exiftool/

thetransformerr:
This was the dump I got from atom "moov/udta" present in above said file where the data was  stored,

```
udta(?xyz?+26.483061-80.124142+38.000?xsp?-0.40?ysp?+0.10?zsp?+0.00?fpt?+2.80?fyw?+82.90?frl?-9.50?gpt?-27.70?gyw?+83.50?grl?+0.00?rec ?repPmeta hdlrmdir?ilst̩cmt?dataDE=TrueColor, Mode=P, DSW=0001Xtra?
           WM/Categorv01.07.2091.1.v1.0.0?uid??QPd ?mdlFC65100?csndb5c42c8e7d3048193df558fbdc77dfb(?aud??mux????free 
```

thetransformerr:
They might be the lines where we can make some changes:

https://github.com/CCExtractor/ccextractor/blob/39b96cc54493d7462702b31195be84c1936551f2/src/gpacmp4/mp4.c#L631

or

https://github.com/CCExtractor/ccextractor/blob/39b96cc54493d7462702b31195be84c1936551f2/src/gpacmp4/mp4.c#L420

thealphadollar:
@cfsmp3 What changes are we looking forward to? I might be able to finish this with some lead and more information.

cfsmp3:
I'd start with detecting the presence of EXIF data, which shouldn't be too
hard.

thealphadollar:
[...]

PR #974, opened 2018-04-19
[IMPROVEMENT] Update libGPAC

thealphadollar:

---

We use libGPAC for all our MP4 operations and, this commit updates it to the latest version.

All previous changes to the original library were restored post straight file updation and bugs have been removed.

concerns #961 

The updated libGPAC supports the file present in the referenced issue but, unfortunately, the new library is detecting that there are no subtitles in the file.

screenshot from 2018-04-20 02-05-12

**change severity:** *very high*


thealphadollar:
@cfsmp3 I inspected the log files for the tests and it seems the problem,

- For linux fail is not with this PR only. Rather I'd say it has been improved since the fail has been reduced to only DVB from DVB, Options and Teletext.

- For Windows the build is failing due to some error in CCExtractor GUI which I don't think is concerned with libGPAC directly.

Is it safe to merge the PR?

NilsIrl:
Is there a guide somewhere on how to upgrade `libGPAC`?

Looking at https://github.com/gpac/gpac/ , the source is a lot different than what is in gpacmp4.

Where can I get more information?

cfsmp3:
Probably not worth the effort until we get a .mp4 file that our current libGPAC doesn't support.

NilsIrl:
Actually that's why I'm asking #1053

cfsmp3:
@NilsIrl I don't really suspect libgpac, more likely a Windows file system issue. 
Anyway if the user says it's fixed we just move on :-)

PR #994, opened 2018-07-17
[IMPROVEMENT] Don't add files with empty filename & show better msg for multiple input files.

saurabhshri:

---

If CCExtractor is passed an empty argument, it tries to open it as a file. E.g.
`ccextractor ""` OR
`ccextractor {file1} "" {file2}`

In case of multiple input files, if one or more failed to open, CCExtractor exists with the message `Failed to open input file: File does not exist.` Since my second parameter was an empty filename `""` (came from an empty key in a script), I kept on thinking somehow "," is being appended to filename. 
It showed : `Input: filename,`.

With this PR:

- If filename is empty, that file is skipped.
- If one or more file fails to open, it tells that.
- Shows number of files while dumping parameters.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -122,7 +122,7 @@ int api_start(struct ccx_s_options api_options)
             switch (ctx->total_inputsize)
             {
                 case -1*ENOENT:
-                    fatal(EXIT_NO_INPUT_FILES, "Failed to open input file: File does not exist.");
+                    fatal(EXIT_NO_INPUT_FILES, "Failed to open one of the input file(s): File does not exist.");
                 case -1*EACCES:
                     fatal(EXIT_READ_ERROR, "Failed to open input file: Unable to access");
                 case -1*EINVAL:
--- a/src/lib_ccx/params_dump.c
+++ b/src/lib_ccx/params_dump.c
@@ -11,8 +11,9 @@ void params_dump(struct lib_ccx_ctx *ctx)
 	switch (ccx_options.input_source)
 	{
 		case CCX_DS_FILE:
+			dbg_print(CCX_DMT_VERBOSE, "Files (%d): ", ctx->num_input_files);
 			for (int i=0;i<ctx->num_input_files;i++)
-				mprint ("%s%s",ctx->inputfile[i],i==(ctx->num_input_files-1)?"":",");
+				mprint ("%s%s",ctx->inputfile[i],i==(ctx->num_input_files-1)?"":", ");
 			break;
 		case CCX_DS_STDIN:
 			mprint ("stdin");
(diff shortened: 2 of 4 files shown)


src/ccextractor.c
```diff
@@ -122,7 +122,7 @@ int api_start(struct ccx_s_options api_options)
             switch (ctx->total_inputsize)
             {
                 case -1*ENOENT:
-                    fatal(EXIT_NO_INPUT_FILES, "Failed to open input file: File does not exist.");
+                    fatal(EXIT_NO_INPUT_FILES, "Failed to open one or more input file(s): File does not exist.");
```
cfsmp3: Is this accurate? It doesn't seem possible that we can fail to open more than once since we are fatal()'ing on the first failure.

src/lib_ccx/params_dump.c
```diff
@@ -11,8 +11,9 @@ void params_dump(struct lib_ccx_ctx *ctx)
 	switch (ccx_options.input_source)
 	{
 		case CCX_DS_FILE:
+			mprint("Files (%d): ", ctx->num_input_files);
 			for (int i=0;i<ctx->num_input_files;i++)
-				mprint ("%s%s",ctx->inputfile[i],i==(ctx->num_input_files-1)?"":",");
+				mprint ("%s%s",ctx->inputfile[i],i==(ctx->num_input_files-1)?"":", ");
```
cfsmp3: This seems more debug that general, maybe mprint() is too much?

saurabhshri: What would you suggest? :)

src/ccextractor.c
saurabhshri: You're right, I am sorry. Updated to `Failed to open one of the input file(s): File does not exist.`

thealphadollar:
@saurabhshri Ping.

cfsmp3:
@saurabhshri don't use mprint :-) If you want to write debug messages, use dbg_print instead.

saurabhshri:
@cfsmp3 Oh, I misunderstood earlier! Done. :)

src/lib_ccx/params_dump.c
cfsmp3: What do we use for debug messages? :-)

saurabhshri: @cfsmp3 I changed this (https://github.com/CCExtractor/ccextractor/pull/994/files#diff-20a92dffa6342caa6a83fd990812e4cbR14) to debug message. Should I change the current line (L16) as well? It was already in mprint though, and I think having the ability to see the files being processed is fine without running it in debug mode.

thealphadollar:
@cfsmp3 can we merge this?

PR #1014, opened 2018-10-24
Update UTF8proc to 2.2.0

MatejMecka:

---

Updated UTF8proc to 2.2.0 released on Jul 24 2018

Changelog has been updated as well


docs/CHANGES.TXT
```diff
@@ -64,6 +64,7 @@
 - Fix: Colors of DVB subtitles with depth 2 broken due to a missing break.
 - Fix: CEA-708: Caption loss due to CW command (#991).
 - Fix: CEA-708: Update patch for windows priority with functions (#990).
+- Fix: Update UTF8Proc to 2.2.0
```
cfsmp3: This should go into a new version (since 0.87 was released)
So add something like
0.88 (unreleased)
-------------------

At the top and then your line.

cfsmp3:
Why delete those exports? If you check the source code you'll see that they resolve to empty IF 

#ifdef UTF8PROC_STATIC
extern "C" {	#  define UTF8PROC_DLLEXPORT
#endif	#else
...
#endif

So the proper solution here is to just #define UTF8PROC_STATIC where appropriate.
Let's not mess with utf8prof source code unless we absolutely have to :-) Otherwise updating it is going to be a nightmare every time.

MatejMecka:
I upgraded utf8proc last year with #822 and DLLEXPORT broke windows. The fix seemed to be to delete UTF8PROC_DLLEXPORT and CCExtractor didn't crash

Issue #995, opened 2018-07-29
[BUG] SDL: failed to load image: ../icon/home.png


chopchipper443:
CCExtractor version: 0.85 (git commit d5f522c8c5edd0bf3e70edd6ce26f634320e4a4c)

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**
- [x] I absolutely love CCExtractor, but have not contributed previously.

**Necessary information**
- Is this a regression (did it work before)? [ ] NO | [ ] YES | [x] MAYBE? 
- What platform did you use? [ ] Windows - [x] Linux - [ ] Mac
- What were the used arguments? 'ccextractorGUI'

**Additional information**
Trying to use ccextractorGUI, running into the mentioned title issue.
- Built ccextractor and ccextractorGUI from latest git master, on linux with all required dependencies. Builds fine.
```
git clone https://github.com/CCExtractor/ccextractor
cd ccextractor/linux
./autogen.sh && ./configure --with-gui && make
```
- Program runs proper by running ./ccextractorGUI in ./ccextractor/linux/
- Trying to run ccextractorGUI in any other folder, including /usr/bin/, results in the mentioned error:
[SDL]: failed to load image: ../icon/home.png
- Manually copying the entire ./ccextractor/icon folder somewhere else, along with ccextractorGUI, and then running ccextractorGUI, results in a segfault core dump.

Not sure if this is some sort of known issue elsewhere, or if there was an option missing to build required external files into an independent, static file, but would like to solve this issue if possible. Many thanks.

starlord1311:
@chopchipper443 can i take up this issue?

saurabhshri:
[...]

PR #1049, opened 2018-12-02
[FIX] #995 for Linux and Mac

MatejMecka:

---

{pull request content here}


src/GUI/ccextractorGUI.c
```diff
@@ -260,31 +279,35 @@ int main(void)
...
-    media.icons.music_file = icon_load("../icon/music.png");
-    media.icons.font_file =  icon_load("../icon/font.png");
-    media.icons.img_file = icon_load("../icon/img.png");
-    media.icons.movie_file = icon_load("../icon/movie.png");
+	const char* path_of_executable = get_executable_path();
+	char path_of_ccextractor[strlen(path_of_executable)-20]; // This is the size of: mac/./ccextractorGUI and linux/ccextractorGUI
+	strncpy(path_of_ccextractor, path_of_executable, strlen(path_of_executable));
+	path_of_ccextractor[strlen(path_of_ccextractor)-20]='\0'; // Fails without null pointer
+	media.icons.home = icon_load(concat(path_of_ccextractor,"icon/home.png"));
+	media.icons.directory = icon_load(concat(path_of_ccextractor, "icon/directory.png"));
```
cfsmp3: This is leaking memory - concat is doing a malloc() but there's no free() anywhere, and you're not holding a reference. 
A simple solution would be to write a function called cc_icon_load (or a better name you come up with) that just takes the name of the icon ("home") for example and adds everything else (path_of_executable) + ".png", then frees everything, and just returns the same same as icon_load().
A *much better* solution would be to write a simple program that takes all the icons and generates a .h file that contains the icons themselves (not the filenames; the binary data) in an array so we can generate a binary and contains the icons completely - so we don't have to have any additional file.

src/GUI/ccextractorGUI.c
```diff
@@ -165,6 +166,24 @@ void drop_callback(GLFWwindow* window, int count, const char **paths)
 //	nk_draw_text(canvas, space, "Hardsubs Extraction: Yes", 24, &font->handle, nk_rgb(88, 81, 96), nk_rgb(0, 0, 0));
 //}
 
+char* get_executable_path(){
```
cfsmp3: This doesn't seem really portable... what about Windows?

src/GUI/ccextractorGUI.c
```diff
@@ -606,16 +629,16 @@ int main(void)
...
-                    nk_label(ctx, concat("Output type: ", output.type[output.type_select]), NK_TEXT_LEFT);
-                    if(output.is_filename)
-                        nk_label(ctx, concat("Output path: ", output.filename), NK_TEXT_LEFT);
-                    else
-                        nk_label(ctx, "Output path: Default", NK_TEXT_LEFT);
-                    if(burned_subs.is_burned_subs)
-                        nk_label(ctx, "Hardsubtitles extraction: Yes", NK_TEXT_LEFT);
-                    else
-                        nk_label(ctx, "Hardsubtitles extraction: No", NK_TEXT_LEFT);
+					nk_label(ctx, concat("Input type: ", input.type[input.type_select]), NK_TEXT_LEFT);
```
cfsmp3: Leaks here... 
Maybe write a printf() function that works on labels - check the source code of mprintf for example.
You could do a nk_label_printf (ctx, NK_TEXT_LEFT, "Input type: %s%s", input.type[input.type_select]);
which would be a lot neater. 
That function in fact could be contributed to nuklear, since it would be quite useful for everybody.

src/GUI/ccextractorGUI.c
MatejMecka: I don't have a Windows Machine to test it on :(

MatejMecka: There is a Method for windows too: https://docs.microsoft.com/en-us/windows/desktop/api/libloaderapi/nf-libloaderapi-getmodulefilenamea but i don't have a machine to test it :) If someone volunteers i'll gladly implement it

src/GUI/ccextractorGUI.c
MatejMecka: I didn't do anything here other than convert the mixed identation from spaces and tabs to tabs only. I'll try my best to fix this

src/GUI/ccextractorGUI.c
MatejMecka: While working on the better solution stbi uses fopen to read the file so unless we rewrite the library the function is the better method won't work in the situation

src/GUI/ccextractorGUI.c
```diff
@@ -165,6 +166,24 @@ void drop_callback(GLFWwindow* window, int count, const char **paths)
...
+        char *resp = (char *) malloc(sizeof(char) * 3);
+        ssize_t len = readlink("/proc/self/exe", buf, sizeof(buf));
+        buf[len]='\0';
+        resp = buf;
+        return resp;
+    #elif __APPLE__
+        char buf[256];
+        uint32_t size = sizeof(buf);
+        if (_NSGetExecutablePath(buf, &size) == 0){
+            return buf;
```
cfsmp3: You can't return this buf, it's an stack variable.

PR #1050, opened 2018-12-07
**[FIX]** for bug "failed to load image: ../icon/home.png" (issue #995) (Linux and Windows)

ShouryaAggarwal:

- [ ] I have checked that another pull request for this purpose does not exist. (I knew about a 2 months old pull request mentioned in comments of bug and just found out about another pull request, was 5 days old and had errors but fixed for mac and windows. I'm open to collab)
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).** --- Changelog's format was according to releases and dates, I didn't understand where to put this small bug fix, under which date or release.

---

Fixed the mentioned bug (issue #995) for lunix and windows. Tested properly on linux but not so greatly on windows.


ShouryaAggarwal:
Ok, I'll do that.

ShouryaAggarwal:
@cfsmp3 I made a function to get the icon address from icon name, so what it returns is passed the parameter to icon_load for different icons.

cfsmp3:
@ShouryaAggarwal this causes the problem that you're using a global variable so now your function is no longer thread_safe.

Instead just write a function that generates the path *and* loads the icon (and then returns that) rather than one function to create the path, returns it from a static variable...

ShouryaAggarwal:
@cfsmp3 thanks for pointing out the thread safe aspect.

PR #1057, opened 2018-12-18
[FIX] #995 for All platforms and removed the icons folder in source package

ShouryaAggarwal:

---
Made a singe function to take icon name and load the icon and also modified the relative loading of config files by the GUI executable.

This is actually a reopen of my previous pull request (#1050) as I had to change branch names in order to work on another issue and it screwed up my repo. Also I noticed that the general method required to return full path to executable in windows doesn't work if a 32-bit compiled application calls that function on a 64-bit OS, so in that case I've left the program to behave as it previously was.

I welcome any suggestions to help me fix the issue on Mac.


src/GUI/ccextractorGUI.c
```diff
@@ -593,7 +645,7 @@ int main(void)
 			nk_layout_space_begin(ctx, NK_STATIC, 10, 1);
 			nk_layout_space_end(ctx);
 
-			//Extraction Information
+			//Extraction Info`rmation
```
cfsmp3: Why?

src/GUI/ccextractorGUI.c
```diff
@@ -14,10 +14,18 @@
...
+#elif __linux__
+	#include <unistd.h>
+	#include <libgen.h>
+#endif
+
 
 #define PATH_LENGTH 66
 #define NAME_LENGTH 56
 #define PREFIX_LENGTH_TRUNCATED 10
+#define MAXPATHLEN 500            /* make this larger if you need to. */
```
cfsmp3: Who's you, the final user that is handed a binary?

src/GUI/ccextractorGUI.c
```diff
@@ -168,6 +181,48 @@ void drop_callback(GLFWwindow* window, int count, const char **paths)
 int main(void)
 {
 
+	/* Get directory of current executable file */
+	#ifdef _WIN32
+		GetModuleFileName(NULL, fullpath, MAXPATHLEN);
```
cfsmp3: We should have a binary array of icons prebuilt so we don't have to deal with _WIN32 etc in the primary program.
Just write a simple tool that does the icon_load and generates a files with the binary data in an array of chars, such as 

struct icon
{
char *icon_name;
char *icon_data;
unsigned icon_data_length;
};

And then an array of those with the data of all the existing icons.
It will make a smaller distribution package and it will prevent all compilation and runtime issues between platforms.
You can generate the data file in your own computer so you don't need to worry so much about it working or not working everywhere.

ShouryaAggarwal: If I am successful in making that data file, still I would store it some place and I'll still need the absolute path of the executable in order to fix the original issue of gui executable not being able to run from different working directory. Also I need the directory of executable to load the configuration file too. What should be done now to so that we can get the path to directory of executable without ifdef ?

cfsmp3: @ShouryaAggarwal It needs to be a C array that is compiled with the program, not an external file to be loaded externally. Just generate the C file using an external program that does what you need (and this program doesn't need to look for paths, it can assume that the icons are in the same place as itself) and then compile it as any other source file.

ShouryaAggarwal: @cfsmp3 I made the C file containing binary arrays for icons and the font and now I use those arrays to create temporary png files which are loaded by the external stbi library used and then those temporary files are deleted. This now removes the need of having icons folder and also platform dependency left. 
Is that okay now ?

cfsmp3: @ShouryaAggarwal Check this out:

https://github.com/nothings/stb/blob/master/stb_image.h

You can see that it has function to import images directly from a memory buffer. So there's no need to write and then read an intermediate file at all.

Other than that, this is the right approach - I mean, having the icons in a .c file so we don't need to have them around (on the system that only expect the binary, of course the source icons and the icons.c generator tool must be part of the source code tree).

ShouryaAggarwal: @cfsmp3 Thank you for pointing it out and sorry that I couldn't find that API call myself. I optimized it now. The font and icons are now loaded directly from memory.

cfsmp3:
Looking better. But the pr should include the program to generate the. c
file and all required files. Otherwise no one would be able to add more
icons or replace.

src/GUI/ccextractorGUI.c
ShouryaAggarwal: @cfsmp3 I have the documented binary to C array converter C program script. Where should it be placed in the repo ? In order to add more icons or replace something, anyone would need to compile that C program and run it on the resource to get the C array which then is to be copied to src/GUI/icon_data.c and then the variable should be included appropriately in ccextractorGUI.c

I have mentioned the above written process in the bin2c.c command line help so running "bin2c -h" after compiling the C program would display the necessary steps to include the resource in project. Will that be enough ? Where should I keep the bin2c.c file ?

cfsmp3: Place the program in tools/icons2c (create it).
Also don't delete the icons, obviously they have to be there.
You program (unless it's really generic) could be called like this:

./icon2c path_to_icons >icons.c

Which would just look for all the files in that path and create the file via stdout redirection.

I think that's all, obviously you've done the hard part already so now it's a matter of making our own life easier.

ShouryaAggarwal: Done.

Issue #1034, opened 2018-11-01
[QUESTION]How to get Teletext pages with possible subtitles without actual extraction


workflowsguy:
CCExtractor version (using the --version parameter preferably) : **0.87**

**In raising this issue, I confirm the following:**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [X] I have checked that the issue I'm posting isn't already reported.
- [X] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows:**

- [X] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [X] NO
- What platform did you use? [ ] Windows - [ ] Linux - [X] Mac
- What were the used arguments? `-out=report`

**Additional information**

When running `ccextractor` against a video file, I get an output e.g. like this:
```
Notice: Teletext page with possible subtitles detected: 152
- No teletext page specified, first received suitable page is 152, not guaranteed
Notice: Teletext page with possible subtitles detected: 888
Notice: Teletext page with possible subtitles detected: 151
Notice: Teletext page with possible subtitles detected: 150
Notice: Teletext page with possible subtitles detected: 889
```

I would like to get this information without `ccextractor` automatically extracting the subtitles because I need to specify the actual page to extract from based on service name and desired language.
I thought the option `-out=report` would achieve this, but for the same video file I get the output
```
//////// Program #10302: ////////
DVB Subtitles: No
Teletext: Yes
Pages With Subtitles: 
ATSC Closed Caption: Yes
EIA-608: No
CEA-708: No

MPEG-4 Timed Text: No
```
Is this a bug/limitation in the `report` parameter or is there a different way to achieve this?

MakarovGCI2018:
@workflowsguy you can easily extract captions to console only with -stdout parameter.
Note that ccextractor will ignore -o, -o1 and -o2 in this case.
[...]

PR #1058, opened 2018-12-19
[FEATURE] displaying possible teletext pages without actual extraction

MakarovGCI2018:

---
According to #1034 added ability to print possible teletext pages in -out=report mode.

### DIFF
--- a/src/lib_ccx/general_loop.c
+++ b/src/lib_ccx/general_loop.c
@@ -672,12 +672,14 @@ int process_data(struct encoder_ctx *enc_ctx, struct lib_cc_decode *dec_ctx, str
 	else if (data_node->bufferdatatype == CCX_TELETEXT)
 	{
 		//telxcc_update_gt(dec_ctx->private_data, ctx->demux_ctx->global_timestamp);
-		if (enc_ctx) {
+		if (enc_ctx) 
+		{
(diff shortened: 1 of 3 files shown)


src/lib_ccx/general_loop.c
```diff
@@ -677,6 +677,10 @@ int process_data(struct encoder_ctx *enc_ctx, struct lib_cc_decode *dec_ctx, str
 			if (ret == CCX_EINVAL)
 				return ret;
 		}
+		else {
```
anshul1912: generally in code, if you check else statement is written like
```
if(condition)
{
    //statement
}
else
{
    \\statement
}
```
we do not do use  K&R coding style in this file
```
if (condition) {
    //statement
} else {
    //statement
}
```

src/lib_ccx/telxcc.c
```diff
@@ -907,13 +907,13 @@ void process_telx_packet(struct TeletextCtx *ctx, data_unit_t data_unit_id, tele
 			if (!ctx->seen_sub_page[thisp])
 			{
 				ctx->seen_sub_page[thisp]=1;
-				mprint ("\rNotice: Teletext page with possible subtitles detected: %03d\n",thisp);
+				printf ("\rNotice: Teletext page with possible subtitles detected: %03d\n",thisp);
```
anshul1912: printf are really NO, I would suggest use dbg_print (CCX_DMT_TELETEXT,

Issue #668, opened 2017-01-25
Windows and multicast


cfsmp3:
We're hitting exactly this problem:

http://stackoverflow.com/questions/6140734/cannot-bind-to-multicast-address-windows

Suppose this real scenario: We have a number of multicast streams. They are in different addresses, for example 

239.168.100.1:5000
239.168.100.2:5000

This means that we cannot just bind to any interface (0.0.0.0) on port 5000, because then we get things mixed up.

The obvious solution is to bind to the complete IP+port. In linux this works fine, but Windows says "IP address not available", i.e. you cannot bind the socket to the multicast IP address.

For now (because I needed to get it working) I've done the following:

```
#if _WIN32
	// Doesn't seem correct, if there's more than one multicast stream with the same
	// port number we get corruption - IP address needs to be specified, but 
	// in Windows we get an error 10049 (cannot bind).
	// http ://stackoverflow.com/questions/6140734/cannot-bind-to-multicast-address-windows
	servaddr.sin_addr.s_addr = htonl(IN_MULTICAST(addr) ? INADDR_ANY : addr);
#else
	servaddr.sin_addr.s_addr = htonl(addr);
#endif
```
Which solves the problem in linux and leaves Windows behaving as before.

It's definitely not the right solution.
http://stackoverflow.com/questions/10692956/what-does-it-mean-to-bind-a-multicast-udp-socket

I'm assigning this to @Izaron since he did the last network changes if I recall correctly.

cfsmp3:
GSoC: This issue gives 2 qualification points.

sagark123:
Having more than 1 interfaces in windows is a problem as windows tends to pick up the wrong address.We can get all the interfaces by using getifaddrs() and selecting the required interface

'struct ifaddrs *addrs,*tmp;

getifaddrs(&addrs);
tmp = addrs;

while (tmp)
{
    if (tmp->ifa_addr && tmp->ifa_addr->sa_family == AF_PACKET)
        printf("%s\n", tmp->ifa_name);

    tmp = tmp->ifa_next;
}

freeifaddrs(addrs);`

PR #1059, opened 2018-12-21
[FIX] for issue #668 (Windows and Multicast)

ShouryaAggarwal:

---

So I read that in windows we need to bind to INADDR_ANY only while joining the multicast group, so I left it at that only. Now while receiving the stream from udp multicast, I peak at the header file and match the source ip with the provided ip and only continue if the IPs match. I have tested it on windows using ffmpeg stream system. Let me know if any more changes need to be done.

### DIFF
--- a/src/lib_ccx/file_functions.c
+++ b/src/lib_ccx/file_functions.c
@@ -7,9 +7,6 @@ long FILEBUFFERSIZE = 1024*1024*16; // 16 Mbytes no less. Minimize number of rea
 #ifdef _WIN32
 WSADATA wsaData = {0};
 int iResult = 0;
-int isWin = 1;
-#else
-int isWin = 0;
 #endif
 
 LLONG get_file_size (int in)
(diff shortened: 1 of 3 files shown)


src/lib_ccx/file_functions.c
```diff
@@ -386,12 +390,35 @@ size_t buffered_read_opt (struct ccx_demuxer *ctx, unsigned char *buffer, size_t
...
+					addr = INADDR_ANY;
+				}
+
 				if (ccx_options.input_source == CCX_DS_FILE || ccx_options.input_source == CCX_DS_STDIN)
 					i = read (ctx->infd, ctx->filebuffer + keep, FILEBUFFERSIZE-keep);
 				else if (ccx_options.input_source == CCX_DS_TCP)
 					i = net_tcp_read(ctx->infd, (char *) ctx->filebuffer + keep, FILEBUFFERSIZE - keep);
-				else
-					i = recvfrom(ctx->infd,(char *) ctx->filebuffer + keep, FILEBUFFERSIZE - keep, 0, NULL, NULL);
+				else {
```
anshul1912: Create a function for udp data reading in file src/lib_ccx/networking.c similar to net_tcp_read.

src/lib_ccx/file_functions.c
```diff
@@ -7,6 +7,9 @@ long FILEBUFFERSIZE = 1024*1024*16; // 16 Mbytes no less. Minimize number of rea
 #ifdef _WIN32
 WSADATA wsaData = {0};
 int iResult = 0;
+int isWin = 1;
```
anshul1912: Dont make it dynamic for finding windows or linux, if code is windows specific then at compile time only include or exclude the code. Use #ifdef instead of isWin

ShouryaAggarwal: I had a dynamic toggle for this since in windows also there are only a the case of source multicast where this is needed, so instead of nested if inside ifdef I reduced code space. Never the less, I'll do that.

src/lib_ccx/file_functions.c
ShouryaAggarwal: Ok. Sure.

ShouryaAggarwal: @anshul1912 Is it okay now ?

PR #1062, opened 2019-01-10
[IMPROVEMENT] Add --noempty

navimakarov:

---

{pull request content here}


src/lib_ccx/output.c
```diff
@@ -52,6 +52,11 @@ int temporarily_open_output(struct ccx_s_write *wb)
 
 int init_write (struct ccx_s_write *wb, char *filename, int with_semaphore)
 {
+#ifdef _WIN32
+    #define empty "nul"
+#else
+    #define empty "/dev/null"
+#endif
```
cfsmp3: This is not really portable... 
A better solution would be not to open the file at all (not fake it by opening a null file) until we're ready to actually write to it.

Issue #1076, opened 2019-02-22
[BUG] Output Type webvtt prints styling


ggnull35:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

CCExtractor version (using the --version parameter preferably) : git master

**In raising this issue, I confirm the following (please check boxes, eg [X] - and delete unchecked ones):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [ ] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**

- [ ] I have never used CCExtractor.
- [x] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [ ] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [ ] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [x] Windows - [ ] Linux - [ ] Mac
- What were the used arguments? `-autoprogram`

**Video links (replace text below with your links) **

Please make the affected input file available for us (no screenshots, those don't help!). Public links to Dropbox, Google Drive, etc, are all fine. If it is not possible to make it available publicly, send us a private invitation (both Dropbox and Google Drive allow that). In this case we will download the file and upload it to the private developer repository.

Do *not* upload your file to any location that will require us to sign up or endure a wait list, slow downloads, etc. If your upload expires make sure you keep it active somehow (replace links if needed). Keep in mind that while we go over all tickets some may take a few days, and it's important we have the file available when we actually need it.

**Additional information**

{issue content here, replace this line with your issue content}

PS: Make sure you set an alert in GitHub so you get notifications about your ticket. We may need to ask questions and we do everything inside GitHub's system.

-------------
This command prints styling:
[...]

PR #1092, opened 2019-06-05
[FIX] Remove webvtt styling when not using webvtt-full

rayfoss:

---

addresses #1076 

leaves behind the null terminator bug, but it hasn't been an issue with anything except gedit as of yet.

### DIFF
--- a/src/lib_ccx/ccx_encoders_webvtt.c
+++ b/src/lib_ccx/ccx_encoders_webvtt.c
@@ -225,7 +225,7 @@ int write_webvtt_header(struct encoder_ctx *context)
 		char* outline_css_file = (char*)malloc((strlen(css_file_name) + strlen(webvtt_outline_css)) * sizeof(char));
 		sprintf(outline_css_file, webvtt_outline_css, css_file_name);
 		write (context->out->fh, outline_css_file, strlen(outline_css_file));
-	} else {
+	} else if (ccx_options.use_webvtt_styling) {
 		write(context->out->fh, webvtt_inline_css, strlen(webvtt_inline_css));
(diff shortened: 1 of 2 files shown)


rayfoss:
Samples of what this produces from
http://samples.ffmpeg.org/ffmpeg-bugs/trac/ticket4616/Full%20Episode.mp4

[webvtt-full.vtt.txt](https://github.com/CCExtractor/ccextractor/files/3259173/webvtt-full.vtt.txt)
[webvtt.vtt.txt](https://github.com/CCExtractor/ccextractor/files/3259174/webvtt.vtt.txt)

cfsmp3:
On a quick visual inspection it seems like webvtt_create_css and use_webvtt_styling are mutually exclusive. If yes, why do we need 2 variables? 

@saurabhshri can you check this out? (the PR otherwise seems OK to me)

rayfoss:
Create css outputs a file with the styles

src/lib_ccx/ccx_encoders_webvtt.c
```diff
@@ -492,7 +492,8 @@ int write_cc_buffer_as_webvtt(struct eia608_screen *data, struct encoder_ctx *co
 				}
 
 				// write current text symbol
-				write(context->out->fh, &(context->subline[j]), 1);
+				if (context->subline[j] != '\0')
+					write(context->out->fh, &(context->subline[j]), 1);
```
cfsmp3: Is this correct, shouldn't we write a space instead?

rayfoss: There are the same number of spaces in the srt as the post patch vtt. It seems to me the null char is some residue from pointer handling.

PR #1097, opened 2019-06-25
[FEATURE] Added support for encoding into an MCC File. (CCExtractor#733)

robdt:

---

{pull request content here}

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,3 +1,7 @@
+0.89 (TBD)
+----------
+- New: Added support to output captions to MCC file (#733).
+
 0.88 (2019-05-21)
 -----------------
 - New: More tapping points for debug image in ccextractor.
--- a/docs/COMPILATION.MD
+++ b/docs/COMPILATION.MD
@@ -14,6 +14,7 @@ git clone https://github.com/CCExtractor/ccextractor.git
 
 1. Make sure all the dependencies are met.
 
+Debian:
 ```
 sudo apt-get install -y libglew-dev
 sudo apt-get install -y libglfw3-dev
@@ -24,6 +25,18 @@ sudo apt-get install -y tesseract-ocr
 sudo apt-get install -y tesseract-ocr-dev
 sudo apt-get install -y libleptonica-dev
 ```
+
+RHEL:
+```
+yum install -y glew-devel
+yum install -y glfw-devel
+yum install -y cmake
+yum install -y gcc
+yum install -y libcurl-devel
+yum install -y tesseract-devel
+yum install -y leptonica-devel
+```
+
 **Note:** On Ubuntu Version 18.04 (Bionic) and (probably) later, install `libtesseract-dev` rather than `tesseract-ocr-dev`, which does not exist anymore.
 
 **Note:** On Ubuntu Version 14.04 (Trusty) and earlier, you should build leptonica and tesseract from source
--- a/linux/Makefile.am
+++ b/linux/Makefile.am
@@ -155,6 +155,8 @@ ccextractor_SOURCES = \
 				../src/lib_ccx/ccx_encoders_g608.c \
 				../src/lib_ccx/ccx_encoders_helpers.c \
 				../src/lib_ccx/ccx_encoders_helpers.h \
+				../src/lib_ccx/ccx_encoders_mcc.c \
+				../src/lib_ccx/ccx_encoders_mcc.h \
 				../src/lib_ccx/ccx_encoders_python.c \
 				../src/lib_ccx/ccx_encoders_sami.c \
 				../src/lib_ccx/ccx_encoders_smptett.c \
--- a/mac/Makefile.am
+++ b/mac/Makefile.am
@@ -155,6 +155,8 @@ ccextractor_SOURCES = \
 				../src/lib_ccx/ccx_encoders_g608.c \
 				../src/lib_ccx/ccx_encoders_helpers.c \
 				../src/lib_ccx/ccx_encoders_helpers.h \
+				../src/lib_ccx/ccx_encoders_mcc.c \
+				../src/lib_ccx/ccx_encoders_mcc.h \
 				../src/lib_ccx/ccx_encoders_python.c \
 				../src/lib_ccx/ccx_encoders_sami.c \
 				../src/lib_ccx/ccx_encoders_smptett.c \
--- a/src/gpacmp4/mp4.c
+++ b/src/gpacmp4/mp4.c
@@ -6,6 +6,7 @@
 #include "lib_ccx.h"
 #include "utility.h"
 #include "ccx_encoders_common.h"
+#include "ccx_encoders_mcc.h"
 #include "ccx_common_option.h"
 #include "ccx_mp4.h"
 #include "activity.h"
@@ -35,8 +36,10 @@ static int process_avc_sample(struct lib_ccx_ctx *ctx, u32 timescale, GF_AVCConf
 	u32 i;
 	s32 signed_cts=(s32) s->CTS_Offset; // Convert from unsigned to signed. GPAC uses u32 but unsigned values are legal.
 	struct lib_cc_decode *dec_ctx = NULL;
+    struct encoder_ctx *enc_ctx = NULL;
 
 	dec_ctx = update_decoder_list(ctx);
+	enc_ctx = update_encoder_list(ctx);
 
 	set_current_pts(dec_ctx->timing, (s->DTS + signed_cts)*MPEG_CLOCK_FREQ/timescale);
 	set_fts(dec_ctx->timing);
@@ -65,7 +68,7 @@ static int process_avc_sample(struct lib_ccx_ctx *ctx, u32 timescale, GF_AVCConf
 		temp_debug=0;
 
 		if (nal_length>0)
-			do_NAL (dec_ctx, (unsigned char *) &(s->data[i]) ,nal_length, sub);
+			do_NAL (enc_ctx, dec_ctx, (unsigned char *) &(s->data[i]) ,nal_length, sub);
 		i += nal_length;
 	} // outer for
 	assert(i == s->dataLength);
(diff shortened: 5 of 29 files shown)


canihavesomecoffee:
Hi @robdt,

Can you take a look at these errors for the Windows build environment?

```
"ccextractor\windows\ccextractor.sln" (default target) (1) ->
"ccextractor\windows\ccextractor.vcxproj" (default target) (2) ->
(Link target) ->
  mp4.obj : error LNK2019: unresolved external symbol _mcc_encode_cc_data referenced in function _process_clcp [ccextractor\windows\ccextractor.vcxproj]
  ccx_decoders_common.obj : error LNK2001: unresolved external symbol _mcc_encode_cc_data [ccextractor\windows\ccextractor.vcxproj]
  Debug\ccextractorwin.exe : fatal error LNK1120: 1 unresolved externals [ccextractor\windows\ccextractor.vcxproj]
 
    1188 Warning(s)
    3 Error(s)
```

Looks like the file should be just added to the .sln :+1:

robdt:
Hey @canihavesomecoffee,

Thanks! It looked like it should be added into the Visual C Project file, so I added it there. There weren't any other files in the .sln like there were in the .vcxproj. I am hoping that is correct. I am still looking at why there are issues with the tests. I think it is related to the command line option I added. Please let me know if you see that it didn't fix the problem with Windows. I am still a little handicapped there as I don't have access to a Windows box or Visual Studio to build and test.

canihavesomecoffee:
Hi @robdt,

Thanks for the changes! We're down to one issue now: the mcc file requires `uuid/uuid.h`, which is not available on Windows.

As this library is not natively available (afaik) for Windows, we might have to look into a solution as mentioned here: https://github.com/djp952/prebuilt-libuuid/issues/1#issuecomment-427341280

@cfsmp3 What would you prefer to do?

cfsmp3:
The less dependencies the better - if we can copy/paste the code for those functions from a well tested source I'd do that, unless there's a good reason not to.

robdt:
@canihavesomecoffee - That would be easy to pull in. In fact I could make that the code for all platforms if that made more sense than leaving Linux and Mac as they are. Let me know if you want me to proceed down that path.

cfsmp3:
@robdt  yes please do that - let's use that code for all platforms so we can avoid the dependency. I find it helps a lot with portability (and building, in general).

robdt:
Thanks @cfsmp3, will do!

robdt:
@canihavesomecoffee @cfsmp3 - I had been redirected, but I am now able to get back to this.

I was re-looking at everything, and I am not sure why I thought that I could leverage the code described in https://github.com/djp952/prebuilt-libuuid/issues/1#issuecomment-427341280 for anything beyond windows. It is clearly based on the windows call `CoCreateGuid()`. Anyway, it still has the requisite functionality for Windows, so my plan was to:
1) Pull the two files from that project into src/lib_ccx as two new files. Maybe win_uuid.cpp/win_uuid.h
2) Leverage preprocessor directives (specifically `_WIN32`) to include the header and function for windows builds
3) Add the new files to the windows build, but not necessarily the others as the current implementation will work for Mac and Linux

This stops me at some of the CMake stuff. The files I am leveraging are C++ files. Is it possible to port that file with the .cpp extension and the build system is smart enough to get it compiled? Or do I have to do something special to get this integrated. Sorry, windows dev isn't my thing so I am not super literate around it.

cfsmp3:
@canihavesomecoffee can you check why the build fails and fix if you can? (I'm unable to build on Windows these days). Feel free to merge when you're happy.

canihavesomecoffee:
Hi @robdt, the PR looks fine. I'll try and build it tomorrow (and see why the CI isn't building), could you take a look at the results from Travis? It seems that for Mac, in 2 tests it's including the windows_uuid.c/h files, which obviously cause errors :)

https://travis-ci.org/CCExtractor/ccextractor/jobs/568105966#L3148

If you could do that (and I check the build on Windows tomorrow), I think this PR will be able to be merged :1st_place_medal:

robdt:
Hey @canihavesomecoffee, sorry about that. Small oversight, fixed now. Please let me know if there is anything else you need from me or that I can do for this.

canihavesomecoffee:
Hi @robdt, if you want to help out, the Windows compilation is failing with this as error:
```
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(40): error C2143: syntax error: missing ')' before '&' [C:\User
s\bot\Documents\ci\ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(40): error C2143: syntax error: missing '{' before '&' [C:\User
s\bot\Documents\ci\ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(40): error C2059: syntax error: '&' [C:\Users\bot\Documents\ci\
ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(40): error C2059: syntax error: ')' [C:\Users\bot\Documents\ci\
ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(45): error C2143: syntax error: missing ')' before '&' [C:\User
s\bot\Documents\ci\ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(45): error C2143: syntax error: missing '{' before '&' [C:\User
s\bot\Documents\ci\ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(45): error C2059: syntax error: '&' [C:\Users\bot\Documents\ci\
ccextractor\windows\ccextractor.vcxproj]
c:\users\bot\documents\ci\ccextractor\src\lib_ccx\win_uuid.h(45): error C2059: syntax error: ')' [C:\Users\bot\Documents\ci\
ccextractor\windows\ccextractor.vcxproj]
```

I tried out some things, but I can't seem to fix it right now :(

robdt:
Hey @canihavesomecoffee I think it is a C++ vs. C problem and your C complier doesn't like the pass by reference when it looks at that header. I am going to rewrite it in a C compliant pointer way. Will check that in in a few mins.

robdt:
Dug closer into what the uuid_t is. Apparently it is an array, so it was already passing a pointer.

Just a thought, if this doesn't work, the MCC spec says that you need to add a UUID in the text there. It doesn't specifically say that it needs to be valid, or unique. ;-)

cfsmp3:
> Dug closer into what the uuid_t is. Apparently it is an array, so it was already passing a pointer.
> 
> Just a thought, if this doesn't work, the MCC spec says that you need to add a UUID in the text there. It doesn't specifically say that it needs to be valid, or unique. ;-)

I see uuid_t being used but not defined anywhere.

Anyway really this should be the same code for all platforms. Why use an internal definition for Windows (that uses Windows specific stuff) and depend on an external thing for linux and Mac? Just have the same code for all platforms. It will make testing much easier for you.

I'd say if you are calling it a uuid then it should be a uuid. At least valid. Unique, since that's hard to generally really well, probably not, but at least use a timestamp so it's different between runs of CCExtractor, even if there's a bigger than desired chance of collision between separately generated uuid's.  rand() is also portable so it should help.

Whatever you do, please don't hardcode a value or anything like that :-)

robdt:
libuuid is part of the linux kernel (util-linux-ng package) and I assume the mac version is a darwin port (I didn't even need a -l to add the library on Mac). Windows seems to have gone their own direction by including that functionality as part of their StdAfx. The APIs looked almost identical, but frustratingly different enough that it wasn't just a simple header swap. Everything seems to be using the uuid_t type (the libuuid stuff on linux/mac for sure, but it also looks like that same type, which is just a collection of bytes, is also used on Windows).

I am hoping that last pass has it fixed. There is no C++ in the header, windows is segregated by WIN32 compiler directive, and the stdafx call looked to be compliant with the spec. Fingers crossed that it compiles now. I unfortunately still don't have access to Windows box to compile and test.

cfsmp3:
> I unfortunately still don't have access to Windows box to compile and test.

Precisely. Remove all that code that is separate between platforms and replace with portable code that you can test in yours. We really don't want to have directives to switch between platforms in anything that is not absolutely necessary. Generating a random string or number definitely doesn't fall in this category.

While we test in Windows, Linux and Mac (this one just from time to time, whenever a Mac user feels like reporting issues and helping one) we aim to be portable and work on FreeBSD or anything else that has a C compiler. Having separate branches make things much harder, and should be unneeded in a program like this.

robdt:
I would have to look deeper at the uuid spec to be certain, but I was under the impression that there were a hand full of different versions of UUID which use the system time, mac address of the host, hashing of an identifier, etc. to build them. But I think there is one version which uses a true random number as its base. Do you want me to pull out use of the library and replace it with an implementation of the uuid from spec? (specifically version 4, that true random one) If so, that would be fine, but it might be a little bit before I have that complete. I got redirected this week and would have to circle back before I could get it coded.

cfsmp3:
Yes, I think that's the way to go.

robdt:
Sounds good! I should be able to circle back and get it implemented in a bit

robdt:
I was able to switch out the uuid lib with a spec compliant implementation. I also got it squished back down so that we didn't have the entire saga of the uuid in the pull request. Please let me know if there is anything more that you need from my end.

canihavesomecoffee:
It's currently not building on the CI linux environment:

`../src/lib_ccx/ccx_encoders_mcc.c: In function ‘generate_mcc_header’:
../src/lib_ccx/ccx_encoders_mcc.c:100:5: error: unknown type name ‘uuid_t’
     uuid_t binuuid;`

thealphadollar:
@canihavesomecoffee @cfsmp3 Should we merge this?

PR #1101, opened 2019-08-15
[IMPROVEMENT] MXF caption frame rates

jgreer:

---

Update the MXF demuxer so that it parses frame rates out of the CDP data.  Without this, ccextractor assumed 29.97 fps in all content, so for example, on a 59.94 fps file the caption timings were double what they should be.


src/lib_ccx/ccx_demuxer_mxf.c
```diff
@@ -307,7 +326,7 @@ static int mxf_read_cdp_data(struct ccx_demuxer *demux, int size, struct demuxer
 	cc_count = buffered_get_byte(demux) & 0x1F;
 	len++;
 	// -4 for cdp footer length
-	if ( (cc_count * 3) != (size - len - 4))
+	if ( (cc_count * 3) > (size - len - 4))
```
jgreer: Some files include extra padding, so this skips the warning when there's extra data, as opposed to missing data.

Issue #1064, opened 2019-01-13
[BUG] rpm build errors - missing freetype sources in tar.gz and configure mismatches between src and rpm


tomas-kuchta:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

CCExtractor version (using the --version parameter preferably) : 0.87

**In raising this issue, I confirm the following (please check boxes, eg [X] - and delete unchecked ones):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**

- [ ] I have never used CCExtractor.
- [x] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [ ] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [x] NO | [] YES - *please specify the last known working version* - I am not sure actually
- What platform did you use? [ ] Windows - [x] Linux - [ ] Mac
- What were the used arguments? `-autoprogram`

**Video links (replace text below with your links) **

None needed

**Additional information**

Running ./rpm.sh from package_creators directory fails due to following issues:
1. incorrect version 0.85 instead of 0.87 in file: ccextractor.spec
2. changed ./configure command in package_creators/tarbal.sh to:
   ./configure --prefix=/usr --enable-ocr
3. missing freetype source files ft2build.h, ... in generated RPMBUILD/SOURCES/ccextractor-0.87.tar.gz
   Adding directory full src/freetype directory to the tar file at creation resolves the build issue.

Build done on: openSuSE 42.3 with following ccextractor specific dependencies:

Thank you, Tomas

[...]

PR #1106, opened 2019-09-10
**[FIX]** Enable RPM creation to work correctly

djotaku:

- [ x] I have checked that another pull request for this purpose does not exist.
- [x ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

- [x ] I absolutely love CCExtractor, but have not contributed previously.

---

{pull request content here}

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,6 +1,7 @@
 0.89 (TBD)
 -----------------
 - Fix: ccx_demuxer_mxf.c: Parse framerate from MXF captions to fix caption timings.
+- Fix: linux/Makefile.am: added some sources to enable rpms to be created.
 
 0.88 (2019-05-21)
 -----------------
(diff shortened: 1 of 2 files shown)


djotaku:
This would fix: https://github.com/CCExtractor/ccextractor/issues/1064

djotaku:
is this PR bot thing right? I changed a file in the linux folder and it says Windows things broke? Or are they broken on the master branch already?

canihavesomecoffee:
Hi Eric,

In this specific case you can indeed ignore these. We still have a couple of cases that always tend to fail :(

About the two broken tests on Linux, I would have to double check, but I suspect that they are also broken (they shouldn't be though) on master.

@cfsmp3 can merge it anyhow if it looks ok for him.

Issue #892, opened 2018-01-09
[BUG] -out=spupng with EIA608/teletext: tags are rendered as-is. 


harrynull:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

CCExtractor version (using the --version parameter preferably) : 0.86

**In raising this issue, I confirm the following (please check boxes, eg [X]):**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [X] I have checked that the issue I'm posting isn't already reported.
- [X] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X]):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [X] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [ ] NO | [X] YES - *please specify the last known working version* 377dc2a48d7dbe7d94d552a7deb430512b3a2c3b
- What platform did you use? All platforms should have this issue.
- What where the used arguments? `-out=spupng`

**Additional information**

![sub0060](https://user-images.githubusercontent.com/7413706/34649418-d03fbd34-f3e9-11e7-82e2-5e7ab617f644.png)
Example of tag rendered as-is.

The issue is introduced in pr #877 , when changing the old implementation to the new FreeType one. The tags are not parsed and rendered as-is.

You can refer to the old implementation for the parsing part (I didn't look into it so I don't know if the old code can apply): 377dc2a48d7dbe7d94d552a7deb430512b3a2c3b

PS: Make sure you set an alert in GitHub so you get notifications about your ticket. We may need to ask questions and we do everything inside GitHub's system.

cfsmp3:
[...]

PR #1132, opened 2019-12-05
[FIX] Added support for font colors and italics

jshin313:

---

The pull request addresses the issue linked [here](https://github.com/CCExtractor/ccextractor/issues/892) as part of a Google Code-in task. Thus, expect less than optimal code from a high school student. The changes were tested on a Windows 10 machine using Visual Studio 2019 to build ccextractor.


src/lib_ccx/params.c
```diff
@@ -2572,6 +2582,12 @@ int parse_parameters (struct ccx_s_options *opt, int argc, char *argv[])
 	opt->enc_cfg.gui_mode_reports = opt->gui_mode_reports;
 	if(opt->enc_cfg.render_font==NULL)
 		opt->enc_cfg.render_font = DEFAULT_FONT_PATH;
+	if (opt->enc_cfg.render_font_italics == NULL)
+		opt->enc_cfg.render_font_italics = DEFAULT_FONT_PATH_ITALICS;
```
cfsmp3: This is always true since there's no play in which opt->enc_cfg.render_font_italics changes. For the standard font we do have the parameter --font, but there isn't an equivalent for --italics (or bold)

src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -21,7 +21,13 @@
 #define CANVAS_WIDTH 600
 
 FT_Library  ft_library = NULL;
+
+// The FT_Face object handles typographical information
+// The different face variables are for the regular, bold, italic, etc. fonts
 FT_Face	 face = NULL;
+FT_Face	 faceb = NULL;
+FT_Face	 facei = NULL;
+FT_Face	 facez = NULL;
```
harrynull: The variable names don't make much sense. Consider renaming them to something like `face_bold_italics`.

src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -608,26 +614,71 @@ void black_background(struct pixel_t *target, int target_w, int x, int y, int w,
...
+		return 0;
+	}
+	if (error = FT_Set_Pixel_Sizes(facei, 0, FONT_SIZE))
+	{
+		mprint("\nFailed to init freetype when trying to set size, error code: %d\n", error);
+		return 0;
+	}
+
+	// Init italics and bold font
+	if (error = FT_New_Face(ft_library, ccx_options.enc_cfg.render_font_italics_bold, 0, &facez))
```
harrynull: Repetitive code. It would be better if you use a function instead.

src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -650,14 +701,75 @@ int spupng_export_string2png(struct spupng_t *sp, uint32_t *str, FILE* output)
...
+
 	// Render characters to image
 	for (uint32_t *iter = str; *iter; ++iter)
 	{
 		uint32_t current_char_code = BigtoLittle32(*iter); // Convert big-endian and little-endian
 
-		if (FT_Load_Char(face, current_char_code, FT_LOAD_RENDER)) continue; // ignore errors
+		// Parse tags like <i> and <b>
+		if (end) {
+			if (current_char_code == '>') {
```
harrynull: It seems to me that if we meet tags like `<b><i>test</i> test2</b>` (if that is valid), it would output "**_test_** test2" instead of "**_test_** **test2**"

src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -650,14 +701,75 @@ int spupng_export_string2png(struct spupng_t *sp, uint32_t *str, FILE* output)
...
+			}
+			continue;
+		}
+
+		if (current_char_code == '<') {
+			bracket = 1;
+			continue;
+		}
+
+		slot = current_face->glyph;
```
harrynull: I think it might be better and easier to decide which typeface to use here based on the value of `italics` and `bold`, instead of changing `current_face` in `if`s.

harrynull:
So I took a quick look at 608 decoder and it seems that `<u>` (underline) might also be a valid tag whereas `<b>` does not seem to be used. Also, it might be possible that the captions contain color information so I'd suggest you to find a sample containing colored captions and see if you could support that too.

src/lib_ccx/ccx_encoders_spupng.c
jshin313: Will do.

src/lib_ccx/ccx_encoders_spupng.c
jshin313: Many truetype fonts use the z suffix to denote a .ttf file for the bold and italic version of a font, especially in Windows (e.g regular = calibri.ttf, italics =calibrii.ttf, bold = calibrib.ttf, bold & italic = calibriz.ttf). I just thought the z suffix was something that people used to denote italics and bold. I'll change the variable names if they're too confusing.

src/lib_ccx/params.c
jshin313: I'll add the flags.

harrynull:
> Should I just implement something for the underline tag and font color then and just get rid of the bold? Are there any other tags that I should be aware of?

Please check the decoders, but as far as I am aware that is all.

jshin313:
I added support for color.

I couldn't find a video sample with had different font colors so I just put the following in the spupng_export_string2png function:
`str = "<font color=\"#66CDAA\"> This should be aquamarine </font> Regular <i> Italics font</i> <font color=\"#00ff00\">This should be green.</font>"; // Test string`

jshin313:
Are you referring to the changes in the functions' parameters and addition changes to the functions' code when you say formatting. Can you clarify what you mean by formatting. Thanks for reviewing my code.

jshin313:
Also I still have to add support for underlining.

PR #1133, opened 2019-12-05
[FIX] Compilation warnings

Fonseverin:

---
I've fixed compilation warnings (through 1% to 27%).
A big part of warnings are caused by (unsigned char *) being converted to other types.
Is there really such a need to have buffers unsigned?

Am I doing everything right or should I do it the other way? And how to resolve implicit type casting warnings with buffers? 

P.s: Also added some comments.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,17 +1,14 @@
-0.90 (2019-12-05)
------------------
-- New: Add a few comments.
-- Fix: Correct if-clauses (missing or misplaced brackets).
-- Fix: Comment out some unused code.
-- Fix: Resolve bunch of compilation's warnings.
-
 0.89 (TBD)
 -----------------
 - New: Added support to output captions to MCC file (#733).
+- New: Added a few comments.
 - Fix: ccx_demuxer_mxf.c: Parse framerate from MXF captions to fix caption timings.
 - Fix: hardsubx_decoder.c: Fix memory leaks using Leptonica API.
 - Fix: linux/Makefile.am: added some sources to enable rpms to be created.
 - Fix: Crash when using -sc (sentence case) option (#1115)
+- Fix: Correct if-clauses (missing or misplaced brackets).
+- Fix: Comment out some unused code.
+- Fix: Resolve bunch of compilation's warnings.
 
 0.88 (2019-05-21)
 -----------------
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -6,6 +6,7 @@
 - Fix: Resolve bunch of compilation's warnings.
 
 0.89 (TBD)
+-----------------
 - New: Added support to output captions to MCC file (#733).
 - Fix: ccx_demuxer_mxf.c: Parse framerate from MXF captions to fix caption timings.
 - Fix: hardsubx_decoder.c: Fix memory leaks using Leptonica API.
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,3 +1,10 @@
+0.90 (2019-12-05)
+-----------------
+- New: Add a few comments.
+- Fix: Correct if-clauses (missing or misplaced brackets).
+- Fix: Comment out some unused code.
+- Fix: Resolve bunch of compilation's warnings.
+
 0.89 (TBD)
 - New: Added support to output captions to MCC file (#733).
 - Fix: ccx_demuxer_mxf.c: Parse framerate from MXF captions to fix caption timings.
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -41,8 +41,8 @@ void print_end_msg(void)
 
 int api_start(struct ccx_s_options api_options)
 {
-    struct lib_ccx_ctx *ctx;
-    struct lib_cc_decode *dec_ctx = NULL;
+    struct lib_ccx_ctx *ctx;              // Context for libs
+    struct lib_cc_decode *dec_ctx = NULL; // Context for decoded captions
     int ret = 0, tmp;
     enum ccx_stream_mode_enum stream_mode;
 
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -503,9 +503,13 @@ int api_param_count(struct ccx_s_options* api_options)
 
 int main(int argc, char* argv[])
 {
-    setlocale(LC_ALL, "");
+    setlocale(LC_ALL, ""); // Supports non-English CCs
+
     struct ccx_s_options* api_options = api_init_options();
-    check_configuration_file(*api_options);
+    check_configuration_file(*api_options); 
+    // If "ccextactor.cnf" is present, takes options from it.
+    // See docs/ccextractor.cnf.sample for more info.
+
 #ifdef PYTHON_API
     for(int i = 1; i < argc; i++)
         api_add_param(api_options,argv[i]);
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -41,10 +41,10 @@ void print_end_msg(void)
 
 int api_start(struct ccx_s_options api_options)
 {
-    struct lib_ccx_ctx *ctx;              // Context for libs
-    struct lib_cc_decode *dec_ctx = NULL; // Context for decoded captions
-    int ret = 0, tmp;
-    enum ccx_stream_mode_enum stream_mode;
+    struct lib_ccx_ctx *ctx       = NULL; // Context for libs
+    struct lib_cc_decode *dec_ctx = NULL; // Context for decoder
+    int ret = 0, tmp = 0;
+    enum ccx_stream_mode_enum stream_mode = CCX_SM_ELEMENTARY_OR_NOT_FOUND;
 
 #if defined(ENABLE_OCR) && defined(_WIN32)
     setMsgSeverity(LEPT_MSG_SEVERITY);
@@ -507,7 +507,7 @@ int main(int argc, char* argv[])
 
     struct ccx_s_options* api_options = api_init_options();
     check_configuration_file(*api_options); 
-    // If "ccextactor.cnf" is present, takes options from it.
+    // If "ccextractor.cnf" is present, takes options from it.
     // See docs/ccextractor.cnf.sample for more info.
 
 #ifdef PYTHON_API
--- a/src/lib_ccx/ccx_common_option.c
+++ b/src/lib_ccx/ccx_common_option.c
@@ -140,7 +140,7 @@ void init_options (struct ccx_s_options *options)
 	options->settings_dtvcc.print_file_reports = 1;
 	options->settings_dtvcc.no_rollup = 0;
 	options->settings_dtvcc.report = NULL;
-	memset(options->settings_dtvcc.services_enabled, 0, CCX_DTVCC_MAX_SERVICES);
+	memset(options->settings_dtvcc.services_enabled, 0, CCX_DTVCC_MAX_SERVICES * sizeof(int));
 
 #ifdef ENABLE_SHARING
 	options->sharing_enabled = 0;
--- a/src/lib_ccx/ccx_common_option.c
+++ b/src/lib_ccx/ccx_common_option.c
@@ -140,7 +140,10 @@ void init_options (struct ccx_s_options *options)
 	options->settings_dtvcc.print_file_reports = 1;
 	options->settings_dtvcc.no_rollup = 0;
 	options->settings_dtvcc.report = NULL;
-	memset(options->settings_dtvcc.services_enabled, 0, CCX_DTVCC_MAX_SERVICES * sizeof(int));
+	memset (
+		options->settings_dtvcc.services_enabled, 0, 
+		CCX_DTVCC_MAX_SERVICES * sizeof(options->settings_dtvcc.services_enabled)
+	);
 
 #ifdef ENABLE_SHARING
 	options->sharing_enabled = 0;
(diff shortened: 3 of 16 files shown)


src/ccextractor.c
```diff
@@ -41,8 +41,8 @@ void print_end_msg(void)
 
 int api_start(struct ccx_s_options api_options)
 {
-    struct lib_ccx_ctx *ctx;
-    struct lib_cc_decode *dec_ctx = NULL;
+    struct lib_ccx_ctx *ctx;              // Context for libs
```
cfsmp3: Probably a good idea to initialize it with NULL.

src/ccextractor.c
```diff
@@ -41,8 +41,8 @@ void print_end_msg(void)
...
+    struct lib_cc_decode *dec_ctx = NULL; // Context for decoded captions
```
cfsmp3: This is the "decoder context".

src/ccextractor.c
```diff
@@ -503,9 +503,13 @@ int api_param_count(struct ccx_s_options* api_options)
 
 int main(int argc, char* argv[])
 {
-    setlocale(LC_ALL, "");
+    setlocale(LC_ALL, ""); // Supports non-English CCs
+
     struct ccx_s_options* api_options = api_init_options();
-    check_configuration_file(*api_options);
+    check_configuration_file(*api_options); 
+    // If "ccextactor.cnf" is present, takes options from it.
```
cfsmp3: typo

src/lib_ccx/ccx_common_option.c
```diff
@@ -140,7 +140,7 @@ void init_options (struct ccx_s_options *options)
 	options->settings_dtvcc.print_file_reports = 1;
 	options->settings_dtvcc.no_rollup = 0;
 	options->settings_dtvcc.report = NULL;
-	memset(options->settings_dtvcc.services_enabled, 0, CCX_DTVCC_MAX_SERVICES);
+	memset(options->settings_dtvcc.services_enabled, 0, CCX_DTVCC_MAX_SERVICES * sizeof(int));
```
cfsmp3: where does this (int) come from? I mean, I know what you mean, but you're adding there some magic I'd say. What if options->settings_dtvcc.services_enabled was something else? Your memset would break.

src/lib_ccx/ccx_common_option.h
```diff
@@ -87,7 +87,7 @@ struct encoder_cfg
 struct ccx_s_options // Options from user parameters
 {
 	int extract;                                               // Extract 1st, 2nd or both fields
-	int no_rollup;
+	int no_rollup;											   // If 1, write one line at a time
```
cfsmp3: This disables roll-up emulation (which causes duplicate output when you open the resulting file with a text editor, but is correct if you actually _play_ it with a media player)

src/lib_ccx/ccx_decoders_isdb.c
```diff
@@ -357,7 +357,7 @@ static struct ISDBText *allocate_text_node(ISDBSubLayout *ls)
 static int reserve_buf(struct ISDBText *text, size_t len)
 {
 	size_t blen;
-	unsigned char *ptr;
+	unsigned char *ptr;
```
cfsmp3: All pointers should be initialized to something (or NULL). Really hate uninitialized variables.

src/lib_ccx/ccx_decoders_isdb.c
```diff
@@ -182,7 +182,7 @@ struct ISDBPos{
 
 struct ISDBText
 {
-	char *buf;
+	unsigned char *buf; // All the buffers in the project are usigned but it. Why?
```
cfsmp3: Probably because it's more convenient in this case (not sure). I see you cast it to char * later...

src/lib_ccx/ccx_decoders_isdb.c
```diff
@@ -1329,9 +1331,11 @@ int isdb_parse_data_group(void *codec_ctx,const uint8_t *buf, struct cc_subtitle
 	ISDBSubContext *ctx = codec_ctx;
 	const uint8_t *buf_pivot = buf;
 	int id = (*buf >> 2);
+#ifdef DEBUG // Fixex unused variables waring
```
cfsmp3: typos there

src/lib_ccx/ccx_decoders_xds.c
```diff
@@ -667,12 +667,13 @@ int xds_do_current_and_future (struct cc_subtitle *sub, struct ccx_decoders_xds_
...
-				int changed = 0;
+				// int changed = 0; /* Currently unused */
 				was_proc = 1;
 				if (ctx->cur_xds_payload_length < 5) // We need 2 data bytes
 					break;
-				if (!ctx->cur_xds_payload[2] & 20 || !ctx->cur_xds_payload[3] & 20) // Bit 6 must be 1
-					break;
+				if (~ctx->cur_xds_payload[2] & 0x20 || ~ctx->cur_xds_payload[3] & 0x20) // Bit 6 must be 1 
+					break; // Should it be inverted? 
+
```
cfsmp3: This was correct as it was. If the bits 6 of those 2 bytes are not 1 then data is invalid so we skip it.

src/lib_ccx/ccx_encoders_smptett.c
```diff
@@ -263,11 +263,18 @@ int write_cc_buffer_as_smptett(struct eia608_screen *data, struct encoder_ctx *c
 					dbg_print(CCX_DMT_DECODER_608, "\r");
 					dbg_print(CCX_DMT_DECODER_608, "%s\n",context->subline);
 				}
+				#pragma GCC diagnostic ignored "-Wunused-variable" /* length is currently unused */
 				int length = get_decoder_line_encoded (context, context->subline, row, data);
 
+				#define char_p (char *)
```
cfsmp3: Uh, #defines in the middle of a C file are ugly - please remove.

src/lib_ccx/ccx_common_option.c
Fonseverin: settings_dtvcc.services_enabled is an array of type int. To fully change its bytes to 0 we need not only size of the array but also size of data stored.
May change to sizeof(settings_dtvcc.services_enabled), though it's too big (as I see)

src/lib_ccx/ccx_decoders_xds.c
Fonseverin: But how do you check it with (int) 20? You use unary operators with int that is in binary 10100. To check for 6th bit, mustn't you use 0x20?

I got about ! operator being right, it just makes little sense (as says compiler) to use operators on bits with it.

src/lib_ccx/ccx_encoders_smptett.c
Fonseverin: I left it there too type less code. EVERY function in there asks for explicit type conversion. Can I change unsigned types to signed here?

src/lib_ccx/ccx_common_option.h
Fonseverin: Thanks! Such things must to be commented! For me (a non-English speaker with a very little knowledge of cc) no-rollup option was a mystery.

src/lib_ccx/ccx_common_option.c
cfsmp3: That's my point - if you are using sizeof you can as well use sizeof (item)*num_items rather than sizeof (type of item you know because you've inspected the code)*num_items.

src/lib_ccx/ccx_encoders_smptett.c
cfsmp3: I don't know. You can, but you're likely to get warnings somewhere else if you do. Give it a try I guess.

src/lib_ccx/ccx_common_option.c
Fonseverin: Got it:)

src/lib_ccx/ccx_encoders_smptett.c
Fonseverin: Yeah. Unsigned to signed (and otherwise) conversions give half of the warnings.

src/ccextractor.c
Fonseverin: Done.

src/ccextractor.c
Fonseverin: Changed to // Context for decoder

Fonseverin:
Can you merge it now? Or do I need to change something else?

docs/CHANGES.TXT
```diff
@@ -1,4 +1,12 @@
+0.90 (2019-12-05)
+-----------------
+- New: Add a few comments.
+- Fix: Correct if-clauses (missing or misplaced brackets).
+- Fix: Comment out some unused code.
+- Fix: Resolve bunch of compilation's warnings.
```
cfsmp3: This stuff should go into 0.89 (since it hasn't be released yet).

src/lib_ccx/ccx_decoders_708.c
```diff
@@ -1198,8 +1198,9 @@ void dtvcc_handle_DLW_DeleteWindows(ccx_dtvcc_ctx *dtvcc,
 {
 	ccx_common_logging.debug_ftn(CCX_DMT_708, "[CEA-708] dtvcc_handle_DLW_DeleteWindows: windows: ");
 
-	int screen_content_changed = 0,
-		window_had_content,current_win_deleted=0;
+	int screen_content_changed = 0, window_had_content;
```
cfsmp3: We should probably init that window_had_content

src/lib_ccx/ccx_decoders_isdb.c
```diff
@@ -1329,9 +1331,11 @@ int isdb_parse_data_group(void *codec_ctx,const uint8_t *buf, struct cc_subtitle
 	ISDBSubContext *ctx = codec_ctx;
 	const uint8_t *buf_pivot = buf;
 	int id = (*buf >> 2);
+#ifdef DEBUG // Fix [-Wunused-variable] when no debug
```
cfsmp3: Where is this DEBUG defined?

src/lib_ccx/ccx_decoders_xds.c
cfsmp3: Ah yes - I didn't see the 0x20. Instead I see that sign instead of the !. Please leave ! because that is correct.
But yes it's 0x20 indeed.

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -179,6 +179,8 @@ unsigned encode_line(struct encoder_ctx *ctx, unsigned char *buffer, unsigned ch
 			bytes += 2;
 			buffer += 2;
 			break;
+		case CCX_ENC_ASCII: // TODO: encode ASCII text
```
cfsmp3: Not sure if this is TODO or just not needed at all.

src/lib_ccx/ccx_encoders_smptett.c
```diff
@@ -263,11 +263,11 @@ int write_cc_buffer_as_smptett(struct eia608_screen *data, struct encoder_ctx *c
 					dbg_print(CCX_DMT_DECODER_608, "\r");
 					dbg_print(CCX_DMT_DECODER_608, "%s\n",context->subline);
 				}
+				#pragma GCC diagnostic ignored "-Wunused-variable" /* length is currently unused */
```
cfsmp3: If we're doing this here we could do it in the #ifdef DEBUG from above

src/zvbi/decoder.c
```diff
@@ -528,6 +528,9 @@ vbi_raw_decoder_resize		(vbi_raw_decoder *	rd,
 				 int *			start,
 				 unsigned int *		count)
 {
+// Currently no actions are taken on variable
+// Ignores warning about it 
+#pragma GCC diagnostic ignored "-Wunused-but-set-variable"
```
cfsmp3: These #pragmas should be used to disable specific errors in known places, but once you pass that point warnings needs to be reenabled. Otherwise we might miss them in places we don't know.

src/zvbi/sampling_par.c
```diff
@@ -448,10 +447,11 @@ _vbi_sampling_par_from_services_log
 				videostd_set |= par->videostd_set;
 		}
 
-		if (VBI_VIDEOSTD_SET_525_60 & videostd_set)
-			margin = 1.0e-6;
-		else
-			margin = 2.0e-6;
+		// double margin; /* Currently unused */
+		// if (VBI_VIDEOSTD_SET_525_60 & videostd_set)
```
cfsmp3: Why is this commented out?
zvbi is not our library. We need to leave it alone or we'll make our life more difficult when upgrading it.

src/lib_ccx/ccx_decoders_isdb.c
Fonseverin: I don't know:) It's not my definition. It was applied to log function earlier

src/lib_ccx/ccx_encoders_helpers.c
Fonseverin: Then why is there CCX_ENC_ASCII in enum?

src/zvbi/decoder.c
Fonseverin: Yes, sorry. I've forgotten to use #pragma GCC diagnostic push and pop

src/zvbi/sampling_par.c
Fonseverin: It just creates warnings :0 I'll try to resolve it the other way

src/lib_ccx/ccx_encoders_smptett.c
Fonseverin: The difference is that these variables have no use even if something is defined.
By using #ifdef we show that these are necessary to debug.

src/zvbi/sampling_par.c
Fonseverin: There is a way to suppress all the warnings from external libs in cmake file. Should I use it?

PR #1134, opened 2019-12-05
[FIX] Make header respect `-lf` for the webvtt encoder

NilsIrl:
- [X] I have never used CCExtractor.

---

See title

### DIFF
--- a/src/lib_ccx/ccx_encoders_webvtt.c
+++ b/src/lib_ccx/ccx_encoders_webvtt.c
@@ -214,18 +214,11 @@ int write_webvtt_header(struct encoder_ctx *context)
 		int used;
 		unsigned h1, m1, s1, ms1;
 		millis_to_time(context->timing->sync_pts2fts_fts, &h1, &m1, &s1, &ms1);
-		if (ccx_options.enc_cfg.line_terminator_lf)
-		{
-			sprintf(header_string,
-				"X-TIMESTAMP-MAP=MPEGTS:%ld,LOCAL:%02u:%02u:%02u.%03u\n\n",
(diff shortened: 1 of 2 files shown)


src/lib_ccx/ccx_encoders_webvtt.c
```diff
@@ -214,9 +214,18 @@ int write_webvtt_header(struct encoder_ctx *context)
...
+			sprintf(header_string,
+				"X-TIMESTAMP-MAP=MPEGTS:%ld,LOCAL:%02u:%02u:%02u.%03u\n\n",
+				context->timing->sync_pts2fts_pts, h1, m1, s1, ms1);
+		}
+		else
+		{
+			sprintf(header_string,
+				"X-TIMESTAMP-MAP=MPEGTS:%ld,LOCAL:%02u:%02u:%02u.%03u\r\n\r\n",
+				context->timing->sync_pts2fts_pts, h1, m1, s1, ms1);
+		}
```
cfsmp3: Instead of adding a new block it's would be better to use the ternary operator here to pass either \n or \r\n to sprintf.

NilsIrl: I added a commit to fix this though I just realised I did it in another way.

The way I did it is by passing a different string literal. Would you prefer to use `sprintf` to select line ending?

cfsmp3: Yes, just change the line ending as a parameter, otherwise you still have to maintain two strings. Something like this would be better:

```
		sprintf(header_string, "....%s", ... , ccx_options.enc_cfg.line_terminator_lf ? "\n\n" : "\r\n\r\n");
```

PR #1137, opened 2019-12-06
[FIX] Add the `X-TIMESTAMP-MAP` header at all time to webvtt files. WIP

### LINKED ISSUE #1127: [Proposal] Adding an option to enable/disable X-TIMESTAMP-MAP in WebVTT

CCExtractor version (using the --version parameter preferably) : **X.X**

**Necessary information**
- Is this a regression (did it work before)? [x] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [ ] Windows - [ ] Linux - [x] Mac
- What were the used arguments? `-out=webvtt`

**Additional information**
Currently when WebVTT is specified as output, X-TIMESTAMP-MAP is automatically added.  And if no subtitle text is found in the input, it generates an empty WebVTT, with just one line "WEBVTT" in the file.

There are potentially two issues with this implementation.
(a) X-TIMESTAMP-MAP is not a WebVTT requirement, but Apple HLS requirement, this may cause a problem if it is used in non-HLS environment.  It should be an option to turn it off.
(b) On the other hand, Apple HLS spec says "an X-TIMESTAMP-MAP metadata header SHOULD be added to each WebVTT header", which includes the empty ones.  The spec also says "If a WebVTT segment does not have the X-TIMESTAMP-MAP, the client MUST assume that the WebVTT cue time of 0 maps to an MPEG-2 timestamp of 0.".  But unfortunately, not all clients follow this rule.  It is safer to add X-TIMESTAMP-MAP in all VTT files, if it is enabled.

So, my proposal is to add an option for WebVTT, whether X-TIMESTAMP-MAP is enabled or disabled.  And if it is enabled, it will be added in every VTT file it generates, including one with no caption cues.

(cf. "HTTP Live Streaming": https://tools.ietf.org/html/rfc8216 section 3.5)

NilsIrl:
Fix #1127

Add the `X-TIMESTAMP-MAP` header at all time to webvtt files.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -425,7 +425,7 @@ static int write_bom(struct encoder_ctx *ctx, struct ccx_s_write *out)
 }
 
 // Returns -1 on error and 0 on success
-static int write_subtitle_file_header(struct encoder_ctx *ctx, struct ccx_s_write *out)
+static int write_subtitle_file_header(struct encoder_ctx *ctx, struct ccx_s_write *out, struct ccx_common_timing_ctx *timing)
 {
 	int used;
 	int header_size = 0;
@@ -485,6 +485,7 @@ static int write_subtitle_file_header(struct encoder_ctx *ctx, struct ccx_s_writ
 					return -1;
 				}
 			}
+			write_webvtt_header(ctx, timing);
 			break;
 		case CCX_OF_SAMI: // This header brought to you by McPoodle's CCASDI
 			//fprintf_encoded (wb->fh, sami_header);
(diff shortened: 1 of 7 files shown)


canihavesomecoffee:
@NilsIrl There are conflicts on this PR. Can you resolve them so that we could proceed merging?

canihavesomecoffee:
This is not mergeable.

@NilsIrl Can you take a look at some of the DVB ones? They're crashing with the current code.

E.g.: https://sampleplatform.ccextractor.org/regression/test/17/view

NilsIrl:
I honestly don't think this is a good fix. It's just a disgusting hack.

And tbh with the way it's currently working I don't think there is any good fix. If there is no subtitle we should just not produce a file. Also this was discussed in the issue linked to this but I don't understand how we can sync subtitles and images when we don't have subtitles.

cfsmp3:
@NilsIrl is this ready for merging?

NilsIrl:
I'd prefer to test it first. And also see if it can break stuff. Because it moves stuff around which I'm not a fan of (risk of breakage).

cfsmp3:
@NilsIrl can't be merged

cfsmp3:
@NilsIrl status?

NilsIrl:
> @NilsIrl status?

I found a solution that should be a lot better and changes only a few lines. I'll try to do it this weekend.

It would call `write_webvtt_header` inside of `general_loop` instead of trying to get the timings to be available.

NilsIrl:
Continuing work here: https://github.com/NilsIrl/ccextractor/tree/webvtt_header_fix

PR #1139, opened 2019-12-08
[IMPROVEMENT] Filter bad words

NilsIrl:

This PR reformats most of the capitalization parts as well to make a better interface.

This PR also removes one of the list that was used for capitalization before `spell_lower` as it was useless.

PS: also btw, for the list of words, I took them from Wikipedia, I don't have that good of a vocab 😂


src/lib_ccx/ccx_common_common.c
```diff
@@ -57,14 +57,13 @@ void millis_to_time(LLONG milli, unsigned *hours, unsigned *minutes,
...
-void freep(void *arg)
+void freep(void **arg)
 {
-	void **ptr = (void **)arg;
-	if (*ptr)
-		free(*ptr);
-	*ptr = NULL;
-
+	if (*arg)
+		free(*arg);
```
cfsmp3: Several things on this change:
You're changing freep(), but I don't see you changing any of its callers to now pass a pointer to pointer.
You're not checking if arg itself is NULL, so *arg would coredump by doing this

void *var = NULL
freep (var); // which would happen unless you change the caller
or
freep(&var); // Which would be the correct way now since you changed the signature of freep()

Also, free(NULL) is OK (as in free() itself just returns if passed NULL). So no need to check if (*arg)

src/lib_ccx/ccx_decoders_structs.h
```diff
@@ -62,7 +62,7 @@ typedef struct eia608_screen // A CC buffer
 {
 	/** format of data inside this structure */
 	enum ccx_eia608_format format;
-	unsigned char characters[15][33];
+	char characters[15][33];
```
cfsmp3: This is a _major_ change in a structure that has been stable for years. Why do you want to do this? There's a reason for unsigned.

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -357,84 +428,74 @@ used=encode_line (ctx->buffer,(unsigned char *) string);
...
-	strcpy(new_correct, word);
-	for (size_t i = 0; i<len; i++)
+
+	strcpy(list->words[list->len++], word);
+	return word_len;
+}
+
+int add_builtin_capitalized_words()
+{
+	static int function_already_ran = 0; // so we don't do it twice
```
cfsmp3: You just made the function multithread unsafe :-)
Do you really need this? Seems like not calling it twice is enough...

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -357,84 +428,74 @@ used=encode_line (ctx->buffer,(unsigned char *) string);
...
 				return -1;
-			i++;
 		}
-		spell_builtin_added = 1;
+		function_already_ran = 1;
 	}
 	return 0;
 }
 
+void fix_subtitles(struct encoder_ctx *context, int line_number, struct eia608_screen *data)
```
cfsmp3: "fix" is too vague, come up with a better verb :-) 
Maybe correct_spelling_and_censor_words_608 ()

src/lib_ccx/ccx_encoders_helpers.h
```diff
@@ -33,8 +31,11 @@ void get_sentence_borders(int *first, int *last, int line_num, struct eia608_scr
 
 int string_cmp(const void *p1, const void *p2);
 int string_cmp_function(const void *p1, const void *p2, void *arg);
-int add_built_in_words(void);
-int add_word(const char *word);
+
+int add_builtin_profane_words();
+int add_builtin_capitalized_words();
```
cfsmp3: If there's no parameters the signature should be (void)

src/lib_ccx/params.c
```diff
@@ -34,47 +34,58 @@
 
 static int inputfile_capacity=0;
 
-int process_cap_file (char *filename)
+#define MAX_WORD_LENGTH 50
```
cfsmp3: Keep in mind that for now at least we're only doing this for 608 subtitles (as far as I can see), and the width of a line in 608 is 32...

NilsIrl: Ok, I thought I was set arbitrarily.

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -31,16 +44,68 @@ static const char *spell_builtin[] =
 	NULL
 };
 
+static const char *profane_builtin[] =
+{
+	"arse",
+	"ass",
```
aadibajpai: what if someone means the donkey? 🤔

NilsIrl: I don't know I just copy and pasted the words from https://en.wikipedia.org/wiki/Category:Profanity and removed the ones that could totally be used, in a non profane way.

Also I didn't know they meant donkey

aadibajpai:
Just generally curious but would it be possible to match patterns? Like `fuck*` would match not only fuck but also fucking or fucker and so on. I found an old google archive with such a wordlist https://code.google.com/archive/p/badwordslist/downloads

NilsIrl:
> I found an old google archive with such a wordlist https://code.google.com/archive/p/badwordslist/downloads

I also found a list from a university that was around 10k words. But most of the words were not profanity in themselves so I gave up, and just took some words to use as placeholder.

e.g. of irrelevant word: gay, transgender, mum...

src/lib_ccx/ccx_common_common.c
NilsIrl: The callers are already passing pointers to pointers so it should be fine. I just changed the type to reflect that.

> You're not checking if arg itself is NULL, so *arg would coredump by doing this

Fixed, but was already the case previously

NilsIrl:
Acted on feedback

src/lib_ccx/ccx_encoders_helpers.c
NilsIrl: I just translated what was previously there.

NilsIrl: The problem is that someone could pass the `--kf` or `-sc` argument twice.

How could I make it multithread safe?

src/lib_ccx/ccx_common_common.c
```diff
@@ -57,14 +57,12 @@ void millis_to_time(LLONG milli, unsigned *hours, unsigned *minutes,
...
```
cfsmp3: It should be
if (arg)
free (*argc)

src/lib_ccx/ccx_encoders_helpers.c
cfsmp3: That's fine, we can leave this list :-) If someone wants a different list they can just edit it.

src/lib_ccx/ccx_encoders_helpers.c
cfsmp3: In general static variables in a function make it MT unsafe until there's a mutex involved (and we definitely don't want that extra complexity). Anyway I think I'm OK-ish with that here.

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -468,6 +540,6 @@ void shell_sort(void *base, int nb, size_t size, int(*compar)(const void*p1, con
 
 void ccx_encoders_helpers_perform_shellsort_words(void)
 {
-	shell_sort(spell_lower, spell_words, sizeof(*spell_lower), string_cmp_function, NULL);
-	shell_sort(spell_correct, spell_words, sizeof(*spell_correct), string_cmp_function, NULL);
+	shell_sort(spell_lower.words,   spell_lower.len,   sizeof(*spell_lower.words),   string_cmp_function, NULL);
```
cfsmp3: Seems strange to be passing  a length there and also using sizeof, why?

src/lib_ccx/params.c
```diff
@@ -34,47 +34,58 @@
...
 {
 	int ret = 0;
-	FILE *fi = fopen (filename,"rt");
-	if (fi == NULL)
+	FILE *fi;
+
+	if ((fi = fopen(filename, "r")) == NULL)
 	{
-		mprint ("\rUnable to open capitalization file: %s\n", filename);
+		mprint("Unable to open word list file: %s\n", filename);
```
cfsmp3: You're missing \r there, which is there so the message appears at the beginning of the line (instead of to the right of the progress info)

src/lib_ccx/ccx_common_common.c
cfsmp3: And remember that this affects the callers which now how to pass a pointer to the pointer they want to free

src/lib_ccx/ccx_encoders_helpers.c
NilsIrl: I didn't write that code but:
> Seems strange to be passing a length

So that it the function knows how many elements it needs

> using sizeof

To know the size of each element

The best would probably to use a stdlib sorting function.

Also I didn't sort the list of profane words, which means binary search won't work (so far it has worked because the list was already sorted).

I'll be fixing that tomorrow

src/lib_ccx/ccx_encoders_helpers.c
NilsIrl: OK, it was already like that before.

src/lib_ccx/params.c
NilsIrl: What's the difference between \r and \n?

Should a '\n' be added instead?

cfsmp3: No. \r moves the cursor to the beginning of the line and \n starts a new line. So if you want to be in the same line would start writing from the left (overwriting whatever is already there) you send a \r.

src/lib_ccx/ccx_encoders_helpers.c
NilsIrl: Actually I wasn't being careful, these functions cannot be called twice as they are dealt with after the parsing of arguments.

src/lib_ccx/ccx_common_common.c
NilsIrl: This had already been fixed when you wrote your comment, probably pushed while you were reviewing

src/lib_ccx/params.c
NilsIrl: I had no clue this was possible. I will probably use this a lot now.

src/lib_ccx/ccx_encoders_helpers.c
NilsIrl: The use of shell sort was added in #41 because it was faster than quicksort.

Also to clarify, `spell_words` was the length of the array (now `spell_correct.len`) so no behaviour should have changed.

cfsmp3: That's not too reassuring :-) You need to test a lot and be sure that you are not introducing any bug...

src/lib_ccx/params.c
aadibajpai: That is basically what CRLF is, Carriage Return (\r) and Line Feed (\n)

cfsmp3:
@NilsIrl once you have tested it write here how you tested specifically, the output, etc, and we'll merge.

NilsIrl:
It would probably be better to modify the subtitles before they are passed to the encoder. This abstracts these features away from encoders.

What do you think @cfsmp3 

For example, right now I'm implementing the scc encoder and have to deal with this which is even worse because I need to implement the "previous" version, the version before this PR.

cfsmp3:
Not so easy - for example for 608 you'd need to modify the grid, which is
fixed in size - it would be a pain in the ass.

It's not a bad idea though, but I don't think it would save as much time as
you'd think :-)

NilsIrl:
Multi word swear words will not work

src/lib_ccx/ccx_encoders_helpers.c
```diff
@@ -31,16 +44,68 @@ static const char *spell_builtin[] =
...
+	"fuck",
+	"goddamn",
+	"godsdamn",
+	"hell",
+	"holy shit",
+	"horseshit",
+	"Jesus fuck",
+	"Jesus Harold Christ",
+	"Jesus wept",
+	"Judas Priest",
```
canihavesomecoffee: I'm not sure this should be considered as a swear word :p

If there's ever a documentary on https://en.wikipedia.org/wiki/Judas_Priest, this will be annoying ;)

canihavesomecoffee:
@NilsIrl The existing capfile (the dictionary) no longer works:

E:\GitHub\ccextractor\windows\Debug>ccextractorwin.exe --capfile ..\..\Dictionary\MattS_dictionary.txt E:\Downloads\c83f765c661595e1bfa4750756a54c006c6f2c697a436bc0726986f71f0706cd.ts
Error: There was an error processing the capitalization file.

Please fix this before we can merge.

cfsmp3:
@canihavesomecoffee all OK now?

canihavesomecoffee:
> 
> 
> @canihavesomecoffee all OK now?

Didn't retest yet. Can do that tomorrow.

canihavesomecoffee:
Dictionary still works as intended, built-in list also works.

Will trigger a final re-run of the Test Suite to check.

canihavesomecoffee:
I'd say it's mergeable.

Issue #780, opened 2017-09-28
Compilation warnings on C++ 11


cfsmp3:
../src/ccextractor.cpp:813:12: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
   mprint ( "Teletext decoder: %"PRIu32" packets processed, %"PRIu32" SRT frames written.\n", tlt_packet_counter, tlt_frames_produced);
            ^
../src/ccextractor.cpp:813:39: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
   mprint ( "Teletext decoder: %"PRIu32" packets processed, %"PRIu32" SRT frames written.\n", tlt_packet_counter, tlt_frames_produced);
                                       ^
../src/telxcc.cpp:599:42: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
    if (wbout1.fh!=-1) fdprintf(wbout1.fh,"%"PRIu32"\r\n%s --> %s\r\n", tlt_frames_produced, timecode_show, timecode_hide);
                                          ^

mackworth:
Oh, man. If you want compiler warnings, there's always been a gazilllion on the mac. (where gazillion is defined as "about 300" currently)

I particularly like the "tautological" ones:
> comparison of constant 256 with expression of type 'unsigned char' is always true [-Wtautological-constant-out-of-range-compare]
>                                         if (c < 256)

cfsmp3:
@mackworth where are those warnings? If they are in GPAC or any other library we just borrowed probably not worth it to do anything about it (except update the library), but if they are in our code we should definitely fix them.

mackworth:
So, the errors are generated from api/extractors/extractor.c,  src/lib_ccx  src/zlib, and src/zvbi
[ccextractor Errors.zip](https://github.com/CCExtractor/ccextractor/files/1384100/ccextractor.Errors.zip)

shucon:
@csfmp3 all the warnings are not eliminated yet , reopen the issue please.

saurabhshri:
@cfsmp3 Please re-open the issue.

Note to the people working on this issue : ff you are building on Linux using the build scripts located inside the `/linux/` directory, please use `./builddebug` for complete list of warnings being generated.

mackworth:
Just as an update, here's the 3,400 lines of compilation warnings on the current master branch when built on the Mac...  
[...]

PR #1160, opened 2019-12-28
[FIX] Get rid of a few compilation warnings

jshin313:

---

The following warnings were displayed after running make on a Debian 9 linux machine using gcc 8.3.0:

```
../src/lib_ccx/params.c: In function ‘add_file_sequence’:
../src/lib_ccx/params.c:184:3: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
   strncpy(filename+m+(strlen(num)-strlen(temp)),temp,strlen(temp));
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/lib_ccx/params.c:182:7: note: length computed here
   if (strlen(temp) > strlen(num)) // From 999 to 1000, etc. 
```
```
 In function ‘spupng_export_string2png.part.3’,
    inlined from ‘spupng_export_string2png’ at ../src/lib_ccx/ccx_encoders_spupng.c:692:5:
 ../src/lib_ccx/ccx_encoders_spupng.c:733:2: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]
  strncpy(tmp, str, strlen(str)+1);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../src/lib_ccx/ccx_encoders_spupng.c: In function ‘spupng_export_string2png’:
../src/lib_ccx/ccx_encoders_spupng.c:733:20: note: length computed here
  strncpy(tmp, str, strlen(str)+1);
```

The above warnings were fixed by changing strncpy to memcpy.  

The next warning was fixed by declaring a struct before a function declaration;
```
 ../src/lib_ccx/ocr.c:1028:113: warning: ‘struct image_copy’ declared inside parameter list will not be visible outside of this definition or declaration
 * ocr_bitmap(png_color *palette,png_byte *alpha, unsigned char* indata,unsigned char d,int w, int h, struct image_copy *copy)

```
#780


src/lib_ccx/ccx_encoders_spupng.c
```diff
@@ -727,10 +727,12 @@ int spupng_export_string2png(struct spupng_t *sp, char *str, FILE* output)
...
 
 	//str = "<font color=\"#66CDAA\"> __ This should be aquamarine </font> Regular <i> Italics font</i> <font color=\"#00ff00\">This should be green.</font> <u> Underlining __ testing </u> Regular text. Even more text. Random text. More text."; // Test string
-	char *tmp = (char*)malloc((strlen(str) + 1) * sizeof(char));
+	int len = strlen(str);
+	char *tmp = (char*)malloc((len + 1) * sizeof(char));
 	if (!tmp)
 		return -1;
-	strncpy(tmp, str, strlen(str)+1);
+	memcpy(tmp, str, len);
+	tmp[len] = '\0';
```
cfsmp3: What warnings is this fixing? strncpy() seems OK to me...
but if you are using memcpy then memcpy (tmp, str, len+1) would be better - you know that the source string does have the final 0 (otherwise strlen() would crash) so there's no need to add it manually.

src/lib_ccx/ocr.c
```diff
@@ -1025,6 +1026,9 @@ char *paraof_ocrtext(struct cc_subtitle *sub, const char *crlf, unsigned crlf_le
 	return str;
 }
 #else
+
+struct image_copy;
```
cfsmp3: What is this for?

src/lib_ccx/params.c
```diff
@@ -179,10 +179,13 @@ int add_file_sequence (struct ccx_s_options *opt, char *filename)
...
 		sprintf(temp, "%d",i);
-		if (strlen(temp) > strlen(num)) // From 999 to 1000, etc.
+
+		int temp_len = strlen(temp);
+		int num_len = strlen(num);
+		if (temp_len > num_len) // From 999 to 1000, etc.
 			break;
-		strncpy(filename+m+(strlen(num)-strlen(temp)),temp,strlen(temp));
-		memset(filename+m,'0',strlen(num)-strlen(temp));
+		memcpy(filename+m+(num_len-temp_len),temp,temp_len);
```
cfsmp3: Same question, what are the warnings here?

src/lib_ccx/ccx_encoders_spupng.c
jshin313: The warning is as follows: warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]

jshin313: The compiler complains because the 3rd parameter depends on the size of the source, instead of the destination.

src/lib_ccx/ocr.c
jshin313: I got this after running make: `warning: ‘struct image_copy’ declared inside parameter list will not be visible outside of this definition or declaration`

src/lib_ccx/ccx_encoders_spupng.c
cfsmp3: Consider just using strdup() instead.

jshin313: OK.

src/lib_ccx/params.c
jshin313: The warning was similar to the first:` warning: ‘strncpy’ specified bound depends on the length of the source argument [-Wstringop-overflow=]`

Issue #1116, opened 2019-10-29
Windows build of 0.88 never works


ahmedtds:
It just crashes every time, you probably never tested it on a 32bit system so that's why nobody still noticed it, I'm waiting for months maybe you will notice and fix it but nope

Please fix it, thanks

This is the build I'm talking about: https://github.com/CCExtractor/ccextractor/releases/download/v0.88/ccextractor.0.88-windows.binaries.zip

0.85 works without an issue

OS: Windows 7 32bit

MatejMecka:
Hey there! Can you leave us any stacktrace or a way how to reproduce it? We know the environment you are running but can we achieve the crash?

ahmedtds:
> 
> 
> Hey there! Can you leave us any stacktrace or a way how to reproduce it? We know the environment you are running but can we achieve the crash?

Hi, just try to use it on 32bit Windows 7 you'll see it will always crash. Open the GUI, drag and drop a mp4 file and Press start and it will crash. Like I said it just never works since 0.88 came out.

vikramcs:
configuration.yaml entries and (fill out even if it seems unimportant):

homekit:
port: 51827
ip_address: 192.168.1.116
filter:
include_domains:
- alarm_control_panel
- light
- switch
- fan
- climate
- media_player
- lock
- sensor

Traceback (if applicable):


Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/homeassistant/setup.py", line 151, in _async_setup_component
    hass, processed_config)
  File "/usr/local/lib/python3.7/site-packages/homeassistant/components/homekit/__init__.py", line 84, in async_setup
    await hass.async_add_executor_job(homekit.setup)
  File "/usr/local/lib/python3.7/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/usr/local/lib/python3.7/site-packages/homeassistant/components/homekit/__init__.py", line 226, in setup
    port=self._port, persist_file=path)
  File "/usr/local/lib/python3.7/site-packages/homeassistant/components/homekit/accessories.py", line 194, in __init__
    super().__init__(**kwargs)
  File "/usr/local/lib/python3.7/site-packages/pyhap/accessory_driver.py", line 195, in __init__
    self.http_server = HAPServer(network_tuple, self)
  File "/usr/local/lib/python3.7/site-packages/pyhap/hap_server.py", line 801, in __init__
[...]

PR #1165, opened 2019-12-30
[FIX] Change inet_ntop to inet_ntoa for Windows XP compatibility

jshin313:

---
#1116
The change allows the command-line version of ccextractor to be built for Windows XP and run on it as well.  The following occurs without the changes because inet_ntop() is not supported on Windows XP.
image

CCX was built on a Windows 10 machine using Visual Studio 2017 targeted for Windows XP and then run on a Windows XP VM. Visual Studio 2019 doesn't support targeting for Windows XP.

### DIFF
--- a/src/lib_ccx/networking.c
+++ b/src/lib_ccx/networking.c
@@ -335,7 +335,7 @@ int net_udp_read(int socket, void *buffer, size_t length, const char *src_str, c
 	assert(length > 0);
 
 	int i;
-	char ip[50] = "";
+	char ip[15];
 	struct sockaddr_in source_addr;
 	socklen_t len = sizeof(source_addr);
 	/* Get address of host to check for udp network mutlicasting */
@@ -354,7 +354,8 @@ int net_udp_read(int socket, void *buffer, size_t length, const char *src_str, c
 	{
 		do {
 			i = recvfrom(socket, (char *) buffer, length, 0, (struct sockaddr*)&source_addr, &len); /* peek at the data*/
-			inet_ntop(AF_INET, &(source_addr.sin_addr), ip, 50);
+			memset(ip, 0, sizeof(char) * 15);
+			memcpy(ip, inet_ntoa(source_addr.sin_addr), sizeof(ip));
 		} while (strcmp(ip, src_str)!=0);												/* Loop till we find intended source */
 	}
 	else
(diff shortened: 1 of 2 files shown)


src/lib_ccx/networking.c
```diff
@@ -335,7 +335,7 @@ int net_udp_read(int socket, void *buffer, size_t length, const char *src_str, c
 	assert(length > 0);
 
 	int i;
-	char ip[50] = "";
+	char ip[15];
```
cfsmp3: I'd say we are one short here (3*4 for digits + 3 for dots + final zero)
Also it would be good to have that as a constant rather than appear twice as a magic number

jshin313: Should I fix the other arrays as well? https://github.com/jshin313/ccextractor/blob/594a83cc4ec76b61ebe2f81dbd6af5c090f0f57b/src/lib_ccx/networking.c#L1096

canihavesomecoffee: If you want to create a PR for this today, yes, please :)

jshin313: Should I use the INET_ADDRSTRLEN constant? The <netinet/in.h> POSIX®header file says the constant is 16 bytes. But the Windows version is 22 bytes
`// Maximum length of address literals (potentially including a port number)
// generated by any address-to-string conversion routine.  This length can
// be used when declaring buffers used with getnameinfo, WSAAddressToString,
// inet_ntoa, etc.  We just provide one define, rather than one per api,
// to avoid confusion.
//
// The totals are derived from the following data:
//  15: IPv4 address
//  45: IPv6 address including embedded IPv4 address
//  11: Scope Id
//   2: Brackets around IPv6 address when port is present
//   6: Port (including colon)
//   1: Terminating null byte
//
#define INET_ADDRSTRLEN  22`

canihavesomecoffee: Seems like a good idea to me to use a built-in constant :)

cfsmp3: If there is a portable built-in constant it's definitely best to use it.
If there isn't one define your own, but really there's nothing wrong with being generous - a few bytes wasted is much much better than an unexpected buffer overrun. Just add a comment next to the definition so the reader doesn't have to figure out where a magic number comes from.

Issue #1166, opened 2020-01-01
[BUG] X-TIMESTAMP-MAP doesn't show up with DVB subs


NilsIrl:
CCExtractor version: 0.88

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [X] I have checked that the issue I'm posting isn't already reported.
- [X] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**

- [X] I have used CCExtractor just a couple of times.

**Necessary information**
- Is this a regression (did it work before)? idk (probably not)
- What platform did you use? [X] Linux

arguments:

```
./ccextractor -datapid 0xc37 -out=webvtt files/test1_dvbsub.ts
```

Video link: http://akini.mbnet.fi/videos/test1_dvbsub.ts

Related to #1127 in goal though they have different "symptoms" and arise in different cases.

NilsIrl:
Info for some who wants to fix it:

This is the line that calls the function that adds the header (as well as the subtitles) when the input file isn't dvb:

https://github.com/CCExtractor/ccextractor/blob/3a1815163fed76683d7c92b0e7aaf32772e9e074/src/lib_ccx/ccx_encoders_common.c#L1160

This is the line that calls the function that doesn't do that:

https://github.com/CCExtractor/ccextractor/blob/3a1815163fed76683d7c92b0e7aaf32772e9e074/src/lib_ccx/ccx_encoders_common.c#L1220

So you basically want to second line to the the same thing as the first line. Feel free to ask more information if my instructions are unclear.

NilsIrl:
This is the line in `write_cc_buffer_as_webvtt` that adds the header:

https://github.com/CCExtractor/ccextractor/blob/3a1815163fed76683d7c92b0e7aaf32772e9e074/src/lib_ccx/ccx_encoders_webvtt.c#L281


[...]

PR #1167, opened 2020-01-01
[Fix] Make -delay all output formats

### LINKED ISSUE #1103: Param "-delay" does not work in out=webvtt but works in out=srt output
(ccextractor version 0.88 from github)

Using `out=webvtt` does not add 4sec delay to text timings but `out=srt` from the same input oboyes a delay parameter fine.
```
## fail, does not add +4sec
ccextractor -codec -dvbsub -utf8 -nobom -nofc -out=webvtt \
  -in=ts "test.ts" -o "sub.vtt" \
  -delay 4000 -noteletext -nocodec teletext -datapid 0x467 -parsePAT -parsePMT
## ok, adds +4sec
ccextractor -codec -dvbsub -utf8 -nobom -nofc -out=srt \
  -in=ts "test.ts" -o "sub.srt" \
  -delay 4000 -noteletext -nocodec teletext -datapid 0x467 -parsePAT -parsePMT
```
Is it a problem here in webvtt.c file, see srt vs webvtt variable handling `ms_start,ms_end,subs_delay` values?
https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_webvtt.c#L399
https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_srt.c#L84

NilsIrl:
Fix #1103 

Move up the pipeline, when the calculation of delay is done.

### DIFF
--- a/src/lib_ccx/ccx_decoders_common.c
+++ b/src/lib_ccx/ccx_decoders_common.c
@@ -15,7 +15,7 @@ made to reuse, not duplicate, as many functions as possible */
 #include "ccx_encoders_mcc.h"
 #include "ccx_dtvcc.h"
 
-uint64_t utc_refvalue = UINT64_MAX;  /* _UI64_MAX means don't use UNIX, 0 = use current system time as reference, +1 use a specific reference */
+uint64_t utc_refvalue = UINT64_MAX;  /* _UI64_MAX/UINT64_MAX means don't use UNIX, 0 = use current system time as reference, +1 use a specific reference */
 extern int in_xds_mode;
 
 
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -1113,9 +1113,11 @@ int encode_sub(struct encoder_ctx *context, struct cc_subtitle *sub)
 				// Determine context based on channel. This replaces the code that was above, as this was incomplete (for cases where -12 was used for example)
 				out = get_output_ctx(context, data->my_field);
 
+				data->end_time += context->subs_delay;
+				data->start_time += context->subs_delay;
+
 				if (data->format == SFORMAT_XDS)
 				{
-					data->end_time = data->end_time + context->subs_delay;
 					xds_write_transcript_line_prefix(context, out, data->start_time, data->end_time, data->cur_xds_packet_class);
 					if (data->xds_len > 0)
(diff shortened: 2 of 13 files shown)


cfsmp3:
How have you tested this? @NilsIrl 
You seem to have good ideas and know what you are doing but lack in the testing side of things :-) That's also really important, our CI system is still not perfect and the rest of us also like to be coding rather than testing...

NilsIrl:
@cfsmp3 I've tried using with and without delay with srt, webvtt, and snupng using the file and arguments of #1166 (except -out=webvtt which changes).

Also please note that the end_time was already calculated earlier in some cases and that there were more than one encoder which didn't take into account subs_delay so this fixed more than webvtt.

NilsIrl:
I also tested with `FX.ts` with ssa, srt and webvtt

cfsmp3:
@canihavesomecoffee OK to merge? I can't test myself for the next 48 hours

canihavesomecoffee:
> 
> 
> @canihavesomecoffee OK to merge? I can't test myself for the next 48 hours

Will check tomorrow

cfsmp3:
@NilsIrl Can you solve the conflicts?

NilsIrl:
Conflict fixed

canihavesomecoffee:
Linux tests look OK to me, Windows is still running.

cfsmp3:
Windows fails with error code 10 on some samples, which means NO_CAPTIONS. But it's producing that even when there's actually subtitles.

Note though that we return "no captions" for any error returned by the main loop (which sucks - the error can be anything else and we just happily convert it to no captions). 

Don't know why this would appear in Win and not linux but clearly there's something weird going on here.

@NilsIrl can you check that out, see if you can figure it out?

NilsIrl:
> Windows fails with error code 10 on some samples, which means NO_CAPTIONS. But it's producing that even when there's actually subtitles.

Is this a regression that was introduced in this PR? Because comparing with [the results from master](https://sampleplatform.ccextractor.org/test/master/windows), I can't see a difference.

canihavesomecoffee:
You are right @NilsIrl. I looked back a bit further, and at least since the release from 0.88 they have been returning non-expected return codes.

As such, this PR doesn't make it worse, and can be merged 👍

PR #1168, opened 2020-01-01
[IMPROVEMENT] Clang format

NilsIrl:

---

Defines a coding standard so that each pr doesn't change the formatting.

Also please give feedback on what rules to use. I just set the ones that I noticed followed the existing style.


NilsIrl:
It would also be nice to add a github action or travis that fails if a PR doesn't match the standard

NilsIrl:
Also the only file manually edited is `.clang-format` in this PR.

cfsmp3:
This we won't merge - at least for now. It changes too many files. It's a huge PR, too hard to check. Maybe after releasing 0.89 we can have this discussion.

PR #1170, opened 2020-01-02
[FIX] Stop searching for tessdata

### LINKED ISSUE #1162: [BUG] TESSDATA_PREFIX requires path separator at its end
**Necessary information**
* arguments: just the filename (including location): 
```
$ TESSDATA_PREFIX=/nix/store/8lr60hp7yv0aysns056b74fsi8fm49zg-tesseract-3.05.00/share/ ./result/bin/ccextractor ~/Downloads/telecine.ts
```

* platform: NixOS (Linux)
* regression: I don't know

I **hope** that it works with any file that uses tesseract (files that store subtitles as images). Because if it isn't then it means that the location of the tesseract data is dealt in a separate way for different files.

I used the one from #1104 (https://edge1.motv.eu/telecine.ts)

**Additional information**

`TESSDATA_PREFIX` is an environment variable that points to the directory/folder containing the tessdata directory/folder. For some reason, ccextractor requires `TESSDATA_PREFIX` to finish with a `/`. It should work without one.

e.g.

```
TESSDATA_PREFIX=/nix/store/8lr60hp7yv0aysns056b74fsi8fm49zg-tesseract-3.05.00/share
```

Should work but it doesn't.

NilsIrl:

---

Removes logic in ccextractor that looks for `tessdata`, in favor of passing `NULL` to tesseract and let it find the directory.

This means that it fixes:
* Fix #1162 
* OCR on NixOS

### DIFF
--- a/src/lib_ccx/hardsubx.c
+++ b/src/lib_ccx/hardsubx.c
@@ -225,37 +225,17 @@ struct lib_hardsubx_ctx* _init_hardsubx(struct ccx_s_options *options)
     char* lang = options->ocrlang;
     if(!lang) lang = "eng"; // English is default language
 
-    tessdata_path = probe_tessdata_location(lang);
-    if(!tessdata_path)
-    {
-        if (strcmp(lang, "eng") == 0)
-        {
-            mprint("eng.traineddata not found! No Switching Possible\n");
-            return NULL;
-        }
-        mprint("%s.traineddata not found! Switching to English\n", lang);
-        lang = "eng";
-        tessdata_path = probe_tessdata_location("eng");
-        if(!tessdata_path)
-        {
-            mprint("eng.traineddata not found! No Switching Possible\n");
-            return NULL;
-        }
-    }
-
     int ret = -1;
 
     if (!strncmp("4.", TessVersion(), 2))
     {
-        char tess_path [1024];
-        snprintf(tess_path, 1024, "%s%s%s", tessdata_path, "/", "tessdata");
         //ccx_options.ocr_oem are deprecated and only supported mode is OEM_LSTM_ONLY
-        ret = TessBaseAPIInit4(ctx->tess_handle, tess_path, lang, 1, NULL, 0, &pars_vec,
+        ret = TessBaseAPIInit4(ctx->tess_handle, NULL, lang, 1, NULL, 0, &pars_vec,
             &pars_values, 1, false);
     }
     else
     {
-        ret = TessBaseAPIInit4(ctx->tess_handle, tessdata_path, lang, ccx_options.ocr_oem, NULL, 0, &pars_vec,
+        ret = TessBaseAPIInit4(ctx->tess_handle, NULL, lang, ccx_options.ocr_oem, NULL, 0, &pars_vec,
             &pars_values, 1, false);
     }
 
@@ -263,6 +243,8 @@ struct lib_hardsubx_ctx* _init_hardsubx(struct ccx_s_options *options)
 	free(pars_values);
 	if(ret != 0)
 	{
+		// TODO: this is not necessarily the reason, tesseract might have not
+		// been able to find trained data
 		fatal(EXIT_NOT_ENOUGH_MEMORY, "Not enough memory to initialize Tesseract");
 	}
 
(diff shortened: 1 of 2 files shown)


NilsIrl:
When it is unable to find the data this is the output:

Error opening data file /nix/store/ifwvhazh5ggbwmh6bgppc4c76nblqjcf-tesseract-3.05.00/share/tessdata/por.traineddata
Please make sure the TESSDATA_PREFIX environment variable is set to the parent directory of your "tessdata" directory.
Failed loading language 'por'
Tesseract couldn't load any languages!

cfsmp3:
Reason for this?

NilsIrl:
@cfsmp3 otherwise it doesn't work on NixOS. also as I said fixes #1162

NilsIrl:
I guess it also makes the thing generally more robust as we let tesseract do it rather than do it ourself

NilsIrl:
The paths aren't hardcoded anymore

cfsmp3:
What I see there is that we are removing functionality instead of improving whatever is broken.

If the train data is not found, are we now providing a better error message (or recovering somehow) than what we used to have?

If we remove everything that doesn't work on a specific distro or for a specific user we're going to end up with a rather thin program :-)

NilsIrl:
We could change the error message if that's the problem. But now we are more likely to find the trained data.

NilsIrl:
Also we aren't recovering in a worse way than before. The exact same thing happen functionality wise (NULL is returned and the program continues).

cfsmp3:
I think it's a good idea to tell the user *where* we are looking for the data (or where tesseract is looking for the data) and how to change that rather than just exit or write a vague error if there's no data. 

If we're changing it it has to be to improve things for everyone, not just make it worse for most people because of the NixOS craziness :-)

NilsIrl:
> I think it's a good idea to tell the user where we are looking for the data

yes why not. But we didn't do that before so...

NilsIrl:
> I think it's a good idea to tell the user where we are looking for the data

Ironically enough tesseract does it actually https://github.com/CCExtractor/ccextractor/pull/1170#issuecomment-570248722

NilsIrl:
This doesn't only affect NixOS, it also allows someone to compile their own tesseract and have it work as well.

cfsmp3:
> > I think it's a good idea to tell the user where we are looking for the data
> 
> Ironically enough tesseract does it actually [#1170 (comment)](https://github.com/CCExtractor/ccextractor/pull/1170#issuecomment-570248722)

A good idea would to be to check github's blame to figure out who added that and then look for the discussion about it - most likely there was a reasonable explanation to implement it that way.

NilsIrl:
This was changed to silent tesseract it seems: https://github.com/CCExtractor/ccextractor/commit/5e0f9a48989892e620cb281c9f88dda0f26027e9

That PR didn't even change it for hardsubx.c at the time.

which does reduce functionality. This was part of a bigger PR to move to tesseract4. However, this PR still works on tesseract 4.

NilsIrl:
Just tested on Ubuntu (via a VM) and it works as expected.

anshul1912:
I think you will break ubuntu version 4 with it, I think it may work on nixOS but break Ubuntu.
what is location of tessdata in your nixOS installation. If you are using only enviorment variable TESSDATA_PREFIX then as you see in function first priority is given to environment variable.
if there is default location in nixOS tessdata but enviorment variable is not set. Then you must add that location in probe function

NilsIrl:
> I think you will break ubuntu version 4 with it

I tested with tesseract 4 on ubuntu and it worked.

> TESSDATA_PREFIX then as you see in function first priority is given to environment variable

Tesseract puts priority on TESSDATA_PREFIX

cfsmp3:
@NilsIrl It's up to you to convince @anshul1912 that this is correct :-) It's his code and he can merge, so I will stay away from this one; if up to you guys.

anshul1912:
> Just tested on Ubuntu (via a VM) and it works as expected.

Which version of Ubuntu and Tesseract?

NilsIrl:
> Which version of Ubuntu and Tesseract?

Tesseract 4.

And ubuntu 18.04 or 19.10 (I don't remember)

cfsmp3:
Closing for now. I think

1) char* probe_tessdata_location(const char *lang) 

needs to be rewritten - use an array of default locations and a loop :-)

2) Our output needs to be useful. Report where tessdata is looking, do our own research (as in "we couldn't find the dictionary in any of these locations"), explain how to fix the problem and so on.

PR #1175, opened 2020-01-10
[IMPROVEMENT] Remove useless if statement

NilsIrl:

---

Increases readability (IMO).

I was told by canihavesomecoffee that code improvement PRs are always welcome so I opened this.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -206,8 +206,6 @@ int api_start(struct ccx_s_options api_options)
 #ifdef ENABLE_FFMPEG
             case CCX_SM_FFMPEG:
 #endif
-                if (!api_options.use_gop_as_pts) // If !0 then the user selected something
-                    api_options.use_gop_as_pts = 0;
                 if (api_options.ignore_pts_jumps)
                     ccx_common_timing_settings.disable_sync_check = 1;
                 mprint("\rAnalyzing data in general mode\n");
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -204,9 +204,9 @@ int api_start(struct ccx_s_options api_options)
             case CCX_SM_GXF:
             case CCX_SM_MXF:
 #ifdef ENABLE_FFMPEG
-            case CCX_SM_FFMPEG:
+		case CCX_SM_FFMPEG:
 #endif
-                if (api_options.ignore_pts_jumps)
+	if (api_options.ignore_pts_jumps)
                     ccx_common_timing_settings.disable_sync_check = 1;
                 mprint("\rAnalyzing data in general mode\n");
                 tmp = general_loop(ctx);
--- a/src/lib_ccx/es_functions.c
+++ b/src/lib_ccx/es_functions.c
@@ -643,9 +643,9 @@ static int gop_header(struct encoder_ctx *enc_ctx, struct lib_cc_decode *dec_ctx
 		dec_ctx->saw_gop_header = 1;
 
 		// If we use GOP timing, reconstruct the PTS from the GOP
-		if (ccx_options.use_gop_as_pts==1)
+		if (ccx_options.use_gop_as_pts == 1)
 		{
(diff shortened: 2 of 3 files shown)


src/ccextractor.c
```diff
@@ -202,8 +202,7 @@ int api_start(struct ccx_s_options api_options)
 #ifdef ENABLE_FFMPEG
                 case CCX_SM_FFMPEG:
 #endif
-                if (!api_options.use_gop_as_pts) // If !0 then the user selected something
-                    api_options.use_gop_as_pts = 0;
+                api_options.use_gop_as_pts = 0;
```
canihavesomecoffee: I'm not so sure if this is the correct logic.

if the use_gop_as_pts is set to 1, it'll skip the statement under old conditions. Now it'd be set to 0.
It seems more logic to me to just delete the two lines.

@cfsmp3 What do you think?

canihavesomecoffee: And since this has been in since the original commit of the repo, we definitely need @cfsmp3 to tell something about it :+1:

NilsIrl: I wrote this more than 1 month ago and just realised it doesn't make sense at all.

if it is 0 then !0 == 1

so the variable is set to 0

And if it is 1 (or anything else) nothing happens.

NilsIrl: I removed the lines.

Also my previous commit was totally wrong as it would zero the variable when it is one which it didn't before

cfsmp3:
This is hiding a bug, not fixing it. Yes, that if is useless, but it's useless because it's not doing what it should be doing, which is passing a user option to the right place:

We have this parameter:
```
    -nogt --nogoptime: Never use GOP timing (use PTS), even if ccextractor
                       detects GOP timing is the reasonable choice.
```

Which I haven't tried but my guess is that it's broken since that if doesn't do anything.

You can grep that variable to get an idea

```
/lib_ccx/es_functions.c:		if (ccx_options.use_gop_as_pts==1)
./lib_ccx/es_functions.c:	if (ccx_options.use_gop_as_pts!=1)
./lib_ccx/ccx_common_option.c:	options->use_gop_as_pts = 0; // Use GOP instead of PTS timing (0=do as needed, 1=always, -1=never)
./lib_ccx/ccx_common_option.h:	int use_gop_as_pts;               // Use GOP instead of PTS timing (0=do as needed, 1=always, -1=never)
./lib_ccx/params_dump.c:	switch (ccx_options.use_gop_as_pts)
./lib_ccx/configuration.c:	{"GOP_TIME",offsetof(struct ccx_s_options,use_gop_as_pts),set_int},
./lib_ccx/params.c:			opt->use_gop_as_pts = 1;
./lib_ccx/params.c:			opt->use_gop_as_pts = -1; // Don't use even if we would want to
./lib_ccx/sequencing.c:	if (ccx_options.use_gop_as_pts==1)
./lib_ccx/sequencing.c:		// Since use_gop_as_pts this is not needed anymore, but left
./ccextractor.c:                if (!api_options.use_gop_as_pts) // If !0 then the user selected something
./ccextractor.c:                    api_options.use_gop_as_pts = 1; // Force GOP timing for ES
./ccextractor.c:                if (!api_options.use_gop_as_pts) // If !0 then the user selected something
./ccextractor.c:                    api_options.use_gop_as_pts = 0;
```

So can you fix the actual problem?

NilsIrl:
Notes: this was added on the initial commit

cfsmp3:
Just because code is old or original it doesn't mean it's correct.

NilsIrl:
> Just because code is old or original it doesn't mean it's correct

Of course. It's just that I git blamed it to see why it had been changed to what it is.

And I thought it was best to report this to prevent someone wasting their time trying to figure out the same thing.

NilsIrl:
What's the difference between pts and gop?

There are files where the option is checked and stuff are done or aren't done depending on the value of the option.

cfsmp3:
pts => Presentation Timestamp
GOP => Group of Pictures (which may have its own timing information)

cfsmp3:
Closing (I really think we'd be hiding a bug)

PR #1176, opened 2020-01-12
[FIX] Added the option to disable timestamps for WebVTT

dhrumilp15:

---

In response to issue #1127, this fix adds an option to remove the X-TIMESTAMP-MAP header for WebVTT.

### DIFF
--- a/src/lib_ccx/ccx_encoders_webvtt.c
+++ b/src/lib_ccx/ccx_encoders_webvtt.c
@@ -216,7 +216,7 @@ int write_webvtt_header(struct encoder_ctx *context)
 		millis_to_time(context->timing->sync_pts2fts_fts, &h1, &m1, &s1, &ms1);
 		 
 		 // If the user has not disabled the X-TIMESTAMP-MAP header
-		if (!ccx_options.no_timestamp_map)
+		if (!context->no_timestamp_map)
 		{
 			sprintf(header_string, "X-TIMESTAMP-MAP=MPEGTS:%ld,LOCAL:%02u:%02u:%02u.%03u%s",
 				context->timing->sync_pts2fts_pts, h1, m1, s1, ms1,
--- a/src/lib_ccx/ccx_encoders_webvtt.c
+++ b/src/lib_ccx/ccx_encoders_webvtt.c
@@ -215,7 +215,7 @@ int write_webvtt_header(struct encoder_ctx *context)
 		unsigned h1, m1, s1, ms1;
 		millis_to_time(context->timing->sync_pts2fts_fts, &h1, &m1, &s1, &ms1);
 		 
-		 // If the user has not disabled the X-TIMESTAMP-MAP header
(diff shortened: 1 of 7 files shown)


NilsIrl:
The problem isn't that we want to remove X-TIMESTAMP-MAP. It's that X-TIMESTAMP-MAP isn't always present.

dhrumilp15:
Now you've confused me - I thought issue #1127 asked for an option to enable / disable the X-TIMESTAMP-MAP header for WebVTT.

NilsIrl:
Yes, that's what it originally asked for. But this assumed X-TIMESTAMP-MAP is always present.

NilsIrl:
If you read the comments on the issue it make clear things up (or make them even worse :joy: )

dhrumilp15:
I've been reading the comments on the issue for several days now and I am just as confused as I was before I had read the comments

NilsIrl:
Well keep, your PR as it is and just make so that in the case where we want timestamps, show them.

dhrumilp15:
My PR doesn't disable the header, it's just adding the option for the --no-timestamp-map flag to disable the X-TIMESTAMP-MAP header. Also, doesn't the X-TIMESTAMP-MAP header only apply to WebVTT?

NilsIrl:
> doesn't the X-TIMESTAMP-MAP header only apply to WebVTT

yes

NilsIrl:
#1137 fixes the problem

I didn't find it good so it was closed

dhrumilp15:
DVB and WebVTT are separate formats, so why would the X-TIMESTAMP-MAP header be in DVB? This is regarding your fix to the issue's thread

NilsIrl:
DVB is a video format

src/lib_ccx/ccx_encoders_webvtt.c
```diff
@@ -214,9 +214,19 @@ int write_webvtt_header(struct encoder_ctx *context)
 		int used;
 		unsigned h1, m1, s1, ms1;
 		millis_to_time(context->timing->sync_pts2fts_fts, &h1, &m1, &s1, &ms1);
-		sprintf(header_string, "X-TIMESTAMP-MAP=MPEGTS:%ld,LOCAL:%02u:%02u:%02u.%03u%s",
-			context->timing->sync_pts2fts_pts, h1, m1, s1, ms1,
-			ccx_options.enc_cfg.line_terminator_lf ? "\n\n" : "\r\n\r\n");
+		 
+		 // If the user has not disabled the X-TIMESTAMP-MAP header
+		if (!ccx_options.no_timestamp_map)
```
cfsmp3: Encoders should get options from their context, not directly from the user parameters. 
So you need to copy from the user options to the encoder context (you'll see plenty of examples of this in the code, and I believe the tutorial we have about this about does it).

dhrumilp15: I added the change!

PR #1178, opened 2020-01-13
[IMPROVEMENT] Comment out issue

NilsIrl:

---

This was discussed already.

This fixes the problem where issues have the instructions in them which can decrease readability.


NilsIrl:
So you would rather have no comment at all? And make it clear that the text has to be removed

cfsmp3:
> So you would rather have no comment at all? And make it clear that the text has to be removed

Correct. Issues need to meet 3 requirements IMHO

- Be exhaustive
- Include everything we need (i.e. samples) to reproduce
- Include the basic information on top, so no boilerplate stuff that makes all issues look identical at first glance.

PR #1179, opened 2020-01-13
[IMPROVEMENT] Add noreturn attribute to fatal

NilsIrl:

---

Removes the following warnings:

```patch
52a52,55
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/asf_functions.c: In function ‘asf_readval’:
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/asf_functions.c:39:9: warning: ‘rval’ may be used uninitialized in this function [-Wmaybe-uninitialized]
>   return rval;
>          ^~~~
1652a1656,1663
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/general_loop.c: In function ‘process_data’:
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/general_loop.c:741:5: warning: ‘got’ may be used uninitialized in this function [-Wmaybe-uninitialized]
>   if (got > data_node->len)
>      ^
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/general_loop.c: In function ‘general_loop’:
> /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/general_loop.c:896:9: warning: ‘get_more_data’ may be used uninitialized in this function [-Wmaybe-uninitialized]
>    ret = get_more_data(ctx, &datalist);
>          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/lib_ccx/lib_ccx.h
+++ b/src/lib_ccx/lib_ccx.h
@@ -250,6 +250,7 @@ int myth_loop(struct lib_ccx_ctx *ctx);
 int matroska_loop(struct lib_ccx_ctx *ctx);
 
 // utility.c
+__attribute__ ((noreturn))
 void fatal(int exit_code, const char *fmt, ...);
 void mprint (const char *fmt, ...);
 void sleep_secs (int secs);
--- a/src/lib_ccx/lib_ccx.h
+++ b/src/lib_ccx/lib_ccx.h
@@ -250,7 +250,11 @@ int myth_loop(struct lib_ccx_ctx *ctx);
 int matroska_loop(struct lib_ccx_ctx *ctx);
 
 // utility.c
+#if defined(__GNUC__)
 __attribute__ ((noreturn))
+#elif defined(_MSC_VER)
+__declspec(noreturn)
+#endif
 void fatal(int exit_code, const char *fmt, ...);
 void mprint (const char *fmt, ...);
 void sleep_secs (int secs);


src/lib_ccx/lib_ccx.h
```diff
@@ -250,7 +250,7 @@ int myth_loop(struct lib_ccx_ctx *ctx);
 int matroska_loop(struct lib_ccx_ctx *ctx);
 
 // utility.c
-void fatal(int exit_code, const char *fmt, ...);
+void fatal(int exit_code, const char *fmt, ...) __attribute__ ((noreturn));
```
cfsmp3: That's not portable - remember, we're not just using GCC, we also use Visual Studio for the Windows builds!

NilsIrl: I was internally dreaming MSVC had the same extension.

Would you be okay with some pre processor stuff to use the equivalent MSVC extenison?

https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2012/dabb5z75(v=vs.110)?redirectedfrom=MSDN

NilsIrl: There also `_Noreturn` which is in the C11 standard

cfsmp3:
This is still not portable, just adds support to more compilers :-)

NilsIrl:
> This is still not portable, just adds support to more compilers :-)

Well, because they're behind #if-s, it will build on any compilers but they will have more warnings.

And even if we used C11 (`stdnoreturn.h`) it wouldn't make it portable to older compilers and most importantly MSVC doesn't even support that.

NilsIrl:
It might actually be better if a NORETURN (or a similar identifier was declared because `fatal` isn't the only place where a warning is emitted because of this.

Another place is `ccx_common_structs.h:15`.

kdrag0n:
@cfsmp3 Any updates regarding this issue? The attribute may not be that portable, but there is no functional change without them. I think it's safe to say that someone using an exotic compiler can deal with a couple extra warnings — it's certainly no worse than before, even on unsupported compilers.

This would also reduce the number of changes in #1204 and #1205, making them cleaner and less intrusive.

PR #1180, opened 2020-01-13
[IMPROVEMENT] Remove freep warning

NilsIrl:

---

Change `freep` into a macro to remove all the warnings it generated.

### DIFF
--- a/src/gpacmp4/mp4.c
+++ b/src/gpacmp4/mp4.c
@@ -688,7 +688,7 @@ int processmp4 (struct lib_ccx_ctx *ctx, struct ccx_s_mp4Cfg *cfg, char *file)
 
 	}
 
-	freep(&dec_ctx->xds_ctx);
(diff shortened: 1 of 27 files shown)


cfsmp3:
Hold on, this must be completely breaking things - if you don't pass a pointer to the pointer you want to free then you can't change the pointer value from inside freep() which is the whole point of it in the first place!
```
char *hello = (char *) malloc (10)
freep (&hello);
assert (hello == NULL);
```

That's going to fail if you pass "hello" instead of "&hello", because freep() won't be able to change the value of the outside variable.

NilsIrl:
Or we could pass a `void *` to `freep` and then cast in there.

cfsmp3:
I've merged your stuff. I don't want to cast to void * in 30 places.

NilsIrl:
> I don't want to cast to void * in 30 places.

187 to be exact :joy:

PR #1185, opened 2020-01-15
[IMPROVEMENT] Improve and simplify dprintf implementation

NilsIrl:

---

It now returns a value like the rest of the printf family. It doesn't
brute force the amount of memory that needs to be allocated.

It also removes a warning.

```patch
0a1
> Scanning dependencies of target ccx
60,63d60
< /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/ccx_common_common.c: In function ‘fdprintf’:
< /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/ccx_common_common.c:27:4: warning: ignoring return value of ‘write’, declared with attribute warn_unused_result [-Wunused-result]
<    27 |    write(fd, p, n);
<       |    ^~~~~~~~~~~~~~~
69c66
<     inlined from ‘add_cc_sub_text’ at /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/ccx_common_common.c:95:3:
---
>     inlined from ‘add_cc_sub_text’ at /home/nils/Documents/codein2019/ccextractor/src/lib_ccx/ccx_common_common.c:73:3:

```

I do not believe there should be any performance concerns with this
implementation as it is what `glibc` does (if anything it probably improves performance by a lot):

https://code.woboq.org/userspace/glibc/libio/iovdprintf.c.html

### DIFF
--- a/src/lib_ccx/ccx_common_common.c
+++ b/src/lib_ccx/ccx_common_common.c
@@ -3,47 +3,25 @@
 int cc608_parity_table[256];
 
 /* printf() for fd instead of FILE*, since dprintf is not portable */
-void fdprintf(int fd, const char *fmt, ...)
+int fdprintf(int fd, const char *fmt, ...)
 {
-	/* Guess we need no more than 100 bytes. */
-	int n, size = 100;
-	char *p, *np;
+	FILE *file;
+
+	if ((file = fdopen(dup(fd), "w")) == NULL)
+	{
+		return -1;
+	}
+
 	va_list ap;
 
-	if (fd < 0)
-		return;
-	if ((p = (char *)malloc(size)) == NULL)
-		return;
+	va_start(ap, fmt);
+	int ret = vfprintf(file, fmt, ap);
+	fclose(file);
+	va_end(ap);
 
-	while (1)
-	{
-		/* Try to print in the allocated space. */
-		va_start(ap, fmt);
-		n = vsnprintf(p, size, fmt, ap);
-		va_end(ap);
-		/* If that worked, return the string. */
-		if (n > -1 && n < size)
-		{
-			write(fd, p, n);
-			free(p);
-			return;
-		}
-		/* Else try again with more space. */
-		if (n > -1)    /* glibc 2.1 */
-			size = n + 1; /* precisely what is needed */
-		else           /* glibc 2.0 */
-			size *= 2;  /* twice the old size */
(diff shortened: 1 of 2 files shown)


cfsmp3:
Don't remember why we decided to do it that way but the original code comes straight from here https://linux.die.net/man/3/snprintf

Anyway how tested is your change? (and what does it bring to the table - how is it better?)

The writing to fileno instead of file handles is so we could write to sockets, by the way. Not everything is a file :-) That's also why we assume we can't seek negatively and so on.

NilsIrl:
> The writing to fileno instead of file handles is so we could write to sockets, by the way. Not everything is a file :-) That's also why we assume we can't seek negatively and so on.

You can write to sockets using `FILE *`.

```c
// To test run:
//
// ```bash
// $ nc -l 9090
// ```


#include <sys/socket.h>
#include <netinet/in.h>
#include <arpa/inet.h>
#include <stdio.h>
#include <unistd.h>

int main()
{
	int socket_fd = socket(AF_INET, SOCK_STREAM, 0);
	struct sockaddr_in serv_addr;
	
	serv_addr.sin_family = AF_INET;
	serv_addr.sin_port = htons(9090);

	if (!inet_aton("127.0.0.1", &serv_addr.sin_addr)) {
		printf("oopsie");
		return 1;
	}

	if (connect(socket_fd, (struct sockaddr *) &serv_addr, sizeof(serv_addr)) == -1) {
		printf("couln't connect rip ;(");
		return 1;
	}


	FILE *socket_file = fdopen(dup(socket_fd), "r+");
	fprintf(socket_file, "hello %s, my name is %d\n", "nils", 45);
	fclose(socket_file);

	char buf[512];

	ssize_t bytes_read = read(socket_fd, buf, 512);
	write(socket_fd, buf, bytes_read);

	dprintf(socket_fd, "here was your message again: %s\n", buf);

	return 0;
}
```

cfsmp3:
How do you create a socket with FILE * -  you just get a file descriptor and then wrap it into a FILE * with fdopen?

Note that the FILE functions include buffering. We often explicitly don't want that.

NilsIrl:
All I'm saying is that there is no reason this should break sending stuff over the network. And buffering isn't a problem in this specific use case.

cfsmp3:
OK, but still - why do you want to replace working code with untested code? What do we gain by doing so?

PR #1202, opened 2020-01-20
[FIX] segfault on windows

NilsIrl:

---

Fixes seg fault on windows. Unfortunately we loose a bit of semantics :( with the enum.

This is more acceptable as it reverts some stuff from the SCC & CCD PR.

### DIFF
--- a/src/lib_ccx/ccx_decoders_structs.h
+++ b/src/lib_ccx/ccx_decoders_structs.h
@@ -85,8 +85,8 @@ struct eia608_screen // A CC buffer
 	/** format of data inside this structure */
 	enum ccx_eia608_format format;
 	unsigned char characters[15][33];
-	enum ccx_decoder_608_color_code colors[15][33];
-	enum font_bits fonts[15][33]; // Extra char at the end for a 0
+	unsigned char colors[15][33];
+	unsigned char fonts[15][33]; // Extra char at the end for a 0
 	int row_used[15];            // Any data in row?
 	int empty;                   // Buffer completely empty?
 	/** start time of this CC buffer */
--- a/src/lib_ccx/ccx_encoders_helpers.c
+++ b/src/lib_ccx/ccx_encoders_helpers.c
@@ -220,7 +220,7 @@ int clever_capitalize(struct encoder_ctx *context, int line_num, struct eia608_s
 // Encodes a generic string. Note that since we use the encoders for closed caption
 // data, text would have to be encoded as CCs... so using special characters here
 // is a bad idea.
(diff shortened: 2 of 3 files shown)


src/lib_ccx/ccx_decoders_structs.h
```diff
@@ -85,8 +85,8 @@ struct eia608_screen // A CC buffer
 	/** format of data inside this structure */
 	enum ccx_eia608_format format;
 	unsigned char characters[15][33];
-	enum ccx_decoder_608_color_code colors[15][33];
-	enum font_bits fonts[15][33]; // Extra char at the end for a 0
+	unsigned char colors[15][33];
```
cfsmp3: The only way I can think this fixing anything is by some other place using unsigned char instead of that enum. 
Let's correct the right place instead.

NilsIrl: It seems that it has to do with stuff being copied like so:
```c
memset(data->colors[i], context->settings->default_color, CCX_DECODER_608_SCREEN_WIDTH + 1);
```

NilsIrl: An enum is larger than 1 byte so it doesn't work

cfsmp3: An enum can be any size actually (and not all enums have to be the same size) - so let's use sizeof() where needed instead of removing the one hint we have about the type .-)

But good catch! Let's do _the_ best fix rather than _a_ fix :-)

NilsIrl: > An enum can be any size actually

how do you do that?

NilsIrl: Like this? https://stackoverflow.com/questions/13789462/is-possible-to-set-default-type-for-enum-to-be-unsigned-char

NilsIrl: Or when we memcpy, we will have to skip the most significant bytes of the enum?

NilsIrl: my bad as suggested by my previous comment, I always saw those `memset` as `memcpy` so that confused me.

NilsIrl: I know how to fix it

cfsmp3: Just use sizeof() :-) Don't replace it with #define or anything like that. Please please don't do "clever" stuff like copying only half the bits or anything like it :-)

NilsIrl:
superseded by #1203

PR #1204, opened 2020-01-21
[FIX] GCC warning fixes

kdrag0n:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [x] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).** (combined in #1205)

**My familiarity with the project is as follows (check one):**

- [x] I have used CCExtractor just a couple of times.

---

This fixes all warnings reported by GCC 9.2.0 in the core CCExtractor code, excluding vendored libraries.

Several actual issues were caught in the process:

- [Missing new line in a CEA-708 log message](https://github.com/CCExtractor/ccextractor/commit/a920bab0d664395ad975653f9828365750a35efb) (not a warning, but caught in the process nonetheless)
- [Empty encoding in parameter dump when ASCII is used](https://github.com/CCExtractor/ccextractor/commit/670d5a045584505f750381269756662f0bfe393a)
- [Incorrect stream mode comparisons](https://github.com/CCExtractor/ccextractor/commit/af9027a29d4a00a619e18801450b394d6dbe1202)

These changes make all of the core code build with zero warnings on the latest stable version of GCC at the time of writing, both with and without OCR enabled.

PR #1205, opened 2020-01-21
[FIX] Clang warning fixes

kdrag0n:

---

This fixes all warnings reported by Clang 9.0.1 in the core CCExtractor code, excluding vendored libraries.

These changes combined with #1204 make all of the core code build with zero warnings on the latest stable versions of both GCC and Clang at the time of writing, both with and without OCR enabled.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -173,10 +173,10 @@ int change_utf8_encoding(unsigned char *dest, unsigned char *src, int len, enum 
 				{
 					if ((src[1] & 0x40) == 0)
 					{
-						c = utf8_to_latin1_map((((unsigned int)(src[0] & 0x1F)) << 6)
+						int cp = utf8_to_latin1_map((((unsigned int)(src[0] & 0x1F)) << 6)
 							| ((unsigned int)(src[1] & 0x3F)));
-						if (c < 256)
-							*dest++ = c;
+						if (cp < 256)
+							*dest++ = cp;
 						else
 							*dest++ = '?';
 					}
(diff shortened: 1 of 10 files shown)


kdrag0n:
Updated as well.

src/lib_ccx/ccx_encoders_common.c
```diff
@@ -226,10 +217,7 @@ int change_utf8_encoding(unsigned char* dest, unsigned char* src, int len, enum
 						| (((unsigned int)(src[2] & 0x3F)) << 12U)
 						| (((unsigned int)(src[3] & 0x3F)) << 6U)
 						|  ((unsigned int)(src[4] & 0x3FU)) );
-					if (c < 256)
-						*(dest++) = c;
-					else
-						*dest++ = '?';
+					*(dest++) = c;
```
cfsmp3: This is not correct.
I understand that of course c is always going to be less than 256 because it's a char here, but if look at the function where it takes its value it returns an integer - so what we need to change is the type of c.

src/lib_ccx/ocr.c
```diff
@@ -526,7 +527,7 @@ char* ocr_bitmap(void* arg, png_color *palette,png_byte *alpha, unsigned char* i
 				else if(max==g_avg) h = 60 * ((b_avg - r_avg)/(max - min)) + 120;
 				else h = 60 * ((r_avg - g_avg)/(max - min)) + 240;
 
-				if(abs(h-h0)>50) // Color has changed
+				if(fabsf(h-h0)>50) // Color has changed
```
cfsmp3: Why? Do we have floats?

kdrag0n: `h` is a float (see a few lines above), and `h0` is declared as a float near the beginning of this block even though it's an integer constant.

kdrag0n:
Rebased, fixed some newly introduced warnings, and addressed the concern about UTF-8 codepoint types.

PR #1207, opened 2020-01-21
[IMPROVEMENT] Remove useless O(N) operations and memory allocations

NilsIrl:

---

And move some processing to compile time. It also uses functions that don't check needlessly against '\0'.

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -283,7 +283,6 @@ char* ocr_bitmap(void* arg, png_color *palette,png_byte *alpha, unsigned char* i
 	PIX*  cpix_gs       = NULL; // Grayscale version
 	PIX*  color_pix     = NULL;
 	PIX*  color_pix_out = NULL;
-	char* text_out      = NULL;
 	int i,j,index;
 	unsigned int wpl;
 	unsigned int *data,*ppixel;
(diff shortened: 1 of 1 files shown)


src/lib_ccx/ocr.c
```diff
@@ -532,43 +533,37 @@ char* ocr_bitmap(void* arg, png_color *palette,png_byte *alpha, unsigned char* i
 					if(ccx_options.write_format==CCX_OF_SRT ||
 					   ccx_options.write_format==CCX_OF_WEBVTT)
 					{
-						char *substr;
+						char substr[sizeof("</font><font color=\"#000000\">")];
```
cfsmp3: A hardcoded string that appears more than once definitely belongs in a #define - otherwise a change in one of the instances will break stuff (particularly if sizeof is involved)

NilsIrl: It only appears once now.

I also managed to remove another O(N) operation. (copying sprintf output).

cfsmp3: We're processing streams, O(N) is the best we can possibly ask for :-) Don't obsess over micro-optimizations. If you see O (N^2) then that's something to worry about. But a 10 character memcpy() is not worth the time, even if it can be avoided.

NilsIrl: yeah I know :joy:

it's not gonna change a lot.

> O(N) is the best we can possibly ask for

yes but it's best to not copy stuff around when you don't have to

Issue #1191, opened 2020-01-18
[BUG/PROPOSAL] SSC and CCD feature implementation


NilsIrl:
Follow up to #1154

Features that are unimplemented:

- [ ] Preamble codes that set the row and font at the same time
- [ ] Colors
- [ ] Allow the user to select the type of line ending they want ("\n" vs "\r\n")
- [ ] The different timing types

cfsmp3:
Closing since this has already been done and merged.

kdrag0n:
There are still parts of this that are unimplemented.

kdrag0n:
> The different timing types - #1221

The different SCC timing types (dropframe and non-dropframe) are actually unrelated to #1221, which is about accounting for over-the-wire transmission time to make players like YouTube happy with the output.

Issue #1120, opened 2019-11-10
[PROPOSAL] Add an option to output Scenarist Closed Caption format (*.SCC) and/or McPoodle's Disassembly format (*.CCD)


The-Bart-The:
Please prefix your issue with one of the following: [BUG], [PROPOSAL], [QUESTION].

CCExtractor version (using the --version parameter preferably) : THIS WOULD APPLY TO VERSIONS AFTER **0.88**

**In raising this issue, I confirm the following (please check boxes, eg [X] - and delete unchecked ones):**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [X] I have checked that the issue I'm posting isn't already reported.
- [X] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [X] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [X] I have used the latest available version of CCExtractor to verify this issue exists.

**My familiarity with the project is as follows (check one, eg [X] - and delete unchecked ones):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor. (I have posted once before. I use CCExtractor A LOT.)

**Necessary information**
- Is this a regression (did it work before)? [X] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [X] Windows - [ ] Linux - [ ] Mac
- What were the used arguments? `-autoprogram` (not applicable yet)

**Video links (replace text below with your links)**

Please make the affected input file available for us (no screenshots, those don't help!). Public links to Dropbox, Google Drive, etc, are all fine. If it is not possible to make it available publicly, send us a private invitation (both Dropbox and Google Drive allow that). In this case we will download the file and upload it to the private developer repository.

[...]

PR #1209, opened 2020-01-22
[IMPROVEMENT] Write CRLF line endings in SCC files

kdrag0n:

---

All the SCC and CCD examples I can find have CRLF line endings. VLC and libavformat (used by MPV) don't care, so just go with the popular convention and switch to CRLF. There's no reason a user would want to choose their line endings in this scenario.

This also adds 2 trailing new lines as is the convention for SCC/CCD files.

This addresses one of the items in #1191.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -97,7 +97,7 @@ static const char *webvtt_header[] = { "WEBVTT", "\r\n", NULL };
 
 static const char *simple_xml_header = "<?xml version=\"1.0\" encoding=\"utf-8\"?>\r\n<captions>\r\n";
 
-static const char CCD_HEADER[] = "SCC_disassembly V1.2\n";
+static const char CCD_HEADER[] = "SCC_disassembly V1.2\r\n";
 static const char SCC_HEADER[] = "Scenarist_SCC V1.0";
 
 void find_limit_characters(const unsigned char *line, int *first_non_blank, int *last_non_blank, int max_len)
@@ -391,8 +391,7 @@ int write_subtitle_file_footer(struct encoder_ctx *ctx, struct ccx_s_write *out)
 			break;
 		case CCX_OF_SCC:
 		case CCX_OF_CCD:
-			// TODO \n vs \r\n
-			ret = write(out->fh, "\n", 1);
+			ret = write(out->fh, "\r\n", 2);
 			break;
 		default: // Nothing to do, no footer on this format
 			break;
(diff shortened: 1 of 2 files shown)


NilsIrl:
> There's no reason a user would want to choose their line endings in this scenario.

Why?

kdrag0n:
> Why?

Every other SCC file I've seen so far has been in CRLF format, and I haven't been able to find a player that doesn't understand CRLF. There's no need to add unnecessary complexity (arguments, help, and code to implement it) when nobody will use it. If someone wants LF for some exotic situation, they can open an issue to request it and demonstrate that there *is* a use case for it.

cfsmp3:
Take a look at this: 
http://www.theneitherworld.com/mcpoodle/SCC_TOOLS/DOCS/SCC_FORMAT.HTML

I haven't seen any official SCC specifications so McPoodle's reverse engineering work is all we've got.

I'd say we should stick to the output of McPoodle's tools. Looking at the source code he uses \n, not \r\n.

NilsIrl:
My plan was to default to a certain value (maybe depending on the platform) and then let the user override this like they can already with some output formats.

> unnecessary complexity (arguments, help, and code to implement it)

There is already all the code present for this that you can borrow from other output formats.

cfsmp3:
> My plan was to default to a certain value (maybe depending on the platform) and then let the user 

But the platform that consumes the output is not necessarily the same as the one that generates it.  When there is an official line ending that is parts of a spec document we have to follow it.

When there is no specification (or reference implementation) then we can either use platform (which CCExtractor does for several formats, but possibly incorrectly).

For SCC - I think we should do what McPoodle does, and test with youtube.

NilsIrl:
> When there is an official line ending that is parts of a spec document we have to follow it.

That's why I said "maybe". But at the same time, if the spec document says to follow something, I wouldn't see the point of being able to choose.

kdrag0n:
YouTube seems to be fine with both, just like all the other players I've tested.

It doesn't look like those tools are very popular because all of the SCC I've seen from other programs are using CRLF:

- All of the examples posted in #1120
- https://archive.org/details/cc_sample
- http://www.capsubservices.com/assets/downloads/caption/DVD%20File%2001%20Hour.scc

If there's no official specification and no specific preference by players, shouldn't we be following the popular convention? If someone writes a CRLF-only player based on their own observations of most other SCCs seen in the wild, our output would be incompatible. It would increase compatibility and simplify things since we wouldn't need to add a switch for it and the user wouldn't ever have to care.

NilsIrl:
These 4 lines explain the solution perfectly IMO:

https://github.com/CCExtractor/ccextractor/blob/4097831b9bcb2686930d3aaa3e9e190e55a24b62/src/lib_ccx/ccx_encoders_common.c#L1041-L1045

cfsmp3:
I've asked on the video-dev slack for SCC samples generated by scenarist. 
I'm closing this PR for now (I hate how CLRF looks on Linux) but if the official tool uses CRLF I'll reopen and merge.

kdrag0n:
Shouldn't this be re-opened since the official sample you provided is CRLF?

cfsmp3:
Yes - I'll merge when the conflicts are resolved.

kdrag0n:
Rebased and dropped double-newline endings since that doesn't seem to be correct.

PR #1212, opened 2020-01-23
[IMPROVEMENT] Clean up SCC control codes

kdrag0n:

---

This makes the control code handling more readable by adding a small info struct that contains all the information about a control code (first byte odd & even, second byte, and assembly). Information is stored in and retrieved from an array, created using an array initializer with the enum values as indices.

This allows us to remove the massive switch-case blocks, leading to much cleaner and more streamlined code.

I have verified that the output is identical using `diff` for both SCC and CCD formats.

**Depends on #1211.**


cfsmp3:
There are conflicts now, can you correct @kdrag0n ?

kdrag0n:
Updated.

NilsIrl:
I don't think this is a good way of doing it. The whole reason I refactored it like that is so it could be flexible.

PR #1213, opened 2020-01-23
[IMPROVEMENT] Add SCC color output support

kdrag0n:

---

This adds support for all colors defined [here](http://www.theneitherworld.com/mcpoodle/SCC_TOOLS/DOCS/CC_CODES.HTML), as well as their italic variants. **Note that I have not tested the output** because MPV doesn't appear to support SCC colors, VLC freezes while trying to open any SCC file, and I haven't been able to find any other players that support SCC. It looks fine with a visual comparison to the examples from #1120, but an actual test would be nice to have before merging.

### DIFF
--- a/src/lib_ccx/ccx_decoders_structs.h
+++ b/src/lib_ccx/ccx_decoders_structs.h
@@ -70,7 +70,10 @@ enum ccx_decoder_608_color_code
 	COL_MAGENTA = 6,
 	COL_USERDEFINED = 7,
 	COL_BLACK = 8,
-	COL_TRANSPARENT = 9
+	COL_TRANSPARENT = 9,
+
+	// Must keep at end
+	COL_MAX
 };
 
 /**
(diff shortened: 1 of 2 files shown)


kdrag0n:
Updated this as well, though it probably shouldn't be merged yet due to the lack of player testing.

cfsmp3:
Use youtube as a testing ground since they support SCC.

NilsIrl:
The person who requested SCC (and CCD) sent a file with colors. Either in the original implementation PR or in the issue.

NilsIrl:
https://github.com/CCExtractor/ccextractor/issues/1120#issuecomment-568306562

kdrag0n:
I'm aware of the examples, I used those to test and compare the output. The problem is that I can't find an SCC *player* which supports color — YouTube and MPV appear to only support white. VLC was able to display color when decoding the original EIA-608 data, but I can't get it to load any SCC files.

kdrag0n:
Rebased.

PR #1219, opened 2020-01-28
[IMPROVEMENT] Move dependencies to a third party directory

NilsIrl:

---

Move thirdparty dependencies to /src/thirdparty and remove redundant code in linux/builddebug and linux/build_hardsubx.

Add `.clang-format` and github action to check if the code is formatted.

Why?

* Allows to easily differentiate between dependencies and ccextractor code:
  - Makes it a **lot** less confusing to new contributors as to why there are files everywhere.
  - Tools can leverage this to only act on some files (e.g. `clang-format`)


NilsIrl:
I've also added a clang-format GitHub action that checks if the code is formatted correctly.

As you can see, it currently fails because the code isn't formatted.

To format the code the following command has to be ran:

```
find src/ -type f -name '*.c' -name '*.c' | xargs clang-format -i
```

If you want I can add a commit with the formatted code. Otherwise, this can be merged and once it is merged, a commit with the code formatted should be pushed.

NilsIrl:
To have an idea of what the current `.clang-format` gives I have a branch which clang-format has been run on which branches from this branch:

https://github.com/CCExtractor/ccextractor/compare/master...NilsIrl:formatted

Please tell me if something about the format is wrong/doesn't fit the current code style.

This branch can also be merged after.

cfsmp3:
cases inside switch needs to be indented

https://github.com/NilsIrl/ccextractor/blob/11f82e6fa6d13601e586c0197ec88f29bffce650/src/ccextractor.c


(line 120 for example)

canihavesomecoffee:
@NilsIrl I'd rather like to see that new GitHub Action for formatting to be in a separate PR. Makes it easier to merge and keep them separated.

canihavesomecoffee:
Once the action is gone (moved to separate PR), I think we can merge this one already.

NilsIrl:
> @NilsIrl I'd rather like to see that new GitHub Action for formatting to be in a separate PR. Makes it easier to merge and keep them separated.

Should the addition of `.clang-format` be moved to another PR as well? If so, should it be the same as the GitHub action PR?

canihavesomecoffee:
> 
> 
> > @NilsIrl I'd rather like to see that new GitHub Action for formatting to be in a separate PR. Makes it easier to merge and keep them separated.
> 
> Should the addition of `.clang-format` be moved to another PR as well? If so, should it be in the same PR as the GitHub action PR?

Yes, I'd do that in the same PR where you introduce the action, as they are related :)

NilsIrl:
Done

Issue #1223, opened 2020-01-31
MP4 (from itunes) files not being processed (no subs)


cfsmp3:
This comes from an email:

```
CCExtractor 0.88, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
Input: D:\Documents\NoteBurner Video Converter\Adam's Rib (1080p HD).mp4
[Extract: 1] [Stream mode: Autodetect]
[Program : Auto ] [Hauppage mode: No] [Use MythTV code: Auto]
[Timing mode: Auto] [Debug: No] [Buffer input: Yes]
[Use pic_order_cnt_lsb for H.264: No] [Print CC decoder traces: No]
[Target format: .srt] [Encoding: Latin-1] [Delay: 0] [Trim lines: No]
[Add font color data: Yes] [Add font typesetting: Yes]
[Convert case: No] [Video-edit join: No]
[Extraction start time: not set (from start)]
[Extraction end time: not set (to end)]
[Live stream: No] [Clock frequency: 90000]
[Teletext page: Autodetect]
[Start credits text: None]
[Quantisation-mode: CCExtractor's internal function]


-----------------------------------------------------------------

Opening file: D:\Documents\NoteBurner Video Converter\Adam's Rib (1080p HD).mp4
Detected MP4 box with name: ftyp
Detected MP4 box with name: free
Detected MP4 box with name: mdat

File seems to be a MP4

Analyzing data with GPAC (MP4 library)

Analyzing data with GPAC (MP4 library)
Opening 'D:\Documents\NoteBurner Video Converter\Adam's Rib (1080p HD).mp4':

This is where the program hangs and doesn’t finish on all of my MP4 conversions
Running PID various numbers come up each time I try
Status Code: Done, Code = - 1073741819

There is nothing that comes up on the preview screen
And the file output for .srt is 0kb
```

The file is available [here](https://drive.google.com/open?id=19cVFdWwhY9stTNIay6xXepWA5524uuR8).

eshandhawan51:
CCExtractor 0.88, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
Input: /home/eshan/Downloads/Adam's Rib.mp4
[Extract: 1] [Stream mode: Autodetect]
[...]

PR #1233, opened 2020-02-17
[FIX] MP4 (from itunes) files not being processed (no subs) (#1223)

growupboron:

---

Fixed the issue by regressing the changes, as discussed on the issue page, 0.85 release was working fine, on further investigation, 0.86 release was also correct. From 0.87 release, this issue started to pop up, so regressed the code back to 0.86 release for the full functionality.

### DIFF
--- a/src/thirdparty/gpacmp4/mp4.c
+++ b/src/thirdparty/gpacmp4/mp4.c
@@ -469,24 +469,23 @@ static int process_tx3g(struct lib_ccx_ctx *ctx, struct encoder_ctx *enc_ctx,
 	return -1; // Assume there's only one subtitle in one atom.
 }
 
+
+//Reverting back to 0.86 release code
 /*
 	Here is application algorithm described in some C-like pseudo code:
 		main(){
 			media = open()
 			for each track in media
-				switch track
-				AVC track:
+				if track is AVC track
 					for each sample in track
 						for each NALU in sample
 							send to avc.c for processing
-				CC track:
-					for each sample in track
-						deliver to corresponding process_xxx functions
 			close(media)
 		}
 
 */
-int processmp4 (struct lib_ccx_ctx *ctx, struct ccx_s_mp4Cfg *cfg, char *file)
+
+int processmp4 (struct lib_ccx_ctx *ctx,struct ccx_s_mp4Cfg *cfg, char *file)
 {
 	int mp4_ret = 0;
 	GF_ISOFile* f;
@@ -508,7 +507,7 @@ int processmp4 (struct lib_ccx_ctx *ctx, struct ccx_s_mp4Cfg *cfg, char *file)
 
 	if((f = gf_isom_open(file, GF_ISOM_OPEN_READ, NULL)) == NULL)
 	{
-		mprint("Failed to open input file (gf_isom_open() returned error)\n");
+		mprint("failed to open\n");
 		free(dec_ctx->xds_ctx);
 		return -2;
 	}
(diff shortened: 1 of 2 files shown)


src/thirdparty/gpacmp4/mp4.c
```diff
@@ -528,189 +527,255 @@ int processmp4 (struct lib_ccx_ctx *ctx, struct ccx_s_mp4Cfg *cfg, char *file)
...
 					do_NAL(enc_ctx, dec_ctx, (unsigned char *)seqcnf->data, seqcnf->size, &dec_sub);
+					//do_NAL (dec_ctx, (unsigned char *) seqcnf->data, seqcnf->size, &dec_sub);
 				}
 			}
-			if (process_avc_track(ctx, file, f, i + 1, &dec_sub) != 0) {
-				mprint("Error on process_avc_track()\n");
+
+			if(process_avc_track(ctx, file, f, i + 1, &dec_sub) != 0)
+			{
+				mprint("error\n");
```
cfsmp3: I pretty much prefer errors message to be specific. Error on process_avc_track() is much better than "error" (same everywhere else).

src/thirdparty/gpacmp4/mp4.c
```diff
@@ -528,189 +527,255 @@ int processmp4 (struct lib_ccx_ctx *ctx, struct ccx_s_mp4Cfg *cfg, char *file)
...
-			(unsigned char)((type >> 8) % 0x100), (unsigned char)(type % 0x100),
-			(unsigned char)(subtype >> 24 % 0x100), (unsigned char)((subtype >> 16) % 0x100),
-			(unsigned char)((subtype >> 8) % 0x100), (unsigned char)(subtype % 0x100));
-
-		const u64 track_type = MEDIA_TYPE(type, subtype);
 
-		switch (track_type) {
-		case MEDIA_TYPE(GF_ISOM_MEDIA_VISUAL, GF_ISOM_SUBTYPE_XDVB): //vide:xdvb
+		if ( type == GF_ISOM_MEDIA_VISUAL && subtype == GF_ISOM_SUBTYPE_XDVB)
+		{
```
cfsmp3: Is there a problem with MEDIA_TYPE? I think the switch() version is cleaner and having if's.

NilsIrl:
@growupboron could you link the commit you're reverting?

growupboron:
> IMHO going back to 0.86 is going to fix that regression but it's also going to remove some fixes that the original PR introduced, so I'd say it's a better idea to fix the actual bug that was introduced (which visually I can't see to be honest, so it will require some actual debugging).

Yes sir, I get the point.  I will get back to fixing this issue, once my mid terms get over in a week, sorry for the late reply. The actually debugging might take some time, but i look forward to the challenge.

src/thirdparty/gpacmp4/mp4.c
growupboron: Will keep in mind.

src/thirdparty/gpacmp4/mp4.c
growupboron: I think so, the problem might be with MEDIA_TYPE, I'll take of the switch() version, here.

growupboron:
> @growupboron could you link the commit you're reverting?



> @growupboron could you link the commit you're reverting?

@NilsIrl I just cloned the 0.86 and regressed the code.

cfsmp3:
Closing - seems to be abandoned?

PR #1236, opened 2020-02-22
[FIX] Typo in README.TXT

lostboy529:

---

{pull request content here}

### DIFF
--- a/docs/README.TXT
+++ b/docs/README.TXT
@@ -58,7 +58,7 @@ Essential CC related information and free (with source) tools.
 
 Encoding
 --------
-This version, in both its Linux and Windows builds generates by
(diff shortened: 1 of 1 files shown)


docs/README.TXT
```diff
@@ -58,7 +58,7 @@ Essential CC related information and free (with source) tools.
 
 Encoding
 --------
-This version, in both its Linux and Windows builds generates by
+This version, in both its Linux and Windows build generates by
```
cfsmp3: That's not a typo - build here is a noun, not a verb and there's two builds - the Linux build and the Windows build. We could use binaries instead, of versions. But replacing builds with build is not correct.

PR #1247, opened 2020-03-28
[IMPROVEMENT] Apply clang-format to all remaining files

canihavesomecoffee:

---

Applies clang-format to all failing files:
- GUI files
- lib_ccx files
- wrappers files

Excludes running checks on:
- third party (src/thirdparty)
- icons file of GUI


.github/workflows/format.yml
```diff
@@ -18,5 +18,5 @@ jobs:
       - uses: actions/checkout@v1
       - name: Format code
         run: |
-          find src/ -type f -name '*.c' -name '*.c' | xargs clang-format -i
+          find src/ -type f -name '*.c' -not -path "src/thirdparty/*" -not -path "src/GUI/icon_data.c" | xargs clang-format -i
```
NilsIrl: Well I guess that's a way to do it :joy:

canihavesomecoffee: Well, we shouldn't touch third party, and that one file is something I'd rather not touch either for now. All the rest is neatly formatted from now on :)

NilsIrl:
Why is the branch name "part1"?

.github/workflows/format.yml
NilsIrl: We already don't touch third party: https://github.com/CCExtractor/ccextractor/blob/master/src/thirdparty/.clang-format

canihavesomecoffee:
> Why is the branch name "part1"?

Because I intended to split it up and then changed my mind :)

.github/workflows/format.yml
canihavesomecoffee: Ah, good remark on the config file for third party. I'll remove that.

Issue #677, opened 2017-02-02
[CEA-708] Is there a way to extract subtitles


fsh240led:
The others are extracted without problems
But the video below is not extract
https://docs.google.com/uc?id=0B42jO8cBUe6baEtqLTNkRnExbVk&export=download

PotPlayer displays subtitles
http://i.imgur.com/SX9KApM.png

cfsmp3:
Are yo sure you aren't using an external subtitle file? I just installed
potplayer for this and it says there's no subtitles available.

fsh240led:
Preferences required

Preferences(F5) -> Filter Control -> Video Decoder -> Built-in codec/DXVA settings -> Enable Closed Captioning (Check)

cfsmp3:
I'm probably missing some fonts, this is what a see after enabling CC. 

![sbsplus](https://cloud.githubusercontent.com/assets/5949913/22720320/e8414a66-ed5e-11e6-81e9-34209d27298e.png)

fsh240led:
korean font download link
[http://ko.cooltext.com/Fonts-Unicode-Korean](url)

fsh240led:
[[img]http://i.imgur.com/SX9KApM.png[/img]](url)

fsh240led:
`[img]http://i.imgur.com/SX9KApM.png[/img]`

cfsmp3:
Assigning to @Izaron since he was the last one looking into Korean.

cfsmp3:
GSoC qualification: 3 points.

Izaron:
@fsh240led Are this output **at 00:09** from the other video player is incorrect also in comparing with PotPlayer? Can you confirm it?

![default](https://cloud.githubusercontent.com/assets/5406399/22783469/034ae50e-eedd-11e6-8fcb-bd1b77a0baad.png)

@cfsmp3 and @AlexBratosin2001 (since you're PTS man)
Problem is there: https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/stream_functions.c#L307
Can you give me specs about this binary stream, when `(nextheader[7]&0xC0) == 0xC0`? Maybe there just problem with hexadecimal hardcoded constants.

fsh240led:
It is an unknown words
It should be output as

<img src="http://i.imgur.com/SX9KApM.png">

mahalwal:
[...]

PR #1304, opened 2021-03-12
Ignore extra padding data in the current_packet

PunitLodha:

---
Fixes issue #677 
Instead of resetting the decoder, when current_length doesn't equal the len in header, change current_length to len(ignore the extra data as padding)

### DIFF
--- a/src/lib_ccx/ccx_decoders_708.c
+++ b/src/lib_ccx/ccx_decoders_708.c
@@ -1743,8 +1743,8 @@ void ccx_dtvcc_process_current_packet(ccx_dtvcc_ctx *dtvcc)
 #endif
 	if (dtvcc->current_packet_length != len)
 	{
-		// Extra padding data, can be ignored
-		dtvcc->current_packet_length = len;
+		// Most likely things are going to be bad for us
+		len = dtvcc->current_packet_length; // At least don't read beyond the buffer
 	}
 	if (dtvcc->last_sequence != CCX_DTVCC_NO_LAST_SEQUENCE &&
 	    (dtvcc->last_sequence + 1) % 4 != seq)
--- a/src/lib_ccx/ccx_decoders_708.c
+++ b/src/lib_ccx/ccx_decoders_708.c
@@ -1741,10 +1741,10 @@ void ccx_dtvcc_process_current_packet(ccx_dtvcc_ctx *dtvcc)
 				   "Sequence: %d, packet length: %d\n",
 				   seq, len);
 #endif
-	if (dtvcc->current_packet_length != len) // Is this possible?
+	if (dtvcc->current_packet_length != len)
 	{
-		_dtvcc_decoders_reset(dtvcc);
-		return;
+		// Extra padding data, can be ignored
+		dtvcc->current_packet_length = len;
 	}
 	if (dtvcc->last_sequence != CCX_DTVCC_NO_LAST_SEQUENCE &&
 	    (dtvcc->last_sequence + 1) % 4 != seq)


PunitLodha:
The CI is failing on code that was not changed by me. How can i fix that?

canihavesomecoffee:
> The CI is failing on code that was not changed by me. How can i fix that?

If it's the format sourcecode, it's not a big problem.

PunitLodha:
@cfsmp3 Can you please review this PR. Also, I don't know why the tests are failing, is it something that I have to fix? or is that not a problem?

src/lib_ccx/ccx_decoders_708.c
```diff
@@ -1741,10 +1741,10 @@ void ccx_dtvcc_process_current_packet(ccx_dtvcc_ctx *dtvcc)
 				   "Sequence: %d, packet length: %d\n",
 				   seq, len);
 #endif
-	if (dtvcc->current_packet_length != len) // Is this possible?
+	if (dtvcc->current_packet_length != len)
 	{
-		_dtvcc_decoders_reset(dtvcc);
-		return;
+		// Extra padding data, can be ignored
+		dtvcc->current_packet_length = len;
```
cfsmp3, line 1747: It's a small thing but worth talking about :-) That variable (current_packet_length) is used here src/lib_ccx/ccx_dtvcc.c by the loop that writes the CC data to the packet buffer... so it originally comes from ourselves.

Is there a bug in the loop? Or some assumption there is incorrect?

On visual inspection I think that if dtvcc->current_packet_length < len then we are in trouble: The encoded data says it has more data than what we actually have, so if we ignore it we're going to get into buffer overruns.

So _maybe_ (this would have to be tested) it would be saner to do 

if (dtvcc->current_packet_length < len) {
// Most likely things are going to be bad for us
len = dtvcc->current_packet_length ; // At least don't read beyond the buffer
}

I'm not sure it's a good idea to overwrite current_packet_length with len. Two reasons
- current_packet_length is the amount data we actually have while len is the amount of data that the file says we should have. So it would be trivial to create a file that makes CCExtractor crash if we trust the external source rather than ourselves :-)
- current_packet_length is not used later in this function at all, so changing it will not affect how data is processed here, but it might affect what happens later.

cfsmp3 (changes requested):

cfsmp3:
@PunitLodha reviewed

src/lib_ccx/ccx_decoders_708.c
PunitLodha: > Is there a bug in the loop? Or some assumption there is incorrect?

As far as i know, it is not a bug in the code. Instead i guess that given file might not follow the specification correctly, as i mentioned in [this comment](https://github.com/CCExtractor/ccextractor/issues/677#issuecomment-796490909) about the two "should be padding" bytes


>So maybe (this would have to be tested) it would be saner to do
if (dtvcc->current_packet_length < len) {
// Most likely things are going to be bad for us
len = dtvcc->current_packet_length ; // At least don't read beyond the buffer
}

I tested it. This causes the extra data to be considered as a service packet, which leads to it being a null service packet header (service number and block size both are 0). We ignore the any packet with service_number=0 in the current code. So i guess we should be fine with this approach


> current_packet_length is not used later in this function at all, so changing it will not affect how data is processed here, but it might affect what happens later.

This shouldn't be an issue because later in the function we clear the packet and set current length to 0


>current_packet_length is the amount data we actually have while len is the amount of data that the file says we should have. So it would be trivial to create a file that makes CCExtractor crash if we trust the external source rather than ourselves :-)

This definitely seems like an issue. We can go ahead with your approach then

PR #1312, opened 2021-03-31
Remove -cf

cfsmp3:
Remove -cf which is just a rather bad attempt to convert a TS or PS to a ES (by just writing non-headers).

I added it maybe in CCExtractor 0.02 to help me with debugging. I don't need anyone else has ever used it and I know better now :-)


canihavesomecoffee (approved):
LGTM

Issue #1252, opened 2020-04-06
[BUG] 0.88 crash on Windows


ahmedtds:
CCExtractor version: **0.88**

# In raising this issue, I confirm the following:

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.
- [x] I have ticked all the boxes in this section and to prove it I'm deleting the section completely to remove boilerplate text.

# Necessary information

- Is this a regression (i.e. did it work before)? **YES** - 0.85 works without an issue
- What platform did you use? **Window 7 32-bit**
- What were the used arguments? `ccextractorwin.exe --gui_mode_reports -out=srt -bom -utf8 --nofontcolor -trim file.m4v`

# Video links

Can provide privately, it's a movie from iTunes, can't post publicly.

# Additional information

```
CCExtractor 0.88, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
Input: C:\itunes\movie.m4v
[Extract: 1] [Stream mode: Autodetect]
[Program : Auto ] [Hauppage mode: No] [Use MythTV code: Auto]
[Timing mode: Auto] [Debug: No] [Buffer input: Yes]
[Use pic_order_cnt_lsb for H.264: No] [Print CC decoder traces: No]
[Target format: .srt] [Encoding: UTF-8] [Delay: 0] [Trim lines: Yes]
[Add font color data: No] [Add font typesetting: Yes]
[Convert case: No] [Video-edit join: No]
[Extraction start time: not set (from start)]
[Extraction end time: not set (to end)]
[Live stream: No] [Clock frequency: 90000]
[Teletext page: Autodetect]
[Start credits text: None]
[Quantisation-mode: CCExtractor's internal function]


[...]

PR #1313, opened 2021-04-02
[FIX] Fix segfault on Windows

pranavrajpal:

-   [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
-   [x] I have checked that another pull request for this purpose does not exist.
-   [x] I have considered, and confirmed that this submission will be valuable to others.
-   [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
-   [x] I give this submission freely, and claim no ownership to its content.
-   [x] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

-   [ ] I have never used CCExtractor.
-   [x] I have used CCExtractor just a couple of times.
-   [ ] I absolutely love CCExtractor, but have not contributed previously.
-   [ ] I am an active contributor to CCExtractor.

---

Using the format specifier %d to print out size is technically undefined behavior, as size is defined as a u64, while %d is meant to print out an int (which is typically 32 bits). This causes a segfault on Windows as this apparently causes the wrong pointer to be passed in for the filename.

The segfault was caused with the file from #1223, but I don't know if it was related as I was not able to reproduce it hanging as mentioned in that issue. This also appears to be similar to #1252 as the output right before the crash is very similar to the output in https://github.com/CCExtractor/ccextractor/issues/1252#issuecomment-657259396 and the change that causes this is bbe2f333997fe8d90f3851ba6dcdbd5d58e3ecf0, which is before 0.88 was released.

### DIFF
--- a/src/thirdparty/gpacmp4/gpac/internal/isomedia_dev.h
+++ b/src/thirdparty/gpacmp4/gpac/internal/isomedia_dev.h
@@ -22,6 +22,7 @@
  *  the Free Software Foundation, 675 Mass Ave, Cambridge, MA 02139, USA.
  *
  */
+#include <inttypes.h>
 
 #ifndef _GF_ISOMEDIA_DEV_H_
 #define _GF_ISOMEDIA_DEV_H_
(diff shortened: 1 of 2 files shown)


src/thirdparty/gpacmp4/gpac/internal/isomedia_dev.h
```diff
@@ -509,7 +510,7 @@ typedef struct
 	tmp->type = __4cc;
 
 #define ISOM_DECREASE_SIZE(__ptr, bytes)	if (__ptr->size < (bytes) ) {\
-			GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("[isom] not enough bytes in box %s: %d left, reading %d (file %s, line %d)\n", gf_4cc_to_str(__ptr->type), __ptr->size, (bytes), __FILE__, __LINE__ )); \
+			GF_LOG(GF_LOG_ERROR, GF_LOG_CONTAINER, ("[isom] not enough bytes in box %s: %" PRIu64 " left, reading %d (file %s, line %d)\n", gf_4cc_to_str(__ptr->type), __ptr->size, (bytes), __FILE__, __LINE__ )); \
```
canihavesomecoffee, line 513: Since this seems to be internal code from GPAC itself, did you verify this same issue isn't present on the upstream? If it's fixed there we might want to upgrade GPAC, if it's not, it might be more interesting to report it over there too.

cfsmp3:
I'm merging this, seems sane :-)
Fixing upstream and/or updating GPAC would be even better of course.

src/thirdparty/gpacmp4/gpac/internal/isomedia_dev.h
pranavrajpal: I haven't checked that yet but I can check right now.

PR #1316, opened 2021-04-04
Build with Bazel

cfsmp3:
First working version of Bazel. Notes:
- Binary is different (in size, and not by a small margin) to other options, so not 100% sure it's correct?
-  First time I (or @DivyamAhuja) use bazel. Most likely this can be _seriously_ improved.


cfsmp3:
@kvshravan take a look at this

Issue #646, opened 2017-01-14
[CEA-708] Missing the last subtitle with "Premature end of file"


Izaron:
Sometimes we have incomplete ts parts because of cropped .ts files, for example:
`Premature end of file - Transport Stream packet is incomplete (expected 188 bytes, got 168).`

After work we must flush subtitles in current window:
```
Done, processing time = 85 seconds
Issues? Open a ticket here
https://github.com/CCExtractor/ccextractor/issues
[CEA-708] Window 0 dump:

00:04:49:623 --> 00:34:57:394
-놳떵볒뾡 낥 맼쇋뢦 샺쇶뢣룩 샚믬쟒 
뿫뇢낡 믽뇢쇶 뻊삻뇮 뷍뻮벭 샚웷샚뇢 
뷉솤삸럎 맼쟠삻 쟟듙냭. 
[CEA-708] Dump done
[CEA-708] Window 1 dump:

00:30:22:755 --> 00:34:57:394
낭뢪뾡벭 몸뎻뗥뢰 욯쇽 뒺붺820 뾩뇢벭 
뢶쒡냚뷀듏듙. 
-(뻞쒿) 뷃쎻쟘쇖뷅 뾩랯뫐, 냭뢿뷀듏듙. 
[CEA-708] Dump done
```
^ This subtitles are written to the file
But if we have broken .ts file at the end, then we can't flush last sub because of missing show time, but CEA-608 have last sub. Example:
```
Done, processing time = 12 seconds
Issues? Open a ticket here
https://github.com/CCExtractor/ccextractor/issues
[CEA-708] _dtvcc_decoder_flush: Flushing decoder
[CEA-708] [W-1] hide time updated to 00:09:59:131
[CEA-708] _dtvcc_window_copy_to_screen: W-1
[CEA-708] For window 1: Anchor point -> 0, size 2:32, real position 65:0
[CEA-708] we have top [65] and left [0]
[CEA-708] 2*32 will be copied to the TV.
[CEA-708] Screen show time: -00:00:00:001 -> -00:00:00:001
[CEA-708] Screen hide time: -00:00:00:001 -> 00:09:59:131
[CEA-708] Window 1 dump:

-00:00:00:001 --> 00:09:59:131
Our suspect told Petty
Officer Lynn a co-worker
[CEA-708] Dump done
[CEA-708] _dtvcc_screen_print
[CEA-708] Screen hide time: 00:09:59:131 -> 00:09:59:131
[...]

PR #1319, opened 2021-04-12
Fix 708 timing issue

### LINKED ISSUE #641: [CEA-708] [Timing] Catchable bug with timing
It is known that timing in our CEA-708 is different from the timing of CEA-708 in video player with a difference of a second or so and in general it's not critical error.
But sometimes the timing differs a lot - end of the time goes ahead ~10 seconds or so in different examples. This always happens due to this bug where window is undefined (**decoder->current_window == -1**) and it's col_count too: [line](https://github.com/Izaron/ccextractor/blob/7078f1015024d68115d8659a9187224d0d7ad363/src/lib_ccx/ccx_decoders_708_output.c#L122) (and line 129). It may be a one line fix or maybe hell with codecs, but I must sleep and somewhere else, so if you want we can fix it together.

### LINKED ISSUE #677: [CEA-708] Is there a way to extract subtitles

PunitLodha:
Process packet as soon as the packet len is equal to the specified len

---
Process current packet as soon as the packet length is equal to the specified length. This leads to the packet being processed in the current FTS, instead of being processed in the next FTS, like before. Fixes #641 

Also ends up fixing #646, as the last packet is now processed instead of waiting for a next packet which never comes

This is causing the issue #677 again. Working on a fix for that

### DIFF
--- a/src/lib_ccx/ccx_decoders_708.c
+++ b/src/lib_ccx/ccx_decoders_708.c
@@ -193,6 +193,7 @@ ccx_dtvcc_window_attribs ccx_dtvcc_predefined_window_styles[] =
 void ccx_dtvcc_clear_packet(ccx_dtvcc_ctx *ctx)
 {
 	ctx->current_packet_length = 0;
+	ctx->is_current_packet_header_parsed = 0;
 	memset(ctx->current_packet, 0, CCX_DTVCC_MAX_PACKET_LENGTH * sizeof(unsigned char));
 }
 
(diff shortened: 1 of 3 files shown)


PunitLodha:
Now #677 is fixed. Also the condition of `dtvcc->current_packet_length != len` is never triggered now. So, should I revert the changes of #1304 due to this?

cfsmp3:
> Now #677 is fixed. Also the condition of `dtvcc->current_packet_length != len` is never triggered now. So, should I revert the changes of #1304 due to this?

Yes, let's revert code that was there to workaround a bug

PR #894, opened 2018-01-10
[IMPROVEMENT] Upgrade libgpac based on Hori75's branch

harrynull:
Please prefix your pull request with one of the following: **[FEATURE]** **[FIX]** **[IMPROVEMENT]**.

**In raising this pull request, I confirm the following (please check boxes):**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [X] I have checked that another pull request for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [X] I am an active contributor to CCExtractor.

---
Upgrade libgpac to <del>0.7.0</del> 0.7.1.

The pr fixes the failing sample, **based on** @Hori75 **'s pr** #871 .

List of changes I have made: (For future references when upgrading it...)
- Change `box_funcs.c`: comment out some of the lines to avoid introduce too many files. See [dfb26f49a265c1cf383a4016fc2ca015bcfb7d4d](https://github.com/CCExtractor/ccextractor/pull/894/commits/dfb26f49a265c1cf383a4016fc2ca015bcfb7d4d)
- In `gpac/configuration.h`, comment out all `#define GPAC_MEMORY_TRACKING` (there are currently two of it) and `#else #error "Unknown target platform used with static configuration file"` (To avoid compilation errors in Linux)
- In `gpac/isomedia.h`, add more constants. See [dfb26f49a265c1cf383a4016fc2ca015bcfb7d4d](https://github.com/CCExtractor/ccextractor/pull/894/commits/dfb26f49a265c1cf383a4016fc2ca015bcfb7d4d)
- Make `gpac/revision.h` yourself; update the version.
- In `/gpac/setup.h`, change functions like `void* gf_malloc(size_t size);` to `#define gf_malloc malloc` (to avoid introducing `alloc.c`)
- In `isom_write.c`, remove line `#include <gpac/iso639.h>`

**Please don't merge until the CI platforms run it**

PR #1328, opened 2021-04-27
[IMPROVEMENT] Update GPAC to version 1.0.1

pranavrajpal:

---

This pull request update GPAC to version 1.0.1, which fixes the issue mentioned in #1313.

I made the following changes to GPAC to avoid having to pull in too many files:
- Remove webvtt handling code in gf_isom_dump_srt_track in `box_dump.c` (see 4a3f0ee85d80bf27bf2ff2cbb9e354ab95cc2b77)
- Comment out unused lines in `box_funcs.c` as recommended in #894 (see a03f0f3018d1bf363d55ed24f5197fc9a0323d61)
- Change all malloc-family functions declared in `gpac/setup.h` to `#define`s (see 0cea6a642e27db93b87e61924bc4aa66d8770400)
- Remove references to Remotery by removing an include in `gpac/tools.h` and removing calls to `gf_rmt_begin` and `gf_rmt_end` in os_divers.c (see 5c0c9cf71e17b95dc4fd8015d81040fd03858ce9)

I also added the python script `tools/update_gpac.py` to do some of the work involved in updating GPAC (It just copies in the newer version of every file without making any of the above changes).


pranavrajpal:
The clang-format CI check seems to be failing on code that this PR doesn't change (not on the GPAC code itself). Should I reformat all of those files in this PR, do that in a separate PR, or just ignore the clang-format check?

Also, I don't exactly understand why the Sample Platform tests are failing. From what I can tell, they don't seem to be related to the update as none of them are failing on any mp4 samples but I could be mistaken.

canihavesomecoffee:
Clang format can be ignored, that's no problem :)

With regards to the SP results: Linux seems okay to me (more or less the exact same failures as in #1329, so they're on master).  

For Windows it's a bit more unclear; 
- The broken ones in the CEA-708 category seem to have some unprintable characters added, so you might want to check that
- Same might be for the failing ones in the options category
- The failing one in XDS was a timeout, could happen if the Windows machine decided to do an update just when that sample was being processed

I'll re-run the Windows tests and see if it's still the same isues at the end of that run...

pranavrajpal:
Looking at [the Sample Platform page for this PR](https://sampleplatform.ccextractor.org/test/3229), I don't see any failures for the CEA-708 tests or the XDS tests.

I don't really know why the test in Options is failing because I couldn't manage to reproduce the failure on my computer. The only thing I can think of that might be causing the test error is if the path to the file with the list of words to capitalize is incorrect.

The list of test failures based on that Sample Platform page seems to be identical to the [list of test failures](https://sampleplatform.ccextractor.org/test/3225) for #1329 so I don't think this PR is causing the test failures.

Also, I noticed that the comment with the test results in GitHub shows different results than the page on Sample Platform, which I'm guessing is related to CCExtractor/sample-platform#535. I was looking mainly at the results on the Sample Platform but I'm not really sure which results are correct.

canihavesomecoffee:
In the meanwhile the tests for Windows were re-run (same page), and now they're consistent with the Linux results. Given that I stated earlier Linux was fine, it indeed means that this PR should be fine too.

cfsmp3:
OK this has been merged. I think with the current work during GSoC on getting our test platform working well again and all new talent joining us (happy about it) we should actually have a really good release in our hands soon-ish.

cfsmp3:
> OK this has been merged. I think with the current work during GSoC on getting our test platform working well again and all new talent joining us (happy about it) we should actually have a really good release in our hands soon-ish. Good job @pranavrajpal

PR #1332, opened 2021-05-05
[FIX] Multitrack, WebVTT, and Segfault issues

emkman99:

---

* The WEBVTT format issue was introduced in #1092 (after v0.88), so I didn't include it in the changelog. See [WebVTT file structure](https://www.w3.org/TR/webvtt1/#file-structure) for newline requirements.
* Added the WEBVTT header to multitrack output files (mp4).
* Generate multitrack filenames based on the output file instead of the input file. This fixes the situation where the -o argument is used and the first track is saved to the specified output location and the other tracks are saved next to the input file. I added `char *original_filename` to `struct ccx_s_write`, to avoid chained suffixes with three or more tracks (e.g. "test_2_3.vtt").

### DIFF
--- a/src/lib_ccx/output.c
+++ b/src/lib_ccx/output.c
@@ -12,6 +12,7 @@ void dinit_write(struct ccx_s_write *wb)
 	if (wb->fh > 0)
 		close(wb->fh);
 	freep(&wb->filename);
+	freep(&wb->original_filename);
 	if (wb->with_semaphore && wb->semaphore_filename)
 		unlink(wb->semaphore_filename);
 	freep(&wb->semaphore_filename);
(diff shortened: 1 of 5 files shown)


src/lib_ccx/output.c
```diff
@@ -50,6 +50,7 @@ int init_write(struct ccx_s_write *wb, char *filename, int with_semaphore)
 	wb->fh = -1;
 	wb->temporarily_closed = 0;
 	wb->filename = filename;
+	wb->original_filename = strdup(filename);
```
cfsmp3, line 53: This should be freed in dinit_write, otherwise we have a leak.
Other than that this PR LGTM. Can you correct that and I'll merge?

cfsmp3 (changes requested):

src/lib_ccx/output.c
emkman99: Ah, thanks for catching that. Fixed.

PR #1341, opened 2021-06-01
[FIX] Attempt to fix long-running regression in TeleText

canihavesomecoffee:

Regression test 78 (https://sampleplatform.ccextractor.org/regression/test/78/view)
has been broken since #622 was merged to fix other issues.

It's been traced back to be caused by not setting t0 at the correct time
(setting it using a calculated PTS time rather than taking it from the video frame),
and this commits attempts to fix that.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -31,6 +31,7 @@
 - Fix: Segmentation fault with unsupported and multitrack file reports
 - Fix: Write subtitle header to multitrack outputs
 - Fix: Write multitrack files to the output file directory
+- Fix: Regression on Teletext that caused dates to be wrong (RT 78 on the sample platform)
 
 0.88 (2019-05-21)
 -----------------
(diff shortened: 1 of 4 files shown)


cfsmp3 (commented):
Let's call this possibly passable :-)
Feel free to merge if the SP is happy with it.

alexbrt (approved):
As long as everything else is still in order, this looks good to me!

canihavesomecoffee:
> As long as everything else is still in order, this looks good to me!

Unfortunately it isn't :P

Timing's off by ~10 minutes, so I'll have to dig a bit further.

alexbrt:
I'll take a look at it myself these days, maybe 15yo @alexbrt omitted something in that PR :p

canihavesomecoffee:
@alexbrt @cfsmp3 I fixed it in a more simple way, using the current_pts rather than a fixed min_pts. This should only affect TT anyway, but I'm not sure if it'll break things Alex fixed...

Now it's just a difference of ~0.02s, due to it taking the PTS from the previous frame (new code) rather than the next (old code). This seems acceptable to me.

cfsmp3:
I suspect it will break multiprogram transport streams but I could be mistaken

canihavesomecoffee:
No regressions noticed when testing with Closedcaption_atsc_multiprog.ts (using `-out=srt -pn 2` with varying pn)

PR #1351, opened 2021-06-21
Add rust library

PunitLodha:

---
Added rust library, bindings for C functions and types required for the 708 decoder
Also ported the dtvcc_process_data() function

Need to add:-
- [ ] Documentation on how to build with rust library
- [ ] Github action for building rust library and formatting using clippy


canihavesomecoffee:
I'd suggest adding another build job on Linux to build the Rust version too :+1:

canihavesomecoffee:
LGTM, all tests on Linux seem to pass.

@cfsmp3 Merge if code looks good to you.

src/rust/src/decoder/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+    let dtvcc = Dtvcc::new(dtvcc);
+    let mut ctx = &mut *dtvcc.ctx;
+    let cc_valid = *data;
+    let cc_type = *data.add(1);
+
+    match cc_type {
+        // type 0 and 1 are for CEA 608 data and are handled before calling this function
+        // valid types for CEA 708 data are only 2 and 3
+        2 => {
+            println!("[CEA-708] dtvcc_process_data: DTVCC Channel Packet Data");
```
cfsmp3, line 36: Are these debug leftovers? I haven't tried it but doesn't this produce thousands of lines on console?

src/rust/src/decoder/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+                    }
+
+                    if ctx.current_packet_length >= max_len as i32 {
+                        dtvcc.process_current_packet(max_len as i32);
+                    }
+                }
+            }
+        }
+        3 => {
+            println!("[CEA-708] dtvcc_process_data: DTVCC Channel Packet Start");
```
cfsmp3, line 62: Same as above

src/rust/src/decoder/mod.rs
```diff
@@ -0,0 +1,97 @@
...
+/// | --- | --- |
+/// | cc_valid | 1 byte |
+/// | cc_type | 1 byte |
+/// | data | 2 bytes |
+///
+/// # Safety
+///
+/// dtvcc and data must not be null pointers
+#[no_mangle]
+pub unsafe extern "C" fn dtvcc_process_cc_data(
```
cfsmp3, line 23: Do we need the whole function to be unsafe? We want as many rust safeguards as possible.
I can live with some additional memory copy (since these are just a few bytes) in return of that.

cfsmp3 (changes requested):

src/rust/src/decoder/mod.rs
PunitLodha: These lines are for logs. I have temporarily added them as println. I have to figure out logging

src/rust/src/decoder/mod.rs
PunitLodha: Since we are dereferencing a raw pointer, we have to mark it as unsafe. We can change data to not be a raw pointer, but dtvcc might be bigger to copy

src/rust/src/decoder/mod.rs
cfsmp3: I know, but this makes the program unusable. This can't be merged as is. 
It's not that rust doesn't have good logging crates :-)

src/rust/src/decoder/mod.rs
cfsmp3: What you can do here is have a very small function that copies the data (which would have to be unsafe) and then pass it to the large function.

closed caption data is always a few bytes. It's not going to have a performance impact.

Let's prioritize correct code. Otherwise we'd have no reason to use rust in the first place.

src/rust/src/decoder/mod.rs
PunitLodha: Yes, I'll fix this

src/rust/src/decoder/mod.rs
PunitLodha: Ok

Issue #1352, opened 2021-06-22
[BUG] Windows - cmake build is failing


canihavesomecoffee:
CCExtractor version: master

# Necessary information

- Is this a regression (i.e. did it work before)? YES (probably)
- What platform did you use? Windows
- What were the used arguments? N/A

# Additional information

Building with cmake yields a "zlib.lib" missing fatal error (LNK1104) on Windows.

If we fix this, we should also extend our action workflow for windows to include cmake building.

Lavanyajc:
Hi  @canihavesomecoffee , I’m exploring CCExtractor for GSoC 2026 and focusing on DevOps/cloud-related contributions.

Along with trying to investigate this Windows CMake build issue, I’d also love to contribute more broadly — especially around Docker, CI/CD pipelines, or improving contributor environments.

If there are specific areas or tasks where help is needed in this direction, I’d really appreciate your guidance. I’m committed to contributing consistently and aligning my work with the project’s goals.

Rahul-2k4:
Hi! I noticed the earlier PR related to this issue was closed.
Is this still unresolved and open for someone to work on, or is there an updated fix planned/under review?
I’m happy to take this up and test on Windows if help is still needed.

cfsmp3:
Closing this issue as the Windows build is working correctly using the officially supported method (MSBuild with Visual Studio solution files).

**Current status:**
- Windows CI builds are passing: https://github.com/CCExtractor/ccextractor/actions/workflows/build_windows.yml
- The CI uses `msbuild ccextractor.sln` with vcpkg for dependencies

**Why CMake doesn't work on Windows:**
The CMake configuration uses `pkg_check_modules()` for finding dependencies (like GPAC), which requires pkg-config - a tool that isn't standard on Windows. The CMake setup was designed for Linux/macOS.

**Recommendation:**
For Windows builds, use the Visual Studio solution in the `windows/` directory, which is the officially supported and tested method. The solution works with vcpkg for dependency management.

If there's interest in adding CMake support for Windows in the future, that would be a new feature request requiring:
- Replacing pkg-config usage with `find_package()` or vcpkg toolchain integration
- Adding Windows-specific library discovery logic

For now, the MSBuild approach is reliable and well-tested.

PR #1355, opened 2021-06-24
Add CI and docs for rust lib

PunitLodha:

---
Add build and format CI for rust code
Add docs on how to compile rust code

Merge only after #1351 is merged

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -9,6 +9,7 @@ on:
     - '**Makefile**'
     - 'linux/**'
     - 'package_creators/**'
+    - 'src/rust/**'
   pull_request:
     types: [opened, synchronize, reopened]
     paths:
@@ -18,6 +19,7 @@ on:
     - '**Makefile**'
     - 'linux/**'
     - 'package_creators/**'
+    - 'src/rust/**'
 jobs:
   build_shell:
     runs-on: ubuntu-latest
@@ -87,3 +89,23 @@ jobs:
     - name: Display version information
       working-directory: ./bazel-bin
       run: ./src/ccextractor --version
+  build_rust:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2.3.4
+      - name: cache 
+        uses: actions/cache@v2
+        with:
+          path: |
+            src/rust/.cargo/registry
+            src/rust/.cargo/git
+            src/rust/target
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+          restore-keys: ${{ runner.os }}-cargo-
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+          override: true
+      - name: build
+        run: cargo build 
+        working-directory: ./src/rust
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -5,12 +5,14 @@ on:
     - '.github/workflows/format.yml'
     - 'src/**.c'
     - 'src/**.h'
+    - 'src/rust/**'
   pull_request:
     types: [opened, synchronize, reopened]
     paths:
     - '.github/workflows/format.yml'
     - 'src/**.c'
     - 'src/**.h'
+    - 'src/rust/**'
 jobs:
   format:
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


canihavesomecoffee:
What about Windows? Is it possible to add a workflow for building on Windows with Rust too?

PunitLodha:
I am facing some issues due to clang/llvm missing on windows. I'll try to fix it

canihavesomecoffee:
> I am facing some issues due to clang/llvm missing on windows. I'll try to fix it

Full code checking isn't needed (unless Rust of course has a dependency on clang/llvm), building the executable is what matters most :)

See what you can do, it's not mandatory for this PR to be mergeable (can be in a later one)

PunitLodha:
Clang is required for rust-bindgen to generate bindings for C code. I did figure that out, but now i get this error:- 

`  ./../lib_ccx/ccx_encoders_common.h:5:11: fatal error: '..\\win_iconv\\iconv.h' file not found, err: true`
https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_common.h#L4-L5

win_iconv is located in thirdparty folder, So shouldn't the path here be ..\\\thirdparty\\\win_iconv\\\iconv.h?

cfsmp3:
> win_iconv is located in thirdparty folder, So shouldn't the path here be ..\\thirdparty\\win_iconv\\iconv.h?

Possibly? Seems strange that the SP hasn't picked that error, though, but if it fixes for you by all means change it.

canihavesomecoffee:
> > win_iconv is located in thirdparty folder, So shouldn't the path here be ..\thirdparty\win_iconv\iconv.h?
> 
> Possibly? Seems strange that the SP hasn't picked that error, though, but if it fixes for you by all means change it.

Note that the cmake build on Windows has been failing for a while (see #1352)... It's not a hard requirement to complete a test run, because the "regular" manner to build the executables (through the VS project) works ok.

So that cmake build error might be hiding the issue Punit is encountering here.

PunitLodha:
> Possibly? Seems strange that the SP hasn't picked that error, though, but if it fixes for you by all means change it.

Changing the path does fix that issue. So I think I should make a different PR for that fix? 
Also, now I have run into another issue, 
```
 ./../lib_ccx/stdintmsc.h:69:27: error: typedef redefinition with different types ('char' vs 'signed char'), err: true
  ./../lib_ccx/stdintmsc.h:90:19: error: typedef redefinition with different types ('int16_t' (aka 'short') vs 'int'), err: true
  ./../lib_ccx/stdintmsc.h:94:19: error: typedef redefinition with different types ('uint16_t' (aka 'unsigned short') vs 'unsigned int'), err: true
```

stdintmsc.h does mention that it must be used only with Microsoft Visual C++ compilers. I have set a regex to only generate bindings for types with `_dtvcc_` in their name. I don't know why bindgen is trying to generate bindings for that header file

cfsmp3:
> Changing the path does fix that issue. So I think I should make a different PR for that fix?

Yes, please.

>  ./../lib_ccx/stdintmsc.h:69:27: error: typedef redefinition with different types ('char' vs 'signed char'), err: true
>   ./../lib_ccx/stdintmsc.h:90:19: error: typedef redefinition with different types ('int16_t' (aka 'short') vs 'int'), err: true
>   ./../lib_ccx/stdintmsc.h:94:19: error: typedef redefinition with different types ('uint16_t' (aka 'unsigned short') vs 'unsigned int'), err: true

Those are equivalent I think - at least in the architectures we care about. So it's a matter of tracking the difference down (where is the other definition?) and adjusting one of them.

There's some value in bindgen and/or rust giving us a hard time for our inconsistencies :-)

PunitLodha:
https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_common_platform.h#L56-L61
Both the #includes here do the same thing
disable_warnings.h -> https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/disable_warnings.h
So we should remove one

Clang also defines _MSC_VER, but we want to include <stdintmsc.h> only on Microsoft Visual C++ compilers, so we need to change this, https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_common_platform.h#L58, to `#if defined(_MSC_VER) && !defined(__clang__)`

PR #1358, opened 2021-06-30
Add functions to rust

PunitLodha:

---
Add process current packet and process service decoders functions to rust

### DIFF
--- a/src/rust/src/decoder/mod.rs
+++ b/src/rust/src/decoder/mod.rs
@@ -7,6 +7,7 @@ use log::{debug, warn};
 use crate::bindings::*;
 
 const CCX_DTVCC_MAX_PACKET_LENGTH: u8 = 128;
+const CCX_DTVCC_NO_LAST_SEQUENCE: i32 = -1;
 
 /// Process data passed from C
 ///
@@ -61,7 +62,7 @@ impl dtvcc_ctx {
                         }
 
                         if self.current_packet_length >= max_len as i32 {
-                            self.process_current_packet(max_len as i32);
+                            self.process_current_packet(max_len);
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+            "dtvcc_process_current_packet: Sequence: {}, packet length: {}",
+            seq, len
+        );
+        if self.current_packet_length == 0 {
+            return;
+        }
+
+        if self.current_packet_length != len as i32 {
+            // Is this possible?
+            self.reset_decoders();
```
cfsmp3, line 104: We should answer that question :-) I'd suppose that illegal, but I don't know. And if yes - is resetting the decoder the right way to go? 
Do we write something on the logs or just silently try to recover?

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+            return;
+        }
+
+        if self.last_sequence != CCX_DTVCC_NO_LAST_SEQUENCE
+            && (self.last_sequence + 1) % 4 != seq as i32
+        {
+            warn!("dtvcc_process_current_packet: Unexpected sequence number, it is {} but should be {}", seq, (self.last_sequence +1) % 4);
+        }
+        self.last_sequence = seq as i32;
```
cfsmp3, line 113: Are all these type conversions/declarations really needed?

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+
+        let mut pos: u8 = 1;
+        while pos < len {
+            let mut service_number = (self.current_packet[pos as usize] & 0xE0) >> 5; // 3 more significant bits
+            let block_length = self.current_packet[pos as usize] & 0x1F; // 5 less significant bits
+            debug!("dtvcc_process_current_packet: Standard header Service number: {}, Block length: {}", service_number, block_length);
+
+            if service_number == 7 {
+                // There is an extended header
```
cfsmp3, line 122: A reference to the part of the specs that explain this magic number would be good :-)

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+            if service_number == 0 && block_length != 0 {
+                // Illegal, but specs say what to do...
+                pos = len; // Move to end
+                break;
+            }
+
+            if block_length != 0 {
+                unsafe {
+                    let mut report = *self.report;
+                    report.services[service_number as usize] = 1;
```
cfsmp3, line 141: Can we avoid unsafe here?

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+                self.process_service_block(service_number - 1, pos, block_length);
+            }
+
+            pos += block_length // Skip data
+        }
+
+        self.clear_packet();
+        if pos != len {
+            // For some reason we didn't parse the whole packet
+            warn!("dtvcc_process_current_packet:  There was a problem with this packet, reseting");
```
cfsmp3, line 155: resetting

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+                } else if self.current_packet[curr] <= 0x7F {
+                    self.handle_G0(decoder, pos + i, block_length - i)
+                } else if self.current_packet[curr] <= 0x9F {
+                    self.handle_C1(decoder, pos + i, block_length - i)
+                } else {
+                    self.handle_G1(decoder, pos + i, block_length - i)
+                };
+
+                if used == -1 {
+                    warn!("dtvcc_process_service_block: There was a problem handling the data.");
```
cfsmp3, line 182: What problem?

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+        }
+    }
+    /// Process service block and call handlers for the respective codesets
+    pub fn process_service_block(&mut self, decoder: u8, pos: u8, block_length: u8) {
+        let mut i = 0;
+        while i < block_length {
+            let curr: usize = (pos + i) as usize;
+
+            let consumed = if self.current_packet[curr] != DTVCC_COMMANDS_C0_CODES_DTVCC_C0_EXT1 {
+                let used = if self.current_packet[curr] <= 0x1F {
```
cfsmp3, line 171: match instead of cascaded if would be better?

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+                used
+            } else {
+                let mut used = self.handle_extended_char(decoder, pos + i, block_length - i);
+                used += 1; // Since we had CCX_DTVCC_C0_EXT1
+                used
+            };
+            i += consumed as u8;
+        }
+    }
+    pub fn handle_C0(&mut self, decoder: u8, pos: u8, block_length: u8) -> i32 {
```
cfsmp3, line 194: What's C0? I know we might not have it in the existing implementation, but we should have some comment for each function explaining each command.

src/rust/src/decoder/mod.rs
```diff
@@ -86,10 +88,161 @@ impl dtvcc_ctx {
...
+        unsafe {
```
cfsmp3, line 195: I don't believe that each command processing function needs to be unsafe completely :-) 
Most likely this can be reduced to one single unsafe in the calling function.

cfsmp3 (changes requested):

src/rust/src/decoder/mod.rs
PunitLodha: `self.report` is a raw pointer, so we have to use unsafe here

src/rust/src/decoder/mod.rs
PunitLodha: I think I'll just remove this check. It's not needed. We will never reach this case due to this check, https://github.com/CCExtractor/ccextractor/blob/master/src/rust/src/decoder/mod.rs#L63

src/rust/src/decoder/mod.rs
PunitLodha: Yes, because the C bindings use `c_int`, which is `i32` in rust

src/rust/src/decoder/mod.rs
PunitLodha: unsafe is being used here for calling the C function. Once I port the function, unsafe won't be needed

src/rust/src/decoder/mod.rs
PunitLodha: Even this one is unreachable, I'll remove it

src/rust/src/decoder/mod.rs
PunitLodha: This comes from either the command being an unknown one. e.g. we get 0x09 in C0, which is an unknown command,
Or if there is missing data for a command. e.g. if the command has 3 bytes of data, but we have only 2.

This is an additional log. Whenever any of the above two cases occur, there will be an appropriate log entry for that

src/rust/src/decoder/mod.rs
PunitLodha: I have added a line for C0. Also, I'll make sure to add explanations when I add the functions for individual commands

PR #1359, opened 2021-07-04
refactor: consistent param names

Techno-Disaster:
Makes the parameter names a bit more consistent (--param name) 
Also contains some autogenerated changes, should I remove those?


Techno-Disaster:
Some params, mainly in hardsubx section are not done yet

cfsmp3:
Changes in the solution files don't belong in this PR.
Make sure your PR includes _only_ the files you need to merge for the specific functionality the PR tackles, not everything that is different between your file system and the upstream repo.

canihavesomecoffee:
I'd suggest to rename all stuff in the help menu too, so that people start using the newer, consistent param names. We can then remove the old ones in a couple of versions (after updating the regression tests)

Techno-Disaster:
> I'd suggest to rename all stuff in the help menu too, so that people start using the newer, consistent param names. We can then remove the old ones in a couple of versions (after updating the regression tests)

Would you suggest just a line at the top of the file stating this as a notice or do I have to change every option D:

src/lib_ccx/params.c
```diff
@@ -1495,17 +1495,17 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 			continue;
 		}
-		if (strcmp(argv[i], "-bom") == 0)
+		if (strcmp(argv[i], "-bom") == 0 || strcmp(argv[i], "--bom") == 0)
 		{
 			opt->enc_cfg.no_bom = 0;
 			continue;
 		}
-		if (strcmp(argv[i], "-nobom") == 0)
+		if (strcmp(argv[i], "-nobom") == 0 || strcmp(argv[i], "-nobom") == 0)
```
cfsmp3, line 1503: How are these two different?

src/lib_ccx/params.c
```diff
@@ -2084,22 +2123,22 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
 				fatal(EXIT_MALFORMED_PARAMETER, "-endat has no argument.\n");
 			}
 		}
-		if (strcmp(argv[i], "-1") == 0)
+		if (strcmp(argv[i], "-1") == 0 || strcmp(argv[i], "-1") == 0)
```
cfsmp3, line 2126: How are these different?

src/lib_ccx/params.c
```diff
@@ -2084,22 +2123,22 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 		{
 			opt->extract = 1;
 			continue;
 		}
-		if (strcmp(argv[i], "-2") == 0)
+		if (strcmp(argv[i], "-2") == 0 || strcmp(argv[i], "-2") == 0)
```
cfsmp3, line 2131: How are these different?

src/lib_ccx/params.c
```diff
@@ -2274,22 +2313,22 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 			continue;
 		}
-		if (strcmp(argv[i], "-mp4vidtrack") == 0)
+		if (strcmp(argv[i], "-mp4vidtrack") == 0 || strcmp(argv[i], "--mp4vidtrack") == 0)
 		{
 			opt->mp4vidtrack = 1;
 			continue;
 		}
-		if (strstr(argv[i], "-unicode") != NULL)
+		if (strstr(argv[i], "-unicode") != NULL || strstr(argv[i], "--unicode") != NULL)
```
cfsmp3, line 2321: This is not needed since we're using strstr() instead of strcmp(), so strstr() would match both.

src/lib_ccx/params.c
```diff
@@ -2274,22 +2313,22 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 		{
 			opt->enc_cfg.encoding = CCX_ENC_UNICODE;
 			continue;
 		}
-		if (strstr(argv[i], "-utf8") != NULL)
+		if (strstr(argv[i], "-utf8") != NULL || strstr(argv[i], "--utf8") != NULL)
```
cfsmp3, line 2326: This is not needed since we're using strstr() instead of strcmp(), so strstr() would match both.

src/lib_ccx/params.c
```diff
@@ -2274,22 +2313,22 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 		{
 			opt->enc_cfg.encoding = CCX_ENC_UTF_8;
 			continue;
 		}
-		if (strstr(argv[i], "-latin1") != NULL)
+		if (strstr(argv[i], "-latin1") != NULL || strstr(argv[i], "--latin1") != NULL)
```
cfsmp3, line 2331: This is not needed since we're using strstr() instead of strcmp(), so strstr() would match both.

src/lib_ccx/params.c
```diff
@@ -2299,39 +2338,26 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
...
 			opt->wtvconvertfix = 1;
 			continue;
 		}
-		if (strstr(argv[i], "-wtvmpeg2") != NULL)
+		if (strstr(argv[i], "-wtvmpeg2") != NULL || strstr(argv[i], "--wtvmpeg2") != NULL)
 		{
 			opt->wtvmpeg2 = 1;
 			continue;
 		}
-		if (strcmp(argv[i], "-o") == 0)
```
cfsmp3, line 2322: Why?

cfsmp3 (changes requested):

src/lib_ccx/params.c
Techno-Disaster: This (-o and --outputfilename) was moved to line 1702 above --out because --outputfilename was being parsed as --out.

src/lib_ccx/params.c
Techno-Disaster: Fixed

src/lib_ccx/params.c
Techno-Disaster: Fixed

src/lib_ccx/params.c
Techno-Disaster: Fixed

src/lib_ccx/params.c
Techno-Disaster: Got it

src/lib_ccx/params.c
Techno-Disaster: Got it

src/lib_ccx/params.c
Techno-Disaster: Got it

src/lib_ccx/params.c
Techno-Disaster: Got it

src/lib_ccx/params.c
```diff
@@ -352,18 +352,18 @@ void print_usage(void)
 	mprint("To see This Help Message: -h or --help\n\n");
 	mprint("File name related options:\n");
 	mprint("            inputfile: file(s) to process\n");
-	mprint("    -o outputfilename: Use -o parameters to define output filename if you don't\n");
+	mprint("    --outputfilename [outputfilename]: Use --outputfilename parameters to define output filename if you don't\n");
```
cfsmp3, line 355: This line is now too long.

src/lib_ccx/params.c
```diff
@@ -375,37 +375,37 @@ void print_usage(void)
...
 	mprint("                        local host\n");
-	mprint("            -tcppassword password: Sets server password for new connections to\n");
+	mprint("           --tcppassword password: Sets server password for new connections to\n");
 	mprint("                                   tcp server\n");
-	mprint("            -tcpdesc description: Sends to the server short description about\n");
+	mprint("           --tcpdesc description: Sends to the server short description about\n");
 	mprint("                                  captions e.g. channel name or file name\n");
 	mprint("Options that affect what will be processed:\n");
-	mprint("          -1, -2, -12: Output Field 1 data, Field 2 data, or both\n");
+	mprint("         --1, --2, --12: Output Field 1 data, Field 2 data, or both\n");
```
cfsmp3, line 403: -- is used for long args and - for short. 
-1 is short.

src/lib_ccx/params.c
```diff
@@ -422,12 +422,12 @@ void print_usage(void)
...
 	mprint("In general, if you want English subtitles you don't need to use these options\n");
 	mprint("as they are broadcast in field 1, channel 1. If you want the second language\n");
-	mprint("(usually Spanish) you may need to try -2, or -cc2, or both.\n\n");
+	mprint("(usually Spanish) you may need to try --2, or --CC2, or both.\n\n");
 	mprint("Input formats:\n");
 	mprint("       With the exception of McPoodle's raw format, which is just the closed\n");
 	mprint("       caption data with no other info, CCExtractor can usually detect the\n");
 	mprint("       input format correctly. To force a specific format:\n\n");
-	mprint("                  -in=format\n\n");
+	mprint("                 --inp [format]\n\n");
```
cfsmp3, line 430: Why inp instead of in?

src/lib_ccx/params.c
```diff
@@ -1698,6 +1724,19 @@ int parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
 			set_output_format(opt, argv[i]);
 			continue;
 		}
+		if (strncmp(argv[i], "--out", 5) == 0)
```
cfsmp3, line 1727: Why this new block? It's identical to -o so you could just add the variants to line 1702.

cfsmp3 (changes requested):

src/lib_ccx/params.c
Techno-Disaster: This is for output format (previously -out-format) so not sure what you mean by putting it with -o/--outputfilename

src/lib_ccx/params.c
Techno-Disaster: in is a keyword in [dart](https://dart.dev/guides/language/language-tour#keywords) So I can't keep the name in inside my SettingsModel in my GUI. I could handle the --in case manually on the GUI side but thought it would be easier here. Should I revert this?

src/lib_ccx/params.c
Techno-Disaster: AFAIK, there is no short/long version of 1,2,12 though. I could handle this case manually in the GUI, but then don't want the code to get messy there :P 
Can I add something like --field1 --field2 just for GUI purpose?

src/lib_ccx/params.c
Techno-Disaster:
```suggestion
	mprint(" --outputfilename [name]: Use this parameter to define output filename if you don't\n");  
```

src/lib_ccx/params.c
cfsmp3: Some args have a short version, some a long version, and some both. The GUI needs to be deal with that - just adding "--" to every arg because it's convenient for you it's not a good reason to break scripts that rely in -1 and not --1.

Just have an array of the proper form for each arg and do a look up...

src/lib_ccx/params.c
cfsmp3: Yes. No way we're going to change the C or Rust code because a totally unrelated program, in a different language, has a keyword. 

I'd say your SettingModels is trying to do directly what clearly needs a translation table...

cfsmp3 (changes requested):

src/lib_ccx/params.c
Techno-Disaster: Ok I have reverted this back to `--in format`

src/lib_ccx/params.c
Techno-Disaster: `if (strcmp(argv[i], "-1") == 0 || strcmp(argv[i], "--1") == 0)` Won't break any scripts though

Techno-Disaster:
@cfsmp3 I have reverted --1 and --inp to -1 and `--in format` now, but as mentioned above `if (strcmp(argv[i], "-1") == 0 || strcmp(argv[i], "--1") == 0)`  won't break any scripts, so can I add that? If not please let's merge it :D  I'll look into this on the GUI side.

Issue #726, opened 2017-03-22
[PROPOSAL] Process closed captions and burned-in subtitles in one pass


cfsmp3:
There's some subs that have both closed captions and burned-in subtitles. This happens typically when part of the audio is in English (which is closed captions so the viewers can enable/disable them) while part of the audio is in a different language, which is subtitled for everyone. 

A very good example is the Americans, in which a good portion of the audio is in Russian. 

**This task gives ENOUGH points to qualify for GSoC. If you're proposal is good and your code to get this done gets merged, you're in.**

cfsmp3:
A sample video:

https://drive.google.com/open?id=0B_61ywKPmI0TeU5rOTlCMWxKbW8

saurabhshah0410:
Is this bug still open?

cfsmp3:
Yes.

thealphadollar:
@cfsmp3 I would like to work on this. Can you please give me some tips for the solution?

cfsmp3:
@thealphadollar First download the sample file we have for it, extract the subtitles in two passes (one for burned-in and one for closed captions), and then figure out how to do everything in one pass :-)

saurabhshah0410:
@cfsmp3 After running ccextractor once on files like these, and having extracted the closed captions as well as hard subtitles, can you clarify if the result should be a single file with both closed captions and hard subtitles, or two separate files which contain the closed captions and hard subtitles respectively?

cfsmp3:
The result needs to be a single file that contains both kind of subtitles.

saurabhshah0410:
Thanks for clarifying this @cfsmp3 . :-)

saurabhshah0410:
I think we should add a new option, something like `hcc` (hard and closed captions) or whatever name you suggest, which when given would process hard subs and closed captions in one pass. What do you think?

cfsmp3:
Yes, it's a good idea - we need the user to specify that he wants to
process both things, we cannot be looking for hard subs by default.

saurabhshah0410:
Thanks, just wanted to be sure about it. Just to play the devil's advocate: If a user gives both the switches: `-hardsubx` and `-hcc`, should we process both hard subs and closed captions? Basically, does `-hcc` takes precedence over anything else(like `-hardsubx`)?

cfsmp3:
If -hcc means both then do both :-)
I don't think -hardsubx and -hcc would be mutually exclusive since they're
not incompatible in their meaning.

[...]

PR #1422, opened 2022-03-14
add option to extract closed captions and burnt in subs in the same pass

shashwat1002:

- [] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**
(Not applicable because still WIP)

---

This PR is in context to issue #726 

The current status is that the generated subtitle file has content from closed-caption extraction and burnt-in extraction, however they are not in the correct order. 

The approach I have in mind right now is to use the timestamp of previously encoded subtitles of one of the modes (burnt-in or closed caption) and make a decision on which packet to process (the image or the closed caption) 

Please advise on whether this approach looks okay. 

Note: I have added appropriate command line options for this already

Any advice is appreciated 

CC: @cfsmp3

### DIFF
--- a/src/lib_ccx/hardsubx_decoder.c
+++ b/src/lib_ccx/hardsubx_decoder.c
@@ -439,7 +439,7 @@ void hardsubx_process_frames_linear(struct lib_hardsubx_ctx *ctx, struct encoder
 			if (got_frame && frame_number % 25 == 0)
 			{
 				float diff = (float)convert_pts_to_ms(ctx->packet.pts - prev_packet_pts, ctx->format_ctx->streams[ctx->video_stream_id]->time_base);
-				if (fabsf(diff) < 1000 * ctx->min_sub_duration) //If the minimum duration of a subtitle line is exceeded, process packet
+				if (fabsf(diff) < 1000 * ctx->min_sub_duration) // If the minimum duration of a subtitle line is exceeded, process packet
 					continue;
 
 				// sws_scale is used to convert the pixel format to RGB24 from all other cases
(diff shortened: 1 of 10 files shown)


cfsmp3:
I don't think duplicating code is the way to go (that seems to be what's happening here). Instead I think if -hcc is used then rather than just executing one of the branches you execute two (which will require some refactoring).

That should be give you in-order subtitles, since in both cases you write to the output once you the subtitles disappear (because that's when you have both the in and out timestamps).

BTW there's a chance there's a collision in timestamps. For example, the burned-in could say the text while the closed captions could say "(speak in Russian)". For now I think we can live with that collision until we see real world examples and figure out the most useful way to solve it, but I just wanted to point it out.

shashwat1002:
@cfsmp3  

Would it be more acceptable to put the duplicated code in functions and call them appropriately in the original functions and the `-hcc` function? 

I assume that's what you mean by refactoring. 

since the both branches use their own separate loops, I have to get inside methods that have the central loops of both branches. Therefore, to get identical effect for both the branches I should probably use the code that's already there. I realize it'll be better to make appropriate functions to put the code in. 

Am I on the right track?

cfsmp3:
> Would it be more acceptable to put the duplicated code in functions 

In general smaller functions are a good idea. We have functions that are huge.

> I assume that's what you mean by refactoring.

What I mean is that instead of having a switch() that does one of the two things, we may need something else that does either one of the two things or both things.

In any case, code duplication needs to be avoided.

shashwat1002:
> What I mean is that instead of having a switch() that does one of the two things, we may need something else that does either one of the two things or both things.

I don't think this can be done without moving some code around as `hardsubx` behaves like a completely independent module and the choice between burnt-in and cc is made very early on. 

I will try to introduce a few functions to avoid too much duplication and deal with the time appropriately. 

What I'm currently working is to turn the code for iterations into methods so that they can be called appropriately (both burned-in and cc) for each iteration

@cfsmp3

src/lib_ccx/hardsubx_decoder.c
```diff
@@ -544,6 +544,352 @@ void hardsubx_process_frames_linear(struct lib_hardsubx_ctx *ctx, struct encoder
 	activity_progress(100, cur_sec / 60, cur_sec % 60);
 }
 
+void process_hardsubx_linear_frames_and_normal_subs(struct lib_hardsubx_ctx *hard_ctx, struct encoder_ctx *enc_ctx, struct lib_ccx_ctx *ctx)
```
cfsmp3, line 547: Does this function just have a lot of copy pasted code?

shashwat1002: Yeah, not exactly duplication but moving it to that method.

That piece of code has been used several places with little or no modification. including the `else` branch of the `!ctx->multiprogram` and since I needed the functionality in this feature as well, I figured that moving the code to the function will reduce some duplication overall and I will call the function appropriately in all the places where the code shows up. 

The code is already duplicated in the existing codebase, as a side effect of working on this issue I hope to generalize the function and call it appropriately.

shashwat1002: @cfsmp3

shashwat1002:
@cfsmp3 

Update: The feature works as expected. The generated file contains both the burnt-in subs and the closed captions in appropriate places with the appropriate timestamps and counter.

For now, the code does not deal overlaps extremely well (in case of overlap, it will put the subtitle that started earlier before the other) 

Note that I haven't yet gotten the progress bar to work properly

But I felt that this would be a good place for a review.

cfsmp3:
It seems reasonable to me. I'd change the variable names though, they're too non-descriptive. 
Also, fix formatting.
And of course, the PR needs to pass the test suite - @canihavesomecoffee can this be enabled for @shashwat1002 's PRs?

shashwat1002:
@cfsmp3 
Please have a look now.

canihavesomecoffee (changes requested):
Not all tests are passing. See https://sampleplatform.ccextractor.org/test/3664 and my comment on Slack.

canihavesomecoffee (approved):

canihavesomecoffee:
@cfsmp3 Fine for me. No new breakages compared to master status.

shashwat1002:
Thank you @canihavesomecoffee 

I will squash the commits and add a change log by end of day (Indian standard time)

CC: @cfsmp3

shashwat1002:
Done @cfsmp3 @canihavesomecoffee 

Note that I had to run clang formatter on the files I changed for the formatting test and that changed some (some comments) in code that has nothing to do with the PR. To make things clearer, I have put all changes made by the clang formatter in the last commit. 

Please let me know if there's anything else for this PR :D

PR #1439, opened 2022-06-06
Port hardsubx imgops

shashwat1002:

---

On determination that the imgops functions have lowest dependency and need to be used as-is, that is where I have begun porting. 

The PR introduces (as of now) a function with an identical signature to `rgb_to_hsv` with rest of the code unchanged. The  `DISABLE_RUST` flag is used to decide if the Rust version is compiled or the C version. 
Edit: hardsubx_imgops is now fully ported

The compilation works and the executable runs properly on the files I have. However, the who suite of tests must be run as soon as the `imgops` module is ported. 

Ps: I discovered that when building using `cmake`, Rust compilation is disabled by default and I had to explicitly set the `DWITHOUT_RUST=OFF` 
I am unsure if this is the expected behaviour. 
Edit: apparently this is because cmake caches options

Cc: @PunitLodha, @cfsmp3  
please look this over once to check if I am not doing something obviously dangerous. I will continue after feedback.


src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,128 @@
...
+        f32::max($a, f32::max($b, $c))
+    };
+}
+
+macro_rules! min_f {
+    ($a: ident, $b: ident, $c:ident) =>{
+        f32::min($a, f32::min($b, $c))
+    };
+}
+
```
PunitLodha, line 14: I don't think macros are needed for a statement that is just used once

src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,128 @@
...
+    () => {
+        20.0
+    };
+}
+
+macro_rules! YELLOW {
+    () => {
+        70.0
+    };
+}
```
PunitLodha, line 25: These should be constants and not macros

src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,128 @@
...
+{
+    let r: f32 = R / 255.0;
+    let g: f32 = G / 255.0;
+    let b: f32 = B / 255.0;
+    let max: f32 = max_f!(r, g, b);
+    let min: f32 = min_f!(r, g, b);
+
+    let mut h: f32;
+    let s: f32;
+    let v: f32 = max;
```
PunitLodha, line 42: No need of `:f32` here. Rust will interpret the types automatically

src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,128 @@
...
+        {
+            h = 60.0 * ((g - b) / (max - min)) + 0.0;
+        }
+        else if max == g {
+            h = 60.0 * ((b - r) / (max - min)) + 120.0;
+        }
+        else {
+            h = 60.0 * ((r - g) / (max - min)) + 240.0;
+        }
+    }
```
PunitLodha, line 62: You could use match here. You can also assign variables directly from values returned from match/if-else, to avoid the variable initialization at the top

src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,128 @@
...
+pub extern "C" fn rgb_to_lab(
+    R: f32, G: f32, B: f32,
+    L: &mut f32, a: &mut f32, b: &mut f32
+)
+{
+
+    let X: f32 = (0.412453 * R + 0.357580 * G + 0.180423 * B) / (255.0 * 0.950456);
+    let Y: f32 = (0.212671 * R + 0.715160 * G + 0.072169 * B) / 255.0;
+    let Z: f32 = (0.019334 * R + 0.119193 * G + 0.950227 * B) / (255.0 * 1.08874);
+
```
PunitLodha, line 85: There are many [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) in this function. You should add some comments to explain what is going on here, and also add documentation for the functions

shashwat1002:
Please check now @PunitLodha 

Also note that, the code now is mathematically different for the `rgb_to_lab` function 
I was looking into the actual algorithm so as to write better comments, I noticed that the existing disagreed with several references I found. I may be missing something, but I am not sure

Source:
- https://en.wikipedia.org/wiki/CIELAB_color_space#From_CIEXYZ_to_CIELAB
- http://cs.haifa.ac.il/hagit/courses/ist/Lectures/Demos/ColorApplet2/t_convert.html

shashwat1002:
@PunitLodha 
Rewrote functions using the `palette` crate

The result is definitely _different_ from the original C code 
There is still a lot of hallucination (both the old and the new version)

I believe the new version is leading to slightly less hallucinations when it comes to lines which sort of make sense in the result. I don't have a good metric as of now.  

Ps: Also it seems the rust compiler

src/rust/src/hardsubx/imgops.rs
```diff
@@ -0,0 +1,24 @@
...
+    *S = hsv_rep.saturation;
+    *V = hsv_rep.value;
+}
+
+#[no_mangle]
+pub extern "C" fn rgb_to_lab(R: f32, G: f32, B: f32, L: &mut f32, a: &mut f32, b: &mut f32) {
+    let rgb = Srgb::new(R, G, B);
+
+    let lab_rep = Lab::<palette::white_point::D65, f32>::from_color(rgb);
+    // The above declaration sets the white-point as per the D65 standard
```
PunitLodha, line 19: Small nit, but I think it's better to have any comment above the line it references
```suggestion
    // This declaration sets the white-point as per the D65 standard
    let lab_rep = Lab::<palette::white_point::D65, f32>::from_color(rgb);
```

src/rust/src/lib.rs
```diff
@@ -7,12 +7,15 @@
 #![allow(non_camel_case_types)]
 #![allow(non_snake_case)]
 
+extern crate palette;
```
PunitLodha, line 10: No need of extern crate in Rust 2018, https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate

PunitLodha:
> The result is definitely different from the original C code

This is expected. Once we update the preprocessing code, we should be seeing better results than before

> Ps: Also it seems the rust compiler does not support deconstruction on left side in assignment statements because it worked on my system but not in the checks, so I just converted and committed

This is because the tests use rust 1.56.0, and destructuring was added in 1.59.0

PunitLodha (commented):
Please fix the formatting, https://github.com/CCExtractor/ccextractor/runs/6866810135?check_suite_focus=true

Always make sure that the CI checks are not failing

Issue #1371, opened 2021-08-06
[BUG] Odd output when inputting an MKV file


Southpaw1496:
# CCExtractor version: 
```
CCExtractor detailed version info
	Version: 0.91
	Git commit: Unknown
	Compilation date: 2021-07-26
	File SHA256: Could not open file
Libraries used by CCExtractor
	Tesseract Version: 4.1.1
	Leptonica Version: leptonica-1.81.1
	libGPAC Version: 1.0.1
	zlib: 1.2.11
	utf8proc Version: 2.6.1
	protobuf-c Version: 1.4.0
	libpng Version: 1.6.37
	FreeType 
	libhash
	nuklear
	libzvbi
```

# Necessary information

- Is this a regression (i.e. did it work before)? No
- What platform did you use? Mac (with Homebrew)
- What were the used arguments? None

# Video links

* https://drive.google.com/file/d/1B9ZeUZ-Pv1dUeu0aJP5G1s_8-_SzA_xU/view?usp=sharing (ZIP encrypted, password shared with relevent people).

# Additional information

MKV was ripped from a disk using MakeMKV. I've included the output of running `ccextreactor [filename]` in the below `output.zip` file.
[Output.zip](https://github.com/CCExtractor/ccextractor/files/6945015/Output.zip)
Please let me know if any more information is required.

Southpaw1496:
Here is another file of a Bugs Bunny short, unencrypted this time https://drive.google.com/file/d/1B9ZeUZ-Pv1dUeu0aJP5G1s_8-_SzA_xU/view?usp=sharing

Output for this file:
[Output-bugs.zip](https://github.com/CCExtractor/ccextractor/files/7801404/Output-bugs.zip)

canihavesomecoffee:
Bugs file has been made available here: https://sampleplatform.ccextractor.org/sample/178

Logs from output zip above:
```
CCExtractor 0.94, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
Input: bugs.mkv
[Extract: 1] [Stream mode: Autodetect]
[Program : Auto ] [Hauppage mode: No] [Use MythTV code: Auto]
[CEA-708: 63 decoders active]
[CEA-708: using charset "none" for all services]
[Timing mode: Auto] [Debug: No] [Buffer input: No]
[...]

PR #1444, opened 2022-07-04
Fix file extension for IDX files

PunitLodha:
Fix file extension for `idx` and add VOBSUB unsupported message
#1371

### DIFF
--- a/src/lib_ccx/matroska.h
+++ b/src/lib_ccx/matroska.h
@@ -172,7 +172,7 @@ char* matroska_track_text_subtitle_id_strings[] = {
 char* matroska_track_text_subtitle_id_extensions[] = {
         "srt", "ssa", "ass",
         "usf", "vtt", "bmp",
-        NULL, NULL  // Unknown
+        "idx", NULL  // Unknown
 };
 
 char* avc_codec_id = "V_MPEG4/ISO/AVC";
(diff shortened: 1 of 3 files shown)


src/lib_ccx/matroska.h
```diff
@@ -172,7 +172,7 @@ char* matroska_track_text_subtitle_id_strings[] = {
 char* matroska_track_text_subtitle_id_extensions[] = {
         "srt", "ssa", "ass",
         "usf", "vtt", "bmp",
-        NULL, NULL  // Unknown
+        "idx", NULL  // Unknown
 };
```
cfsmp3, line 177: Are yo sure replacing that NULL doesn't break things? I don't know why we had two, maybe it's some marker...(I haven't looked into the code).

PunitLodha: > Are yo sure replacing that NULL doesn't break things?

It's just used once, to write the filename with its extension.

>I don't know why we had two, maybe it's some marker...(I haven't looked into the code).

It is a mapping of `codec_id` to their file extensions. Last 2 codecs are `VOBSUB` and `KATE`. I guess their extensions were previously unknown, hence kept as null.

cfsmp3 (approved):

PR #1446, opened 2022-07-27
[WIP] Port hardsubx classifier

shashwat1002:
Starting port of the classifier module.

### DIFF
--- a/src/rust/src/hardsubx/classifier.rs
+++ b/src/rust/src/hardsubx/classifier.rs
@@ -70,7 +70,6 @@ pub unsafe extern "C" fn get_ocr_text_simple(
     ctx: *mut lib_hardsubx_ctx,
     image: *mut Pix,
 ) -> *mut ::std::os::raw::c_char {
-
     get_ocr_text_simple_threshold(ctx, image, 0.0)
 }
 
--- a/src/rust/src/hardsubx/classifier.rs
+++ b/src/rust/src/hardsubx/classifier.rs
@@ -202,3 +202,40 @@ pub unsafe extern "C" fn get_ocr_text_wordwise(
     // TODO: this is a memory leak that can only be plugged when the caller functions have been ported
     return string_to_c_char(&text_out);
 }
+
+pub unsafe extern "C" fn get_ocr_text_letterwise(
+    ctx: *mut lib_hardsubx_ctx,
+    image: *mut PIX,
+) -> *mut ::std::os::raw::c_char {
+
+    let mut text_out: String = String::new();
+
+    TessBaseAPISetImage2((*ctx).tess_handle, image);
+
+    if TessBaseAPIRecognize((*ctx).tess_handle, null::<ETEXT_DESC>() as *mut ETEXT_DESC) != 0
+    {
+        warn!("Error in Tesseract recognition, skipping symbol\n");
+        return null::<c_char>() as *mut c_char;
+    }
+
+    let it: *mut TessResultIterator = TessBaseAPIGetIterator((*ctx).tess_handle);
+    let level: TessPageIteratorLevel = TessPageIteratorLevel_RIL_SYMBOL;
+
+    if it != null::<TessResultIterator>() as *mut TessResultIterator
+    {
+        loop {
+
+            let letter = _tess_string_helper(it, level);
+            text_out = format!("{}{}", text_out, letter);
+
+            if TessPageIteratorNext(it as *mut TessPageIterator, level) == 0 {
+                break;
+            }
+        }
+    }
+
+    TessResultIteratorDelete(it);
+
+    // TODO: this is a memory leak that can only be plugged when the caller functions have been ported
+    return string_to_c_char(&text_out);
+}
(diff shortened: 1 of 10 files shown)


shashwat1002:
@PunitLodha awaiting feedback before I continue here because I have taken certain liberties. 

The `get_ocr_text_wordwise` has a memory leak because it gives ownership of a Rust originating string to the C caller. According to documentation of `CString` it has to be deallocated using `from_raw`. That can only be done when the callers above have been ported. Therefore, I'm leaving that in for now. 
I might have to do it for the other functions in this module as well. Therefore, please take a look before I continue.

src/rust/src/hardsubx/classifier.rs
```diff
@@ -0,0 +1,202 @@
...
+        return String::new();
+    }
+
+    let ts_word = ffi::CStr::from_ptr(ts_word_ptr);
+    let ts_word_arr = ffi::CStr::to_bytes_with_nul(&ts_word);
+
+    let ts_word_string: String = match String::from_utf8(ts_word_arr.to_vec()) {
+        Ok(string_rep) => string_rep,
+        Err(error) => std::panic::panic_any(error),
+    };
```
PunitLodha, line 128: Can this be changed to,
```rust
unsafe {
        CStr::from_ptr(ts_word_ptr).to_string_lossy().into_owned()
    }
```

PunitLodha:
Logic seems fine. You can continue with the work.
There are a few unwraps/panics and formatting issues, which can be reviewed later

src/rust/src/hardsubx/classifier.rs
shashwat1002: The null check has to be there because it will fail otherwise (I have tested that eventuality) 
Otherwise yes

I will make the change.

shashwat1002:
@PunitLodha 

The classifier's porting is done. Please check


Note the currently I'm using the binding definition for `AVPacket` as opposed to `ffmpeg_sys_next`'s
I realised that it'll be better to make that change after this is merged.

src/rust/src/hardsubx/classifier.rs
```diff
@@ -0,0 +1,280 @@
...
+
+use std::os::raw::c_char;
+
+use log::warn;
+
+/// # Safety
+/// The function accepts and dereferences a raw pointer
+/// The function also makes calls to functions whose safety is not guaranteed
+/// The function returns a raw pointer which is a string made in C
+#[no_mangle]
```
PunitLodha, line 26: For the `Safety` section, you need to mention how the caller of this function can ensure that calling this function will be safe. 
Take a look at the example here, https://rust-lang.github.io/api-guidelines/documentation.html#function-docs-include-error-panic-and-safety-considerations-c-failure

src/rust/src/hardsubx/classifier.rs
```diff
@@ -0,0 +1,280 @@
...
+pub unsafe extern "C" fn get_ocr_text_simple(
+    ctx: *mut lib_hardsubx_ctx,
+    image: *mut Pix,
+) -> *mut ::std::os::raw::c_char {
+    // basically the get_oct_text_simple function without threshold
+    // This function is being kept only for backwards compatibility reasons
```
PunitLodha, line 70: You could change this to a function documentation, instead of just a comment

src/rust/src/hardsubx/decoder.rs
```diff
@@ -0,0 +1,334 @@
...
+// use crate::bindings::{hardsubx_ocr_mode_HARDSUBX_OCRMODE_WORD};
+use crate::bindings::AVFrame;
+use crate::hardsubx::classifier::*;
+use crate::hardsubx::imgops::{rgb_to_hsv, rgb_to_lab};
+use crate::hardsubx::lib_hardsubx_ctx;
+use crate::utils::string_to_c_char;
+use std::convert::TryInto;
+use std::ffi;
+use std::os::raw::c_char;
+use std::ptr::null;
```
PunitLodha, line 21: You should group the imports in all the files. I think there is some setting in rust analyzer that helps with it

src/rust/src/hardsubx/decoder.rs
```diff
@@ -0,0 +1,334 @@
...
+// TODO: turn into an enum definition when the hardsubx context is rewritten
+// static HARDSUBX_OCRMODE_FRAME: i32 = 0;
+static HARDSUBX_OCRMODE_WORD: i32 = 1;
+// static HARDSUBX_OCRMODE_LETTER: i32 = 2;
+
+// enum hardsubx_ocr_mode {
+//     HARDSUBX_OCRMODE_FRAME,
+//     HARDSUBX_OCRMODE_WORD,
+//     HARDSUBX_OCRMODE_LETTER
+// };
```
PunitLodha, line 34: Will this be fixed later?

PunitLodha (changes requested):

src/rust/src/hardsubx/decoder.rs
shashwat1002: yeah

PunitLodha (approved):

PR #1454, opened 2022-09-26
[WIP] attempt compatibility with ffmpeg

shashwat1002:
This is the work for compatibility with ffmpeg. The current problem is that certain hardsubx functions on the C side are not discoverable by the linker when Rust code tries to call them.

I think eventually I will have to close this PR and PR #1446  and have the common changes in a third PR that'll be easier than dealing with the merge conflicts.

Because this has taken far too much time, I'll roll back to ffmpeg 4 for now and continue porting. I really wanted to deal with this before. 

PS: Pardon the impolite commit language. I plan to clean up the commit history 
cc: @PunitLodha @cfsmp3

### DIFF
--- a/src/lib_ccx/hardsubx.c
+++ b/src/lib_ccx/hardsubx.c
@@ -10,6 +10,7 @@
 #include <libavutil/imgutils.h>
 #include <libswscale/swscale.h>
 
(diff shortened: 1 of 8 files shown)


shashwat1002:
This is redundant now

Issue #1453, opened 2022-09-09
[BUG] stdout argument not respected when multiple CC track found.


Adityashaw:
```
ccextractor --version
CCExtractor 0.94, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
CCExtractor detailed version info
	Version: 0.94
	Git commit: Unknown
	Compilation date: 2022-09-06
	CEA-708 decoder: Rust
	File SHA256: Could not open file
Libraries used by CCExtractor
	libGPAC Version: 1.0.1
	zlib: 1.2.11
	utf8proc Version: 2.4.0
	protobuf-c Version: 1.3.1
	libpng Version: 1.6.37
	FreeType 
	libhash
	nuklear
	libzvbi
```

# In raising this issue, I confirm the following:

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [ ] I have used the latest available version of CCExtractor to verify this issue exists.
- [ ] I have ticked all the boxes in this section and to prove it I'm deleting the section completely to remove boilerplate text.

# Necessary information

- Is this a regression (i.e. did it work before)? NO
- What platform did you use? Linux (`Linux hp 5.15.61-1-lts #1 SMP Wed, 17 Aug 2022 15:11:32 +0000 x86_64 GNU/Linux
`)
- What were the used arguments? `ccextractor -stdout -quiet input.mkv`

PS: I downloaded and renamed below file.

# Video links

*  `https://www.libde265.org/hevc-bitstreams/tos-4096x1720-tiles.mkv`

# Additional information

*Expected*: Output on stdout/terminal
*Actual*: Output on multiple srt files

How can I get all/one CC output on stdout instead of creating multiple files?

abhi-kr-2100:
I'm working on this.

[...]

PR #1460, opened 2022-11-24
[FIX] Fix issue #1453: Respect `-stdout` if multiple CC tracks are found in a Matroska input file

abhi-kr-2100:

---

When passed the `-stdout` flag, CCExtractor should write the subtitles to standard output, instead of an output file.

However, as noted in Issue #1453, CCExtractor doesn't respect the `-stdout` flag when multiple CC tracks are present in a Matroska input file (usually .mkv).

The commit/s in this pull request ensure/s that output is written to standard output if `-stdout` is present even if the input file is a Matroska container with multiple CC tracks.

Signed-off-by: Abhishek Kumar <abhi.kr.2100@gmail.com>

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,6 +1,6 @@
 0.95 (to be released)
 -----------------
-- Fix: respect `-stdout` even if multiple CC tracks are present in a Matroksa input file (#1453)
+- Fix: respect `-stdout` even if multiple CC tracks are present in a Matroska input file (#1453)
 - Fix: crash in Rust decoder on ATSC1.0 TS Files (#1407)
 - Removed the --with-gui flag for linux/configure and mac/configure (use the Flutter GUI instead)
 - Fix: segmentation fault in using hardsubx
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 0.95 (to be released)
 -----------------
+- Fix: respect `-stdout` even if multiple CC tracks are present in a Matroksa input file (#1453)
 - Fix: crash in Rust decoder on ATSC1.0 TS Files (#1407)
 - Removed the --with-gui flag for linux/configure and mac/configure (use the Flutter GUI instead)
 - Fix: segmentation fault in using hardsubx
--- a/src/lib_ccx/matroska.c
+++ b/src/lib_ccx/matroska.c
@@ -1099,15 +1099,24 @@ char *ass_ssa_sentence_erase_read_order(char *text)
 
 void save_sub_track(struct matroska_ctx *mkv_ctx, struct matroska_sub_track *track)
 {
-	char *filename = generate_filename_from_track(mkv_ctx, track);
-	mprint("\nOutput file: %s", filename);
+	char *filename;
 	int desc;
+
+	if (mkv_ctx->ctx->cc_to_stdout == CCX_TRUE)
+	{
+		desc = 1; // file descriptor of stdout
+	}
+	else
+	{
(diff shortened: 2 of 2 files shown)


abhi-kr-2100:
Build CCExtractor on Windows CI has failed. From the details:

```
EXEC : error : failed to run custom build command for `leptonica-sys v0.4.1` [D:\a\ccextractor\ccextractor\windows\ccextractor.vcxproj]
```

It appears to be a problem with the CI as the only file this PR changes is `matroska.c`.

docs/CHANGES.TXT
```diff
@@ -1,5 +1,6 @@
 0.95 (to be released)
 -----------------
+- Fix: respect `-stdout` even if multiple CC tracks are present in a Matroksa input file (#1453)
```
cfsmp3, line 3: Typo in Matroska

src/lib_ccx/matroska.c
```diff
@@ -1099,15 +1099,24 @@ char *ass_ssa_sentence_erase_read_order(char *text)
 
 void save_sub_track(struct matroska_ctx *mkv_ctx, struct matroska_sub_track *track)
 {
-	char *filename = generate_filename_from_track(mkv_ctx, track);
-	mprint("\nOutput file: %s", filename);
```
cfsmp3, line 1103: Don't think removing it in all cases even when not using stdout is a good idea

src/lib_ccx/matroska.c
```diff
@@ -1099,15 +1099,24 @@ char *ass_ssa_sentence_erase_read_order(char *text)
...
+	{
+		desc = 1; // file descriptor of stdout
+	}
+	else
+	{
+		filename = generate_filename_from_track(mkv_ctx, track);
+		mprint("\nOutput file: %s", filename);
 #ifdef WIN32
-	desc = open(filename, O_WRONLY | O_CREAT | O_TRUNC | O_APPEND, S_IREAD | S_IWRITE);
+		desc = open(filename, O_WRONLY | O_CREAT | O_TRUNC | O_APPEND, S_IREAD | S_IWRITE);
```
cfsmp3, line 1114: Tabs instead of spaces there

cfsmp3 (changes requested):

src/lib_ccx/matroska.c
abhi-kr-2100: I'm using an `if` statement to check if CCExtractor has been asked to output to standard output. Those two lines are present in the `else` block: [line #1110](https://github.com/CCExtractor/ccextractor/pull/1460/files/1f0eb852b269a01e5cc374ef26db060d42f9b7d3#diff-a404e3db34ed3c7c409c47344a2aa2668179b6004ce9857b6e625f54ec3d6c67R1111).

src/lib_ccx/matroska.c
abhi-kr-2100: I've indeed used a tab there. Here's `vim`'s result of running the `:ascii` command at that character:
image
You can also see in the bottom right that `vim` considers the cursor to be at line 1114, character 2, and display width 16 (each tab occupying 8 screen characters): `1114,2-16`.

docs/CHANGES.TXT
abhi-kr-2100: Fixed in 7f91df5403ff4fcf58461ae0d9ef559f10ea6c6e. Thanks!

src/lib_ccx/matroska.c
abhi-kr-2100: I might have misunderstood you.

Do you mean I should use spaces instead of tabs? The rest of the file (untouched by me) uses tabs.

Issue #1463, opened 2022-12-13
[BUG] Incorrect placement of X-TIMESTAMP-MAP in WebVTT


bbgdzxng1:
### Summary

When ccextractor is generating webVTT output, `X-TIMESTAMP-MAP` does not immediately follow the `WEBVTT` header.

tl;dr: Propose that https://github.com/emkman99/ccextractor/commit/12b9f939fe55b34d88b54ec83a09013fad6ee62c is reverted.

#### Reference

According to Roger Pantos, the author of the HLS RFC...

_"To be clear, what HLS expects (and what the VTT spec defined prior to that 2016 change) is for the `X-TIMESTAMP-MAP` line to be among a set of non-blank lines immediately after the `WEBVTT` header line, followed by two or more line terminators, followed by the rest of the body."_

Here is Roger's full statement, clarifying the expected behavior https://mailarchive.ietf.org/arch/msg/hls-interest/4vmLpEsV-EnmkEwMQZkzbGQai_4/ clarifying the background around the .  Roger is the authoritative reference on [HLS RFC8216](https://www.rfc-editor.org/rfc/rfc8216).

See also:
https://github.com/w3c/webvtt.js/issues/38
https://github.com/w3c/webvtt/issues/485

#### Expected Behavior

Note that there is no blank line after the WEBVTT statement.

```webvtt
WEBVTT
X-TIMESTAMP-MAP=MPEGTS:5785169281,LOCAL:00:00:00.000

00:00:06.640 --> 00:00:08.307 line:79.33%
    GEICO has a long history    

00:00:06.640 --> 00:00:08.307 line:84.66%
        of great savings        

00:00:08.342 --> 00:00:09.208 line:84.66%
       and great service.
```

#### Current Behavior

```webvtt
WEBVTT

X-TIMESTAMP-MAP=MPEGTS:5785169281,LOCAL:00:00:00.000

00:00:06.640 --> 00:00:08.307 line:79.33%
    GEICO has a long history    

00:00:06.640 --> 00:00:08.307 line:84.66%
        of great savings        

00:00:08.342 --> 00:00:09.208 line:84.66%
       and great service.
```

Command to replicate...

```bash
$ ccextractor "./CNN.ts" -in='ts' -1 -out='vtt' -stdout | head -n10
```
Where CNN.ts is taken from CNN.ts from _"US TV recordings, 10 minutes samples, HDHomeRun"_ located at https://ccextractor.org/public/general/tvsamples/.

[...]

PR #1464, opened 2022-12-13
[FIX] WebVTT X-TIMESTAMP-MAP header placement (#1463)

emkman99:

---

This fixes issue #1463. If enabled, the X-TIMESTAMP-MAP header is placed on the second line. There are always two newlines between the headers and body.

This also fixes the --no-timestamp-map option, which wasn't disabling the header.

I added on the commit to disable X-TIMESTAMP-MAP header by default since it is not valid WebVTT. It doesn't make sense to generate invalid WebVTT by default, especially when the input is not MPEG-TS. This is a breaking change (--no-timestamp-map becomes --timestamp-map).

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -5,6 +5,7 @@
 - Fix: segmentation fault in using hardsubx
 - New: Add function (and command) that extracts closed caption subtitles as well as burnt-in subtitles from a file in a single pass. (As proposed in issue 726)
 - Refactored: the `general_loop` function has some code moved to a new function
+- Fix: WebVTT X-TIMESTAMP-MAP placement (#1463)
 
 
 0.94 (2021-12-14)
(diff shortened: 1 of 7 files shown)


bbgdzxng1:
@emkman99 Even though it _is_ a breaking change, your solution to always generate valid W3C webVTT by default but still allow the user to optionally choose HLS-compliant webVTT with `--timestamp_map` is the correct technical solution.

The default output of ccextractor will now pass the W3C webVTT validator, but still allow us HLS-folks to have the tags that are needed by some hardware players.

Thanks for your work on this.

cfsmp3:
@emkman99 can you fix formatting

emkman99:
> @emkman99 can you fix formatting

Perhaps I'm reading the error wrong, but the format failure appears to be caused by `src/GUI/ccextractorGUI.c`, which I did not modify.

canihavesomecoffee:
@cfsmp3 @emkman99  There's errors in multiple files, including ones that are not touched by this PR.

@cfsmp3 I'd suggest (if formatting is the only remark you have) to merge, and then just apply the formatting tool in a single commit to make the check pass again... And make sure new PR's don't break it this time.

cfsmp3:
> @cfsmp3 @emkman99 There's errors in multiple files, including ones that are not touched by this PR.
> 
> @cfsmp3 I'd suggest (if formatting is the only remark you have) to merge, and then just apply the formatting tool in a single commit to make the check pass again... And make sure new PR's don't break it this time.

OK, merged.
Can you send the format fix PR to get this out of the way?

PR #1480, opened 2023-02-16
[IMPROVEMENT] Update Rust and fix windows build

prateekmedia:

---

- Update Rust to stable (done)
- Switch to `rusty_ffmpeg` and `rsmpeg` from `ffmpeg-sys-next`
- Fix windows CI (add support of vcpkg for static linking and supporting both architectures)


prateekmedia:
The test results doesn't look good, 0 passed out of them.

cfsmp3:
> The test results doesn't look good, 0 passed out of them.

I suspect the CI, but @canihavesomecoffee will have to confirm what's going on :-)

prateekmedia:
I checked the ccextractorwinfull.exe and It couldn't find the eng.traineddata filed for hardsubx, I see that vcpkg doesn't include that by default so we would have to copy that in the PostBuildScript of vccxproj. But the problem is that I am not sure if It will be able to find the eng.traineddata path even after bundling it in the script at some location.

cfsmp3:
> I checked the ccextractorwinfull.exe and It couldn't find the eng.traineddata file for hardsubx, I see that vcpkg doesn't include that by default so we would have to copy that in the PostBuildScript of vccxproj. But the problem is that I am not sure if It will be able to find the eng.traineddata path even after bundling it in the exe at some location.

We should be able to just wget it from the github action itself  ( https://github.com/tesseract-ocr/tessdata/raw/main/eng.traineddata )

And then use it.

PR #1489, opened 2023-03-08
[IMPROVEMENT] adding the release flag to `cargo build` in linux/build

ArchitBhonsle:

---

I noticed that the Linux `build` script did not compile Rust with the release flag. I haven't mentioned this in the changelog because I'm not sure if it qualifies to make that list.

### DIFF
--- a/linux/build
+++ b/linux/build
@@ -91,7 +91,7 @@ else
     else
         (cd ../src/rust && CARGO_TARGET_DIR=../../linux/rust cargo build --release) || { echo "Failed. " ; exit 1; }
     fi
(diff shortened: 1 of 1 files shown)


ArchitBhonsle:
Note that, this conflicts with the `builddebug` script which adds a `-g` to `BLD_FLAGS`. This change disregards this option and compiles the rust library in release. I could simply search for `-g` and switch between the debug and release build based on that.

Another more robust way to do this would be to add a `-debug` flag to the `build` and switch between the two based on that.

ArchitBhonsle:
Closing because #1494 solves it better.

PR #1495, opened 2023-03-12
[FEAT] Port all the `common_*` files, teletext and networking portions of `libccx` to rust

elbertronnie:

---

This PR aims to port most of the functions in files `common_*` ,  `teletext.h`, `telxcc.c`, `networking.c` and certain parts of `utility.c` (such as logging) to rust. Certain parts of the codebase is left blank if they are dependant on some code which this PR has not aimed to convert. Such parts will have to be revisited once their dependencies are also converted to rust.

I have placed the new rust code inside a new crate named as `lib_ccxr`, which is a normal rust library crate. This was because a static library crate does not allow doctests. This would also help in seperating ffi code to idiomatic Rust code. Since the Rust approach and C approach to solve a problem are very different, there are 3 layers of indirection I have used for porting:
#### 1. The C-FFI layer
This layers will have function names same as defined in C but with the prefix of `ccxr_`. These functions will handle the conversion between C-native types and Rust-native types and then pass it on to the C-like Rust layer. These are the functions defined in the `ccx_rust` crate under `libccxr_exports` module.
#### 2. The C-like Rust layer
This layer will have function names same as defined in C but work with Rust-native types. The code written using these functions would still be procedural but in Rust, hence the name C-like Rust. The entire task of these functions will be to translate the procedural code to idiomatic Rust code. These are the functions defined in the `c_functions.rs` file in appropriate modules inside the `lib_ccxr` crate.
#### 3. The Idiomatic Rust layer
This layer will be the code written as one is supposed to write in idiomatic Rust. It will have complete documentation and tests. This code will will be situated in the `lib_ccxr` crate.


elbertronnie:
@PunitLodha Can you review this PR?
I have split the PR into multiple meaningful commits, so that atleast each individual commit is easy to review.

cfsmp3:
> @PunitLodha Can you review this PR? I have split the PR into multiple meaningful commits, so that atleast each individual commit is easy to review.

Could you make each commit a separate PR? If you can't, that means each change is not correctly isolated from the rest.

We need each commit to be self-contained (at least don't break anything) and be reversible by itself - we don't want to merge a huge PR with lots of commits that must exist together.

(we're still paying for playing fast-and-loose in the past)

elbertronnie:
@PunitLodha @cfsmp3 I have tried to this into multiple PRs: #1551, #1552, #1553, #1554, #1555, #1556, #1557, #1558, #1559, #1560.

PR #1510, opened 2023-03-21
[IMPROVEMENT] add erosion then dilation after quantization

ziexess:

---

This PR is the suggested fix to the infamous OCR noise problem in issue #929 
I've checked the discussion there and found that the last step that hasn't been implemented is the erosion-dilation to remove the noise surrounding the text after using the built-in `quantize_map` function.

I've spent the last two days decrypting the `quantize_map` function, partly because of how weird the png format stores the color palette and bitmap seperately and partly because the function output doesn't respect it's input parameters. i.e: when passing the `max_color=3` to the function, it actually outputs an image with only 2 colors, one of which is _transparency_.

This is a proof of concept of how erosion and dilation do actually fix the problem and improve the output quality by **a lot**. I've done some experimenting with the kernel size and found that 2x2 is the most suitable for removing that kind of noise. 
Right now, there's a small quirk that needs fixing which is that the erosion sometimes removes the second line of the subtitle altogether.
I'll solve that issue and mark the PR as ready for review.

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -691,6 +691,54 @@ char *ocr_bitmap(void *arg, png_color *palette, png_byte *alpha, unsigned char *
 	return text_out;
 }
 
+void erode(png_color *palette, png_byte *alpha, uint8_t *bitmap, int w, int h, int nb_color)
+{
+	int background_index;
+	for (background_index = 0; background_index < nb_color; background_index++)
+	{
+		if (alpha[background_index])
+		{
+			break;
+		}
+	}
+	// we will use a 2*2 kernel for the erosion
+	for (int row = 0; row < h - 1; row++)
+	{
+		for (int col = 0; col < w - 1; col++)
+		{
+			if (alpha[bitmap[row * w + col]] || alpha[bitmap[(row + 1) * w + col]] ||
+			    alpha[bitmap[row * w + (col + 1)]] || alpha[bitmap[(row + 1) * w + (col + 1)]])
+			{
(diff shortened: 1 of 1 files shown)


ziexess:
After further debugging and saving intermediate images, It seems that the result of the quantization -> erosion -> dilation looks fine and the missing line bug is coming from somewhere else. Probably from this function:
https://github.com/CCExtractor/ccextractor/blob/a0bb56b49be82621243eff9d648ed960544da54f/src/lib_ccx/ocr.c#L960

cfsmp3:
Could you add some example images (before and after would be great :-) ) to your PR so we can see the improvement? 

(I'm reading about these basic operations but the OCR/image were the work of @Abhinav95 if I remember correctly). Also @anshul1912 .

ziexess:
Hi @cfsmp3, sure I will attach some photos the quantization output without and with the erosion-dilation:
4b
4a
1b
1a
2b
2a
Also, to show the improved OCR results, here is a sample from the diff of the resulted subtitle:
Screenshot 2023-03-21 210850

ziexess:
Hey @Abhinav95, thanks for taking the time to review my PR. The kernel size depends on the kind of noise generated from the `quantize_map()` which in turn depends on the implementation of it. So even if the kernel size was variable, the current implementation and its results necessitate that the 2x2 kernel is used. I am actually applying for GSOC this summer to rewrite the OCR system in rust and thought this would be good enough solution until then, since new implementations may introduce different noise which may change the kernel size as you mentioned.

Issue #1483, opened 2023-02-28
Output format depends on the POSIX locale LC_NUMERIC


v-chvila:
For example: https://github.com/CCExtractor/ccextractor/blob/b488126d09de590268a49d6bb18f6c8285147cbf/src/lib_ccx/ccx_encoders_smptett.c#L221
`sprintf()` is configured by LC_NUMERIC. See: https://linux.die.net/man/3/sprintf `Format of the format string`
Final output is inconsistent and it is hard to parse. Do you consider some better solution?

marcybelardo:
We can specify a locale before using sprintf to get the formatting we want, and then immediately return locale to C afterwards I believe?

marcybelardo:
Locales are a nightmare. I might try a fix for this, but wow

cfsmp3:
> The more I research, the more I think locales are a nightmare. I might try a fix for this, but wow

The simplest solution is two just don't use float and use two ints instead. Here's what ChatGPT suggest (I haven't tried, but it gives you an idea)

```
sprintf(str, "      <p begin=\"%02u:%02u:%02u.%03u\" end=\"%02u:%02u:%02u.%03u\" tts:origin=\"%d.%03d%% %d.%03d%%\">\n        <span>",
        h1, m1, s1, ms1, h2, m2, s2, ms2, (int)(col1), (int)((col1 - (int)col1) * 1000), (int)(row1), (int)((row1 - (int)row1) * 1000));
```

Forget about locales :-)

marcybelardo:
Oh I see! I thought that they had to stay as floats :P

cfsmp3:
> Oh I see! I thought that they had to stay as floats :P

No - just make sure the output is correct, i.e. always 3 digits (test for numbers ending in 0, 00, 000)...

cfsmp3:
@v-chvila Could you see if it's fixed for you in master?

cfsmp3:
Closing. @v-chvila feel free to reopen if you still have problems.

PR #1512, opened 2023-03-22
[FIX] ccx_encoders_smptett.c: split float into ints for sprintf

marcybelardo:

---

For issue #1483, refactors formatting of `col1` and `row1` variables to avoid issues with `LC_NUMERIC` auto-formatting float values.

Decided to use `%03u` to keep precision at 3 digits, used `%03u` over `%.3u` for better code readability in the format string.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -13,6 +13,7 @@
 - Fix: tesseract 5.x traineddata location in ocr
 - Fix: fix autoconf tesseract detection problem (#1503)
 - Fix: add missing compile_info_real.h source to Autotools build
+- Fix: refactors SMPTE-TT encoding with int values to avoid locale issues with sprintf 
 
 0.94 (2021-12-14)
 -----------------
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -13,7 +13,7 @@
 - Fix: tesseract 5.x traineddata location in ocr
 - Fix: fix autoconf tesseract detection problem (#1503)
 - Fix: add missing compile_info_real.h source to Autotools build
-- Fix: refactors SMPTE-TT encoding with int values to avoid locale issues with sprintf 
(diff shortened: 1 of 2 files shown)


cfsmp3:
I've been experimenting with this - I think using the proper locale for numeric is the best way to go, otherwise we'll have to change lots of lines of code just to get what the C library gives us for free. I pushed a one-liner comment to use the POSIX locale.

Here's a trivial test:

```
#include <stdio.h>
#include <locale.h>

int main() {
    float col1 = 3000000000000;
    setlocale (LC_ALL, "en_DK.utf8");
    printf ("%1.3f\n", col1);
    setlocale (LC_NUMERIC, "POSIX");
    printf ("%1.3f", col1);
}
```
Which outputs:
```
3000000053248,000
3000000053248.000
```

Hopefully this will fix this issue everywhere.

marcybelardo:
> I've been experimenting with this - I think using the proper locale for numeric is the best way to go, otherwise we'll have to change lots of lines of code just to get what the C library gives us for free. I pushed a one-liner comment to use the POSIX locale.
> 
> 
> 
> Here's a trivial test:
> 
> 
> 
> ```
> 
> #include <stdio.h>
> 
> #include <locale.h>
> 
> 
> 
> int main() {
> 
>     float col1 = 3000000000000;
> 
>     setlocale (LC_ALL, "en_DK.utf8");
> 
>     printf ("%1.3f\n", col1);
> 
>     setlocale (LC_NUMERIC, "POSIX");
> 
>     printf ("%1.3f", col1);
> 
> }
> 
> ```
> 
> Which outputs:
> 
> ```
> 
> 3000000053248,000
> 
> 3000000053248.000
> 
> ```
> 
> 
> 
> Hopefully this will fix this issue everywhere.
> 
> 

Oh I see! Right, I was thinking that fixing every sprintf eventually would be tedious. Well, thank you for your help with all this!

PR #1513, opened 2023-03-22
hardsubx: Add missing `-lavfilter` for hardsubx linking

ldm0:

---

Fixes https://github.com/CCExtractor/rusty_ffmpeg/issues/95


PunitLodha:
@cfsmp3 Won't this fix also be required for cmake and autoconf? I think we should try to keep all of them updated

cfsmp3:
> @cfsmp3 Won't this fix also be required for cmake and autoconf? I think we should try to keep all of them updated

It might be there already (I don't know, I'm on my cell) which would explain why I hit this but not the CI.

Feel free to check if you have time, otherwise I'll do it when I get home 😀

PunitLodha:
Quite weird that it worked without avfilter for me and also on the CI

Issue #1516, opened 2023-03-25
[BUG] WebVTT-Full formatted output splits multi-byte UTF-8 characters


dhouck:
CCExtractor version:
```
CCExtractor 0.94, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
CCExtractor detailed version info
        Version: 0.94
        Git commit: Unknown
        Compilation date: 2022-08-10
        CEA-708 decoder: Rust
        File SHA256: Could not open file
Libraries used by CCExtractor
        libGPAC Version: 1.0.1
        zlib: 1.2.11
        utf8proc Version: 2.4.0
        protobuf-c Version: 1.3.1
        libpng Version: 1.6.37
        FreeType
        libhash
        nuklear
        libzvbi
```
Since some of that is listed as unknown, Iʼll add that this is from the Arch Linux AUR compiled as shown [here](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=ccextractor&id=f1cf3b736614441d9373e63cb0eb8b45714493fd)

# Necessary information

- Is this a regression (i.e. did it work before)? Unknown
- What platform did you use? Linux
- What were the used arguments? `ccextractor test.bin -out=webvtt-full`

# Video links

Input and output files attached, both separately gziped and in a combined zip file (for some reason GitHub wouldnʼt let me attach a TGZ despite claiming it was supported)
[test.bin.gz](https://github.com/CCExtractor/ccextractor/files/11068120/test.bin.gz)
[test.vtt.gz](https://github.com/CCExtractor/ccextractor/files/11068121/test.vtt.gz)
[test.zip](https://github.com/CCExtractor/ccextractor/files/11068123/test.zip)

# Additional information

The captions contain an italicized eighth note, with one space on either side (_` ♪ `_).  The SRT output is correct (`<i> ♪ </i>`); the WebVTT-full output format should have the same line, but instead it puts three bytes inside the `<i>` element.  The UTF-8 encoding for U+266A EIGTH NOTE is 0xE2 0x99 0xAA, so this puts one space and 2/3 of the eighth note bytes inside the italics, and 1/3 of the eighth note bytes outside the italics, making the file no longer valid UTF-8 and instead having two unknown byte sequences.

dhouck:
I did not mean to create this issue yet; I hit the wrong keys on my keyboard and accidentally submitted it before it was ready.

[...]

PR #1518, opened 2023-03-26
[FIX] #1516 in webvtt added support to two-three-four utf-8 bytes

chidam333:

- [x] I have considered and confirmed that this submission will be valuable to others.

---

This pr solves the but #1516


cfsmp3:
@chidam333 Rebase, run clang on your file, and update your PR - that should get you past the format error :-)

rghdrizzle:
@chidam333 type git rebase -head~4

dhouck:
As discussed in #1516, there seems to be an off-by-one error I havenʼt figured out here.  Also, I think it might be easier (at least for UTF-8) to output one byte and then any bytes beginning with `0b10`, if thereʼs no better API for detecting UTF-8 characters here

PR #1522, opened 2023-03-29
[FIX] encoding of solid block in latin-1 and unicode

elbertronnie:

---

The solid block character in Line-21 encoding (`0x7f`) is correctly encoded when converted to UTF-8 as `E2 96 A0` (which is `U+25A0`). But this character is missing in the conversion to latin-1 and unicode.

### DIFF
--- a/src/lib_ccx/ccx_common_char_encoding.c
+++ b/src/lib_ccx/ccx_common_char_encoding.c
@@ -35,6 +35,8 @@ void get_char_in_latin_1(unsigned char *buffer, unsigned char c)
 			case 0x7e: // lowercase n tilde
 				c1 = 0xf1;
 				break;
+			case 0x7f: // Solid Block - Does not exist in Latin 1
+				break;
 			default:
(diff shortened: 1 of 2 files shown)


src/lib_ccx/ccx_common_char_encoding.c
```diff
@@ -292,6 +294,9 @@ void get_char_in_unicode(unsigned char *buffer, unsigned char c)
 	unsigned char c1, c2;
 	switch (c)
 	{
+		case 0x7f: // Solid block
+			c2 = 0x25;
+			c1 = 0xa0;
```
cfsmp3, line 299: Aren't you missing a break here?

cfsmp3 (changes requested):

src/lib_ccx/ccx_common_char_encoding.c
elbertronnie: Added the missing break

PR #1528, opened 2023-04-07
[FIX] Incorrect skipping of packets

### LINKED ISSUE #1468: [BUG] Unable to extract CEA708 closed captions
CCExtractor version: 0.94

# In raising this issue, I confirm the following:

- [x] I have ticked all the boxes in this section and to prove it I'm deleting the section completely to remove boilerplate text.

# Necessary information

- Is this a regression (i.e. did it work before)? `No`
- What platform did you use? `Windows`
- What were the used arguments? `ccextractorwinfull.exe cea708_dtv.mpg`

# Video links

https://www.pixeltools.com/mpeg/cea708_dtv.mpg

# Additional information

Relevant MediaInfo portion:
```
Text
ID                                       : 481 (0x1E1)-1
Menu ID                                  : 1 (0x1)
Format                                   : EIA-708
Muxing mode                              : A/53 / DTVCC Transport
Muxing mode, more info                   : Muxed in Video #1
Duration                                 : 5 s 5 ms
Bit rate mode                            : Constant
Stream size                              : 0.00 Byte (0%)
```
Relevant info from [pixeltools website](https://www.pixeltools.com/tech-tip-closed-captioning-existing.html):
`CEA 708 DTV User moveable and re-sizable captioning for HDTV`

Expected text content:
```
This is a demo of the 
Closed Captioning Insertion 
From PixelTools 
Have a Nice Day!
```

elbertronnie:

---

Partially Fixes #1468 

The reason why none of the text was showing up was because the given sample had service blocks of this kind:
`80 ff 8c ff 9c 18 46 19 00 1f 11 84 92 00 04 90 05 00 91 2a 00 00`
- `80` - Set Current Window to 0 - Does nothing since Window was never defined before.
- `ff` - character from G1 charset - The current implementation skips the entire packet if character is attempted to be inserted in an undefined window, which it was during this point.
- `8c ff` - Delete Windows specified with bitmap `ff` - Does nothing since no windows were defined in the first place.
- `9c ...` - Define Window 4 - Here comes the actual definition of the window that will contain all the text but this is never executed due to the second byte (`ff`).

All the window definitions followed the above pattern, hence a window was never defined, so all the text did not show up.
After fixing the implementation to only skip 1 character instead of entire packet, the following text was extracted:
```
1
00:00:01,234 --> 00:00:01,634
    <font color="#aaaaaa">Closed Captioning Insertion</font>

2
00:00:02,235 --> 00:00:02,635
    <font color="#aaaaaa">From PixelTools </font>

3
00:00:03,236 --> 00:00:03,636
    <font color="#aaaaaa">Have a Nice Day!</font>
```

Only the first line is skipped because the packet stream literally started from `This is a demo of the`. Definition of Window 4 only came later. We could fix it by defining all windows before hand but that will against the CEA-708 spec.

This was enough to get it fixed at the Rust side. The C side required more work.

The given sample file had packets of invalid length specified. The current implementation assumes the given packet length is valid. If the specified length is more than the actual length, it just keeps appending the content of the next packet too. (Here the actual length is till the next Packet Start). This problem was fixed by clearing the packet before the start of a new packet if it was not cleared previously. Hence this would skip the packets of invalid length.
Even though this error was not visible in Rust, I have still ported the code for parity's sake. The code at Rust side was able to clear the packet correctly even before this. I still haven't figured out how. Some help on this would be great.

Also, it is debatable if packets of invalid length should be processed or skipped. For now, I have chosen to skip them but they could very well be processed since such packets in the sample file had service number 0. Some advice on this will be very helpful.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -17,6 +17,7 @@
 - Fix: make webvtt-full work correctly with multi-byte utf-8 characters
 - Fix: encoding of solid block in latin-1 and unicode
 - Fix: McPoodle Broadcast Raw format for field 1
+- Fix: Incorrect skipping of packets
 
 0.94 (2021-12-14)
 -----------------
--- a/src/lib_ccx/ccx_decoders_708.c
+++ b/src/lib_ccx/ccx_decoders_708.c
@@ -1406,7 +1406,7 @@ int dtvcc_handle_G0(dtvcc_service_decoder *decoder, unsigned char *data, int dat
 	if (decoder->current_window == -1)
 	{
 		ccx_common_logging.log_ftn("[CEA-708] dtvcc_handle_G0: Window has to be defined first\n");
-		return data_length;
+		return 1;
 	}
 
 	unsigned char c = data[0];
--- a/src/lib_ccx/ccx_dtvcc.c
+++ b/src/lib_ccx/ccx_dtvcc.c
@@ -58,6 +58,12 @@ void dtvcc_process_data(struct dtvcc_ctx *dtvcc,
 				}
 				else
 				{
+					if (dtvcc->is_current_packet_header_parsed)
+					{
+						ccx_common_logging.debug_ftn(CCX_DMT_708, "[CEA-708] dtvcc_process_data: "
+											  "Warning: Incorrect packet length specified. Packet will be skipped.\n");
+						dtvcc_clear_packet(dtvcc);
+					}
 					dtvcc->current_packet[dtvcc->current_packet_length++] = data[2];
 					dtvcc->current_packet[dtvcc->current_packet_length++] = data[3];
 					dtvcc->is_current_packet_header_parsed = 1;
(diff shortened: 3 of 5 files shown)


cfsmp3:
> Also, it is debatable if packets of invalid length should be processed or skipped. For now, I have chosen to skip them but they could very well be processed since such packets in the sample file had service number 0. Some advice on this will be very helpful.

I don't know. In general, we should do the same thing a TV would do.

The problem with processing packets with invalid length is that by definition they are incorrect. Why is the length incorrect? How do we know the payload is correct? And how do we know the length of the payload if we don't know the length of the packet?

cfsmp3:
@elbertronnie can you rebase? That should fix the CI problem

ProjectEGU:
Thank you very much for this fix!

PunitLodha:
> Even though this error was not visible in Rust, I have still ported the code for parity's sake. The code at Rust side was able to clear the packet correctly even before this. I still haven't figured out how. Some help on this would be great.

@elbertronnie This might be the reason,
> Currently, the entry point for the rust decoder is [here](https://github.com/CCExtractor/ccextractor/blob/master/src/rust/src/lib.rs#L53-L75) where we create a new `Dtvcc` struct each time. This causes the data to be reset each time `ccxr_process_cc_data()` is called.

elbertronnie:
> Currently, the entry point for the rust decoder is [here](https://github.com/CCExtractor/ccextractor/blob/master/src/rust/src/lib.rs#L53-L75) where we create a new `Dtvcc` struct each time. This causes the data to be reset each time `ccxr_process_cc_data()` is called.

I see, so porting the code was the correct choice. After the `Dtvcc` struct will be made persisitant that part will become important.

PR #1547, opened 2023-07-28
[FEAT] Add argument parsing in ccextractor rust

prateekmedia:

---

Rewrote argument parsing with all the structs and enums too into rust.

Help arguments demo:
image


src/rust/src/params.rs
```diff
@@ -0,0 +1,1440 @@
...
+
+#[cfg(target_os = "macos")]
+const DEFAULT_FONT_PATH: &str = "/System/Library/Fonts/Helvetica.ttc";
+#[cfg(target_os = "macos")]
+const DEFAULT_FONT_PATH_ITALICS: &str = "/System/Library/Fonts/Helvetica-Oblique.ttf";
+
+#[cfg(not(any(target_os = "windows", target_os = "macos")))]
+const DEFAULT_FONT_PATH: &str = "/usr/share/fonts/truetype/noto/NotoSans-Regular.ttf";
+#[cfg(not(any(target_os = "windows", target_os = "macos")))]
+const DEFAULT_FONT_PATH_ITALICS: &str = "/usr/share/fonts/truetype/noto/NotoSans-Italic.ttf";
```
PunitLodha, line 70: Can use `cfg-if` here too

src/rust/src/params.rs
```diff
@@ -0,0 +1,1440 @@
...
+static mut FILEBUFFERSIZE: i64 = 1024 * 1024 * 16;
+static mut MPEG_CLOCK_FREQ: i64 = 0;
+static mut USERCOLOR_RGB: String = String::new();
+static mut UTC_REFVALUE: u64 = 0;
+const CCX_DECODER_608_SCREEN_WIDTH: u16 = 32;
+
+pub trait FromStr {
+    #[allow(dead_code)]
+    fn to_ms(&self) -> CcxBoundaryTime;
+}
```
PunitLodha, line 81: Is this trait required? Can you not add a function `to_ms(value: &str) -> CcxBoundaryTime` instead?

src/rust/src/params.rs
```diff
@@ -0,0 +1,1440 @@
...
+        }
+    }
+    Ok(())
+}
+
+fn set_output_format_bin(opt: &mut CcxSOptions) {
+    opt.write_format = CcxOutputFormat::Rcwt;
+}
+
+fn set_output_format(opt: &mut CcxSOptions, args: &Args) {
```
PunitLodha, line 190: Every branch in this function requires `args.out` to be `Some`. So you could write it as follows:-
```
if let Some(out) = args.out {
	// if else ladder here
	// use (out == OutFormat::) instead of (can_unwrap && args.out.unwrap() == OutFormat::)
} else {
	// Error part here
}

**Review by PunitLodha — CHANGES_REQUESTED** [review]
Just gone over the formatting issues for now. Haven't looked at the logic yet.
Please also try to fix the clippy warnings

File: src/rust/src/structs.rs [hunk 4]
```diff
@@ -0,0 +1,624 @@
+use crate::{args::OutputField, enums::CcxDebugMessageTypes};
+
+#[derive(Debug, Default)]
+pub struct CcxSTeletextConfig {
```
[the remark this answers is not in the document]
prateekmedia: It would be better to use rust ones as the aim is to completely remove any C code dependency and this would help in that.

src/rust/src/params.rs
prateekmedia: Fixed, although I have used something different but end result is same.

src/rust/src/params.rs
```diff
@@ -0,0 +1,1440 @@
...
+        );
+        std::process::exit(ExitCode::MalformedParameter as i32);
+    }
+}
+
+fn set_input_format_bin(opt: &mut CcxSOptions) {
+    opt.demux_cfg.auto_stream = CcxStreamMode::Rcwt;
+}
+
+fn set_input_format(opt: &mut CcxSOptions, args: &Args) {
```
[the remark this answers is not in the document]
prateekmedia: Same!

src/rust/tests/resources/dictionary.txt
PunitLodha, line 1: You should delete this blank file

src/rust/src/lib.rs
```diff
@@ -169,3 +175,1499 @@ extern "C" fn ccxr_close_handle(handle: RawHandle) {
...
+            let mut tlt_config = CcxSTeletextConfig {
+                ..Default::default()
+            };
+
+            parse_parameters(&mut opt, &args, &mut tlt_config);
+
+            assert!(opt.demux_cfg.ts_autoprogram);
+            assert_eq!(opt.write_format, CcxOutputFormat::Srt);
+            assert_eq!(opt.enc_cfg.encoding, CcxEncodingType::Latin1);
+        }
```
PunitLodha, line 208: You can extract the common code out, in a function and reuse it

```rust
fn parse_args(args: &[&str]) -> CcxSOptions {
    let mut common_args = vec!["./ccextractor", "input_file"];
    common_args.extend_from_slice(args);
    let args = Args::try_parse_from(common_args).expect("Failed to parse arguments");
    let mut options: CcxSOptions = CcxSOptions {
        ..Default::default()
    };
    let mut tlt_config = CcxSTeletextConfig {
        ..Default::default()
    };

    parse_parameters(&mut options, &args, &mut tlt_config);
    options
}

#[test]
fn broken_1() {
    let options = parse_args(&["--autoprogram", "--out", "srt", "--latin1"]);

    assert!(options.demux_cfg.ts_autoprogram);
    assert_eq!(options.write_format, CcxOutputFormat::Srt);
    assert_eq!(options.enc_cfg.encoding, CcxEncodingType::Latin1);
}
```

src/rust/src/params.rs
```diff
@@ -0,0 +1,1445 @@
...
+            opts.enc_cfg.services_charsets[svc - 1] = charsets[i].as_ref().unwrap().to_owned();
+        }
+    }
+
+    if opts.settings_dtvcc.active_services_count == 0 {
+        panic!("[CEA-708] Malformed parameter: no services\n");
+    }
+}
+
+pub fn parse_parameters(opt: &mut CcxSOptions, args: &Args, tlt_config: &mut CcxSTeletextConfig) {
```
PunitLodha, line 427: Functions that take a `opt: &mut CcxSOptions`, should rather be implemented on `CcxSOptions` using `impl CcxSOptions`

src/rust/src/lib.rs
```diff
@@ -169,3 +175,1499 @@ extern "C" fn ccxr_close_handle(handle: RawHandle) {
         let _file = File::from_raw_handle(handle);
     }
 }
+
+#[cfg(test)]
+pub mod tests {
+    pub mod params {
```
PunitLodha, line 181: Is the `pub mod params` needed here?

src/rust/src/params.rs
PunitLodha, line 1: This file should instead be named `parser.rs`

src/rust/src/structs.rs
PunitLodha, line 1: Ideally, all structs and enums would be in the file where they are used/relevant. But for now, you can merge `structs.rs` and `enums.rs` into `common.rs`

src/rust/src/lib.rs
```diff
@@ -169,3 +175,1499 @@ extern "C" fn ccxr_close_handle(handle: RawHandle) {
         let _file = File::from_raw_handle(handle);
     }
 }
+
+#[cfg(test)]
+pub mod tests {
```
PunitLodha, line 180: We are testing the parser here, so you can place the tests in `parser.rs`

PunitLodha (changes requested):
Good progress!

src/rust/src/params.rs
PunitLodha: Side note: Not sure why these structs were named with `_s_` in them in C. You can drop these.
`CcxSOptions` => `CcxOptions`
`CcxSTeletextConfig` => `CcxTeletextConfig`

src/rust/src/lib.rs
prateekmedia: I read the unit tests and decided to move these tests in the same file they are being tested upon, so now they are in params.rs under test module.

src/rust/src/lib.rs
prateekmedia: yes, done!

src/rust/src/params.rs
prateekmedia: @PunitLodha now there's a conflict as opt is &mut and tlt_config is also &mut, so how will Implement for both?

PunitLodha: Most functions take in `opt: &mut CcxOptions`, so you can use `impl CcxOptions` and pass `tlt_config` as `tlt_config: &mut CcxTeletextConfig`

prateekmedia: Done!

PunitLodha:
Now you need to create an extern function in `lib.rs` which parses the arguments as passed by the C code, and which can be called by the C code.
```rust
/// Parse parameters from argv and argc
#[no_mangle]
pub extern "C" fn ccxr_parse_parameters(options: *mut ccx_s_options, argc: c_int, argv: *mut *mut i8 (or something similar...)) -> c_int {
   // Convert argv to Vec<String> and pass it to parse_parameters
    let args: Vec<String> = ...
    let args = Args::try_parse_from(args); // Handle the error here
    let opt = CcxOptions::default();
    opt.parse_parameters(args); 
    // Convert the rust struct (CcxOptions) to C struct (ccx_s_options), so that it can be used by the C code 
    options = opt.to_options(); 
}
```

And then call it from C

```
// At the start
#ifndef DISABLE_RUST
extern int ccxr_parse_parameters(struct ccx_s_options *opt, int argc, char *argv[])
#endif


// Inside main()
#ifndef DISABLE_RUST
	int compile_ret = ccxr_parse_parameters(api_options, argc, argv);
#else
	int compile_ret = parse_parameters(api_options, argc, argv);
#endif
```

PunitLodha:
You'll have to generate bindings for it. Check out `build.rs`

src/lib_ccx/lib_ccx.h
PunitLodha, line 1: Could you remove these formatting changes

prateekmedia:
cc @cfsmp3 @canihavesomecoffee, The tests are all with 0's as they are ran on C code, which uses old params and not on rust code.
Rust code is already unit tested but still after C parameter PR is merged this can also be rebased and merged so its all good from my side. Already reviewed by @PunitLodha

prateekmedia:
It seems to be not working. This is the issue:
```
Invalid option to CCextractor Library
```

prateekmedia:
This PR needs to implement these methods to let tests run:
add_file_sequence
append_file_to_queue

Or use these existing methods from c as extern.

prateekmedia:
So now the Args are parsing correctly but the processing output is not visible.

PunitLodha:
> So now the Args are parsing correctly but the processing output is not visible.
@prateekmedia is this fixed?

PunitLodha:
@prateekmedia could you split out the CI changes into another PR?

prateekmedia:
@PunitLodha #1612

PR #1551, opened 2023-08-27
[FEAT] Create `lib_ccxr` and `libccxr_exports`

elbertronnie:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor.

---

This is an attempt to split #1495 into multiple independant PRs.

This is the first PR that sets up the modules structure and their respective cargo toml files. Every upcoming PR for `lib_ccxr` will be dependant on this.

PR #1556, opened 2023-08-27
[FEAT] Add time units module in `lib_ccxr`

elbertronnie:

---

This is an attempt to split #1495 into multiple independant PRs. Dependant on #1551.

This PR adds types that enables to store time in different representations like simple milliseconds, number of frames, etc. And also allow to interconvert between them. It also provides functions for formatting time in different kinds of formats. Certain functions are integrated into C codebase.


src/rust/lib_ccxr/src/util/time/units.rs
```diff
@@ -0,0 +1,634 @@
...
+
+/// Represent the number of clock ticks as defined in Mpeg standard.
+///
+/// This number can never be negetive.
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Ord, PartialOrd, Add, Sub)]
+pub struct MpegClockTick(i64);
+
+impl MpegClockTick {
+    /// The ratio to convert a clock tick to time duration.
+    pub const MPEG_CLOCK_FREQ: i64 = 90000;
```
PunitLodha, line 452: This is a global value in C. AFAIK, If you redefine it here it will contain an incorrect value, if the C constant is changed

PunitLodha (changes requested):

src/rust/lib_ccxr/src/util/time/units.rs
elbertronnie: I have converted it into a AtomicI64

PunitLodha: Why is there a need to use atomics here? We are running a single thread

elbertronnie: The only way to create a global mutable in rust is to declare a static variable. Because a static variable can be changed by multiple threads, it requires atomics. There is a way to bypass it using a mutable static variable and using unsafe eveytime we try to access or change it. Do you want me go with this approach? I was hesitant at first since it used unsafe rust.

PunitLodha: Yes, as we don't use any multithreading, I would prefer if you just re-use the C global variable and not redeclare it in rust.
This is what we have decided in #1547 too

elbertronnie: I have now re-used `MPEG_CLOCK_FREQ` from C.

PR #1557, opened 2023-08-27
[FEAT] Add net module in `lib_ccxr`

elbertronnie:

---

This is an attempt to split #1495 into multiple independant PRs. Dependant on #1551, #1556 and #1553.

This PR adds networking primitives to send and receive captions across the internet. The networking functions are integrated into C codebase.

NOTE: I was able to test the TCP implementation but I could check if the UDP implementation is correct or not, since there is no way to send caption through UDP.


src/rust/lib_ccxr/src/util/net/common.rs
```diff
@@ -0,0 +1,353 @@
...
+        let title_len = title.len() + 1;
+        let desc_len = desc.len() + 1;
+        let lang_len = lang.len() + 1;
+        let category_len = category.len() + 1;
+
+        let total_len = start_len + stop_len + title_len + desc_len + lang_len + category_len;
+        let mut data = Vec::with_capacity(total_len);
+
+        data.extend_from_slice(start.as_bytes());
+        data.extend_from_slice("\0".as_bytes());
```
IshanGrover2004, line 169: Is this necessary?? @elbertronnie

IshanGrover2004: Is this because C code uses `memcpy`. But still is this really needed?

elbertronnie:
```c
t = 1;
if (title != NULL)
    t += strlen(title);

// ------- other code -------

if (title != NULL)
    memcpy(end, title, t);
end += t;
```
Yes, because `memcpy` will also also copy the `NUL` character. This is necessary, otherwise how will the program know where one string starts and other ends. You are not passing the length of each string anyway.

elbertronnie: And I was busy for the past week, I will probably review both your PRs tomorrow or day after tomorrow.

Issue #1421, opened 2022-03-08
[BUG] MP4 box name as random video data at beginning of TS-recording trips format-detection


hurda:
CCExtractor version: 0.94

# In raising this issue, I confirm the following:

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.

# Necessary information

- Is this a regression (i.e. did it work before)? NO
- What platform did you use? WINDOWS
- What were the used arguments? `ccextractorwinfull.exe -autoprogram -out=srt -bom -utf8 file.ts`

# Video links

* Is one needed?

# Additional information

After running countless of DVB-recordings through ccextractor to the subtitles from the teletext, this file was the first to not getting processed at all, instead I got this console-output:
```
>ccextractorwinfull.exe -out=srt -bom -utf8 file.ts
CCExtractor 0.94, Carlos Fernandez Sanz, Volker Quetschke.
Teletext portions taken from Petr Kutalek's telxcc
--------------------------------------------------------------------------
Input: K:\file.ts
[Extract: 1] [Stream mode: Autodetect]
[Program : Auto ] [Hauppage mode: No] [Use MythTV code: Auto]
[CEA-708: 63 decoders active]
[CEA-708: using charset "none" for all services]
[Timing mode: Auto] [Debug: No] [Buffer input: Yes]
[Use pic_order_cnt_lsb for H.264: No] [Print CC decoder traces: No]
[Target format: .srt] [Encoding: UTF-8] [Delay: 0] [Trim lines: No]
[Add font color data: Yes] [Add font typesetting: Yes]
[Convert case: No][Filter profanity: No] [Video-edit join: No]
[Extraction start time: not set (from start)]
[Extraction end time: not set (to end)]
[Live stream: No] [Clock frequency: 90000]
[...]

PR #1566, opened 2023-09-04
**[FIX]** fix infinite loop in MP4 file type detector and processor

xopok:
On bad inputs containing e.g. the following sequence of bytes within the first 1MiB "ff ff ff ff 6d 65 74 61" `detect_stream_type` was executing an infinite loop because "ff ff ff ff" was interpreted as a length of the candidate "meta" MP4 box, caused the size_t overflow inside `isValidMP4Box` which pointed `nextBoxLocation` to the previous byte and the execution flow processed the same "meta" again.

---

{pull request content here}

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -21,6 +21,7 @@
 - Fix: Repeated values for enums
 - Cleanup: Remove the (unmaintained) Nuklear GUI code
 - Cleanup: Reduce the amount of Windows build options in the project file
+- Fix: infinite loop in MP4 file type detector.
 
 0.94 (2021-12-14)
 -----------------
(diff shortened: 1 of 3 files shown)


xopok:
This is actually a more robust fix to https://github.com/CCExtractor/ccextractor/issues/1421.
Tests failures look very unrelated.

cfsmp3:
It seems sane to me @canihavesomecoffee  but I think you're the last one playing with the mp4 stuff? I don't remember :-)

canihavesomecoffee (approved):

canihavesomecoffee:
I don't remember why I didn't leave a review, but it looks good to me as well.

Issue #1348, opened 2021-06-13
linux/build-static.sh is obsolete


cfsmp3:
It builds CCExtractor 0.82...  we should either try to update it (best) or remove it.

canihavesomecoffee:
It clones from master, so it will build the most recent... However the deps might not be up to date at all.

cfsmp3:
> It clones from master, so it will build the most recent... However the deps might not be up to date at all.

Ah yes. We just like to leave irrelevant comments around
```
# wget https://sourceforge.net/projects/ccextractor/files/ccextractor/0.82/ccextractor.src.0.82.zip;
```

cfsmp3:
@prateekmedia could you check this out? Probably you can fix it without a lot of pain.

Better - just make -static an arg to build :-)

prateekmedia:
Will look at it

prateekmedia:
Will look at it

prateekmedia:
@cfsmp3 Could you explain what needs to be done in this? I looked at the file, ran it and it gave ccextractor-README.txt not found error.

cfsmp3:
> @cfsmp3 Could you explain what needs to be done in this? I looked at the file, ran it and it gave `ccextractor-README.txt` not found error.

This will give you some context

https://github.com/CCExtractor/ccextractor/issues/597

Note that I renamed the script from @traceypooh to build-static.sh without actually changing the context (not a great idea, I realize).

Still, the script is obsolete now but the idea is good. 

Are we able to produce a static binary using Docker as container? (docker is great for this because you can have a reliable environment no matter the starting point in the host).

If you think that's not possible, or really a lot of work, I guess we should just remove build-static.sh for now because obviously it's not working and it's just going to confuse people.

prateekmedia:
I will try to do some testing from my side, let's see if I get it working.

prateekmedia:
@cfsmp3 I think building static is not something we should do, instead why not build an AppImage (which is a universal format) and it will be able to run without being dependent on the system dependencies.

cfsmp3:
> @cfsmp3 I think building static is not something we should do, instead why not build an AppImage (which is a universal format) and it will be able to run without being dependent on the system dependencies.

That's a really good idea.
It would also solve the eternal issues with tesseract training files, etc

prateekmedia:
@cfsmp3 Do you mean that tessdata should not be in the AppImage itself but installed separately?

cfsmp3:
[...]

PR #1592, opened 2024-01-27
[FEATURE] Create linux AppImage for building CCExtractor

IshanGrover2004:

Addresses Issue #1348 

What I did (Current state):
- Replace non-working link
- Replace not working or irrelevant commands
- All library which are required in gcc command are installed correctly
- To compile the ccextractor with all static libraries, we need to have correct `configure.ac` and `Makefile.am` file (currently its giving error, maybe I am doing anything wrong [here](https://github.com/IshanGrover2004/ccextractor-fork/blob/6f70cdbf452a5b4f1f1158062d2b5dba54d27fe7/linux/build-static.sh#L117))

Goal for now:
- Compiling and make `ccextracter` binary file(which is required while making `AppImage`)

Code breaks at:
- When it clones ccextracter and try to run `make` command([here](https://github.com/IshanGrover2004/ccextractor-fork/blob/6f70cdbf452a5b4f1f1158062d2b5dba54d27fe7/linux/build-static.sh#L117)).


IshanGrover2004:
I am using these commands([link](https://github.com/IshanGrover2004/ccextractor-fork/blob/6f70cdbf452a5b4f1f1158062d2b5dba54d27fe7/linux/build-static.sh#L117)):
```sh
autoreconf --install
./configure --without-rust
make ENABLE_OCR=yes
```
which gives error:
```c
In file included from ../src/lib_ccx/params.c:13:
../src/lib_ccx/compile_info.h:7:10: fatal error: compile_info_real.h: No such file or directory
    7 | #include "compile_info_real.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:3073: ../src/lib_ccx/ccextractor-params.o] Error 1
```
I wanted to confirm, am I doing this correct?

prateekmedia:
@IshanGrover2004 is it working yet? Or are you still stuck with the error

IshanGrover2004:
> Is it working yet? Or are you still stuck with the error

Stuck with the error

prateekmedia:
Maybe alpine and docker are causing this, are you sure this is the right way to build AppImage? There's also appimagetool used to build it.

canihavesomecoffee:
You're not running the pre-build.sh script that generates that header file ;)

IshanGrover2004:
> Maybe alpine and docker are causing this, are you sure this is the right way to build AppImage? There's also appimagetool used for building it.

Currently my goal is to get ccextractor binary file which should statically linked then i will attach that binary file in AppImage
For more info read PR discription.

IshanGrover2004:
> You're not running the pre-build.sh script that generates that header file ;)
> 
> (or using cmake; https://github.com/CCExtractor/ccextractor/commit/4fe82abbfcb61f90cf542bb3ebf0e3702782140e)

Okk, I'll try this.
Can you have a look [here](https://github.com/IshanGrover2004/ccextractor-fork/blob/6f70cdbf452a5b4f1f1158062d2b5dba54d27fe7/linux/build-static.sh#L117), is this the right way or should i have to change anything

IshanGrover2004:
> You're not running the pre-build.sh script that generates that header file ;)

Yeah that fixed it (I used `autogen.sh`)

From past few days, I tried many thing to fix this script and I am just stuck with the
```bash
/usr/lib/gcc/x86_64-alpine-linux-musl/13.2.1/../../../../x86_64-alpine-linux-musl/bin/ld: cannot find -lgpac: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [Makefile:1811: ccextractor] Error 1
```
I did build libgpac from source and i able to run `gcc temp.c -lgpac` and i can see the binaries installed(at /usr/local/lib/ & ...)

prateekmedia:
@IshanGrover2004 have you installed libgpac on your system or you are installing it in the alpine linux under this script?

IshanGrover2004:
> @IshanGrover2004 have you installed libgpac on your system or you are installing it in the alpine linux under this script?

I am installing in alpine linux in docker

It is failing in mine as well (I use Arch)
```bash
/usr/bin/ld: cannot find -lgpac: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [Makefile:1811: ccextractor] Error 1
```

prateekmedia:
@IshanGrover2004 The alpine and docker method may have some outdated steps, if you really want to build AppImage I would suggest to use this package:
https://github.com/AppImageCrafters/appimage-builder

This will help you write for ubuntu or other distro and write recipe and generate AppImage.

IshanGrover2004:
> if you really want to build AppImage I would suggest to use this package: https://github.com/AppImageCrafters/appimage-builder

@prateekmedia  Still we want to have first `ccextractor` executable bin file which should be statically linked
And there is where problem occurs

prateekmedia:
I am able to build already, I think you should refer to this:
https://github.com/CCExtractor/ccextractor/wiki/Installation#linux

and 

https://github.com/CCExtractor/ccextractor/blob/master/docs/COMPILATION.MD

IshanGrover2004:
> I am able to build already, I think you should refer to this: https://github.com/CCExtractor/ccextractor/wiki/Installation#linux
> 
> and
> 
> https://github.com/CCExtractor/ccextractor/blob/master/docs/COMPILATION.MD

I am already aware of these resources, this will result a binary which is dynamic linked(you can check by `ldd <binary>`)
but we need to perform **static** build & that is what I am trying to do by 
```sh
git clone https://github.com/CCExtractor/ccextractor
cd ccextractor/linux/
perl -i -pe 's/O3 /O3 -static /' Makefile.am
set +e
./autogen.sh
./configure --without-rust
make ENABLE_OCR=yes
```

My thought process is that if somehow I can have some files(lets say a header file or obj files) which i can attach with gcc compiler to hand compile with all static(`.a`) binaries needed like [this](https://github.com/IshanGrover2004/ccextractor-fork/blob/1e669593c8b2b7d08e5da30f5d5f595b6552b488/linux/build-static.sh#L120)

Lets take a example to explain my approach:
I will install all necessary library statically(Eg. `lib1.a`, `lib2.a`) and I have some program(main.c, etc)
So to compile I can just do `gcc main.c [more program files] lib1.a lib2.a -o out`

That's why I add `obj/*` [here](https://github.com/IshanGrover2004/ccextractor-fork/blob/1e669593c8b2b7d08e5da30f5d5f595b6552b488/linux/build-static.sh#L122) by thinking if i can get obj files from C source files by `make` 
I may be very wrong so correct me @prateekmedia . And if you understand my approach so now you could tell how can i get `objs/*` kinda thing

prateekmedia:
You should not care about dynamic or static, at the end you have to bundle the libraries with .so related to build in the appimage

IshanGrover2004:
> You should not care about dynamic or static, at the end you have to bundle the libraries with .so related to build in the appimage

So do I need to make just AppImage(and doesn't care about static or dynamic)
Because i thought, maybe some people really care about static or dynamic build(that's why `build-static` exist)

prateekmedia:
If you are bundling everything then nothing matters, although do remember if some system have minimum library version requirement like GLIBC 2.32 or more, then you just have to cater that.

IshanGrover2004:
So should I change this PR to actually build AppImage for ccextractor (instead of build-static)

prateekmedia:
You should just build and bundle libraries to AppImage and test in different ubuntu versions to check backward compatibility.

IshanGrover2004:
> You should just build and bundle libraries to AppImage and test in different ubuntu versions to check backward compatibility.

ig [this](https://appimage-builder.readthedocs.io/en/latest/advanced/full_bundle.html#instructions) is the exactly what you want, Is it @prateekmedia ?

prateekmedia:
> > You should just build and bundle libraries to AppImage and test in different ubuntu versions to check backward compatibility.
> 
> ig [this](https://appimage-builder.readthedocs.io/en/latest/advanced/full_bundle.html#instructions) is the exactly what you want, Is it @prateekmedia ?

Yeah, something like that.

IshanGrover2004:
After a lot struggle with different AppImage tools, I successfully made an AppImage which is working.
@prateekmedia Do we want a script for building AppImage(currently failing somewhere but delivered working AppImage) or Just binary for the release page(most organisation do directly have link to the AppImage binary)

IshanGrover2004:
This PR is ready to review and get merged

prateekmedia:
@IshanGrover2004 have you tested the generated AppImage in your system and some other system without any dependency installed like in fresh Ubuntu 20.04?

IshanGrover2004:
> @IshanGrover2004 have you tested the generated AppImage in your system and some other system without any dependency installed like in fresh Ubuntu 20.04?

This script file result AppImage when your system has necessary lib installed(able to run `cmake ../src; make;` in ccextrcator repo)
For making it system independent we have to install every library manually  and setup them in AppDir which will be ig not good idea

prateekmedia:
@IshanGrover2004 We need to supply tessdata within AppImage, that was the purpose of this:
https://github.com/CCExtractor/ccextractor/issues/1348#issuecomment-1481494889

prateekmedia:
Packaging library in the AppImage is needed to make it really replacement of static i.e. system independent as static libraries are packaged already.

IshanGrover2004:
Okk let me try that as well

canihavesomecoffee (approved):
lgtm, it builds on my machine; seems to work with a random sample for test purposes.

We should possibly update and include this in a release build.

IshanGrover2004:
@canihavesomecoffee  This currently build AppImage successfully but it needs improvisation, So Whenever I'll get time to update I will and once achieved open PR

PR #1607, opened 2024-04-19
[FEATURE] Port 708 decoder encoding module to RUST

IshanGrover2004:

---

I have migrated the logic and functions from the [lib_ccx/ccx_decoder_708_encoding.c](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_decoders_708_encoding.c) file to Rust, located in decoder/encoding.rs. This conversion maintains the same functionality that was originally written in C, now implemented in Rust.

To know more See [ccx_decoder_708_encoding.c](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_decoders_708_encoding.c) and compare with File Changed in PR, they are both same

### Background & Testing
- **Current codebase scenerio:** Only C code of [ccx_decoder_708_encoding.c](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_decoders_708_encoding.c) file is available. So while making & running **Rust** test cases which requires these functions do not have body because `cargo test` is just seeing the rust code only.
- **After PR changes:**  Exact same logic of rust code will be there. And I did also added the conditional compilation so that if user build with `-without-rust` then only C code will work or else Rust code will build. And Yes I have manually tested that when compiling with rust src files, rust code is being used and giving the same result.

### DIFF
--- a/src/rust/src/decoder/encoding.rs
+++ b/src/rust/src/decoder/encoding.rs
@@ -25,13 +25,10 @@ pub extern "C" fn dtvcc_get_internal_from_G1(g1_char: u8) -> u8 {
 /// G2: Extended Control Code Set 1
 #[no_mangle]
 pub extern "C" fn dtvcc_get_internal_from_G2(g2_char: u8) -> u8 {
-    if (0x20..=0x3F).contains(&g2_char) {
-        g2_char - 0x20
-    } else if (0x60..=0x7F).contains(&g2_char) {
-        g2_char + 0x20
-    } else {
-        // Rest unmapped, so we return a blank space
-        0x20
+    match g2_char {
+        0x20..=0x3F => g2_char - 0x20,
+        0x60..=0x7F => g2_char + 0x20,
+        _ => 0x20,
     }
 }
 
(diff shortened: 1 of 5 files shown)


thealphadollar (approved):
LGTM.

src/rust/src/decoder/encoding.rs
```diff
@@ -0,0 +1,48 @@
...
+pub extern "C" fn dtvcc_get_internal_from_G2(g2_char: u8) -> u8 {
+    if (0x20..=0x3F).contains(&g2_char) {
+        g2_char - 0x20
+    } else if (0x60..=0x7F).contains(&g2_char) {
+        g2_char + 0x20
+    } else {
+        // Rest unmapped, so we return a blank space
+        0x20
+    }
+}
```
PunitLodha, line 36: Small nitpick but you could use a match here instead

PunitLodha (commented):
Are the rust extern functions automatically picked up by C because they have the same name? I remember having to define `extern fn ...` in C and then having to call it like this,
```c
#ifndef DISABLE_RUST
	// Use rust function here
#else
	// Use C function here
#endif
```

Is that not needed?

src/rust/src/decoder/encoding.rs
IshanGrover2004: Done @PunitLodha

PunitLodha:
@prateekmedia Any idea why Windows is failing here?

IshanGrover2004:
@PunitLodha You can go on for merge if you find correct, btw this method of having same name is working, I have checked it

And another thing I have also Enable the `default_feature` of bindgen bcs this becomes useful while developing in Rust

PunitLodha (approved):

PR #1630, opened 2024-08-01
[IMPROVEMENT] Use Corrosion to build Rust code

emilazy:

---

Currently, using the CMake Ninja generator fails when Rust code is enabled, because the dependencies and outputs of the Rust code compilation are not fully declared. I tried to fix this with the `BYPRODUCTS` feature, but it seemed like it wouldn’t really work. Instead, I ported the Rust build to [Corrosion](https://github.com/corrosion-rs/corrosion), which is a CMake package designed to integrate CMake and Cargo. Corrosion is used by other projects that combine C++ and Rust, like the work‐in‐progress Rust port of the [fish shell](https://github.com/fish-shell/fish-shell). Now people can get faster, more reliable builds using Ninja, and the CMake code is simpler and cleaner.

Developers and users do not need to explicitly install Corrosion, thanks to CMake’s `FetchContent` feature. However, if Corrosion is installed on the system, that version will be used.

I bumped the minimum CMake requirement to 3.15.0, as Corrosion requires this version and CMake was warning that compatibility for versions as old as the current declaration will be removed in the future. However, if this is not desired, I could lower it back down so that users on older versions of CMake can still build the project as long as they disable Rust, and the `policy_max` argument of [`cmake_minimum_required`](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) could be used to fix the warning.


emilazy:
The CI failure looks like an FFmpeg compatibility error rather than anything this should have caused, although I’m not sure why it would be happening only on macOS.

emilazy:
Pushed a change to use the more elegant `FIND_PACKAGE_ARGS` argument to `FetchContent_Declare` to achieve the behaviour of using the system Corrosion package if present. This bumps up the required CMake version to compile the Rust code a little further, although the required version is still older than the one present in e.g. Debian stable. I can undo this if it‘s not preferred.

PunitLodha (approved):
This seems like a good improvement. Thanks for adding Corrosion!

PR #1652, opened 2025-01-07
Update dockerfile

Maus-313:
added /usr for three libs

before fixing this error ->
  docker output:=> 
  
    ERROR [final 14/20] COPY --from=builder /lib/libz.so.1 /lib/                                    0.0s 
     => ERROR [final 15/20] COPY --from=builder /lib/libssl.so.3 /lib/                                  0.0s 
     => ERROR [final 16/20] COPY --from=builder /lib/libcrypto.so.3 /lib/                               0.0s 
    ------
     > [final 14/20] COPY --from=builder /lib/libz.so.1 /lib/:
    ------
    ------
     > [final 15/20] COPY --from=builder /lib/libssl.so.3 /lib/:
    ------
    ------
     > [final 16/20] COPY --from=builder /lib/libcrypto.so.3 /lib/:

which means the library is missing at the location we defined. So I get the official latest alpine image from the docker hub to find this file. Using the docker desktop I check the "Files" of this container. I found that these three files' positions have been changed from "/lib" to "/usr/lib".
image


prateekmedia (approved):
LGTM! Thanks.

Issue #1511, opened 2023-03-22
Memory leak on OCR code


cfsmp3:
According to address sanitizer. I tried to track this down for a bit but the logic we want (clearing data structures where we export the subtitles to file instead of close to where they are allocated... sigh) I didn't succeed today. Anyway, this is the output, which seems straight-forward until you start hunting it down :-)

The test file is here https://www.polyml.org/download/ (it's from another bug report):

```

Direct leak of 873472 byte(s) in 853 object(s) allocated from:
    #0 0x7f08ebcbf90f in __interceptor_malloc ../../../../src/libsanitizer/asan/asan_malloc_linux.cpp:69
    #1 0x55caf596b9b0 in write_dvb_sub ../src/lib_ccx/dvb_subtitle_decoder.c:1615
    #2 0x55caf596d4d1 in dvbsub_handle_display_segment ../src/lib_ccx/dvb_subtitle_decoder.c:1768
    #3 0x55caf596ddca in dvbsub_decode ../src/lib_ccx/dvb_subtitle_decoder.c:1874
    #4 0x55caf59c2699 in process_data ../src/lib_ccx/general_loop.c:644
    #5 0x55caf59c567e in process_non_multiprogram_general_loop ../src/lib_ccx/general_loop.c:968
    #6 0x55caf59c629b in general_loop ../src/lib_ccx/general_loop.c:1062
    #7 0x55caf58c3986 in api_start ../src/ccextractor.c:205
    #8 0x55caf58c5cdb in main ../src/ccextractor.c:463
    #9 0x7f08eb02350f in __libc_start_call_main ../sysdeps/nptl/libc_start_call_main.h:58

Direct leak of 1436 byte(s) in 1436 object(s) allocated from:
    #0 0x7f08ebc73658 in __interceptor_strdup ../../../../src/libsanitizer/asan/asan_interceptors.cpp:439
    #1 0x55caf593cb6d in ocr_bitmap ../src/lib_ccx/ocr.c:355
    #2 0x55caf59426c4 in ocr_rect ../src/lib_ccx/ocr.c:963
    #3 0x55caf596c35d in write_dvb_sub ../src/lib_ccx/dvb_subtitle_decoder.c:1675
    #4 0x55caf596d4d1 in dvbsub_handle_display_segment ../src/lib_ccx/dvb_subtitle_decoder.c:1768
[...]

PR #1675, opened 2025-03-15
[FIX] DVB OCR: Memory Leak & Quantization Issues

### LINKED ISSUE #985: [PROPOSAL] Unable to extract subtitles from DVB video samples when build using cmake by passing -DWITH_FFMPEG=ON flag

CCExtractor version (using the --version parameter preferably) : **X.X**

- [x ] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x ] I have checked that the issue I'm posting isn't already reported.
- [x ] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x ] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x ] I have used the latest available version of CCExtractor to verify this issue exists.

- [x ] I am an active contributor to CCExtractor.

**Necessary information**
- Is this a regression (did it work before)? [x ] NO | [ ] YES - *please specify the last known working version*
- What platform did you use? [ ] Windows - [x ] Linux - [ ] Mac
- What were the used arguments? `-autoprogram`

**Video links (replace text below with your links) **
https://drive.google.com/open?id=0B_61ywKPmI0TdTdSTi1iN251WHM

**Additional information**

Unable to extract subtitles from DVB video samples.
screenshot from 2018-05-29 15-32-00

This occurs if I build ccextractor using cmake by passing the -DWITH_FFMPEG=ON flag
Exact build steps : 
screenshot from 2018-05-29 15-34-05

If I build without the -DWITH_FFMPEG=ON flag, ccextractor is able to extract the subtitles
screenshot from 2018-05-29 15-46-47

PS: Make sure you set an alert in GitHub so you get notifications about your ticket. We may need to ask questions and we do everything inside GitHub's system.

hrideshmg:

---
Closes #985

DVB subtitle extraction is currently broken on the latest master build. I've verified this by testing it on the following few files:
- `09-ITV_Red_Heat.ts`  
- `2016-12-15-BBC4.ts`  
- `CHANNEL_4_2016-06-21.ts`  
- `chan7_BBC NEWS.ts`  

I've found two root issues on why this is the case:
1. The first issue is that the ocr'd text in `ocr_bitmap()` is [freed](https://github.com/CCExtractor/ccextractor/blob/a84256da017fa3c8d993262cb8a6436309fc9197/src/lib_ccx/ocr.c#L689) before being returned. Removing this free causes memory leaks (as pointed out by #1511).
2. The second issue lies within the `quantize_map()` function
    - Passing `--quant 0 (or 2)` with the first fix enables proper extraction of DVB subtitles. 

I've spent the past two days trying to understand this function and have narrowed it down to the `erode()` function introduced in [PR 1510](https://github.com/CCExtractor/ccextractor/pull/1510). I believe this is better explained visually, so here are the subtitle bitmaps before and after the `erode()` call for two different video files:

## Before
before_erosion
before_erosion_CHANNEL_4_2016-06-21

## After
after_erosion
after_erosion_CHANNEL_4_2016-06-21

## Fixes
1. The memory leaks are caused due to empty strings that were not being freed due to an [if condition](https://github.com/CCExtractor/ccextractor/blob/a84256da017fa3c8d993262cb8a6436309fc9197/src/lib_ccx/ocr.c#L1064-L1065) that was prematurely returning. I've handled this case and tested it on the files mentioned in #1511.
2. After analyzing the [`erode()`](https://github.com/CCExtractor/ccextractor/blob/a84256da017fa3c8d993262cb8a6436309fc9197/src/lib_ccx/ocr.c#L701-L723) function, I noticed that the text was being eroded based on transparency rather than the text background. This method will only work for bitmaps which have their quantized text color be transparent.
```c
if (alpha[bitmap[row * w + col]] || alpha[bitmap[(row + 1) * w + col]] ||
    alpha[bitmap[row * w + (col + 1)]] || alpha[bitmap[(row + 1) * w + (col + 1)]])
```
I've modified erode and dilate so that they now use the text and text background color rather than the alpha.
I'm getting these colors from the loop which populates the mcit variable. This approach has been pretty successful in my limited amount of testing, however it relies on the assumption that the background and text color will always be the second and third most frequently occurring colors respectively.

`channel5-2018-02-12.ts` is one exception though, in it the text color is the fourth most frequently occurring color (black, the bg color is repeated twice for some reason). So erosion succeeds but dilation fails, the result is still better than the raw quantized results but it might be worthwhile to disable quantization by default.

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -696,23 +696,15 @@ char *ocr_bitmap(void *arg, png_color *palette, png_byte *alpha, unsigned char *
 	return text_out;
 }
 
-void erode(png_color *palette, png_byte *alpha, uint8_t *bitmap, int w, int h, int nb_color)
+void erode(png_color *palette, png_byte *alpha, uint8_t *bitmap, int w, int h, int nb_color, int background_index)
 {
-	int background_index;
-	for (background_index = 0; background_index < nb_color; background_index++)
-	{
-		if (alpha[background_index])
-		{
-			break;
-		}
-	}
 	// we will use a 2*2 kernel for the erosion
 	for (int row = 0; row < h - 1; row++)
 	{
 		for (int col = 0; col < w - 1; col++)
 		{
-			if (alpha[bitmap[row * w + col]] || alpha[bitmap[(row + 1) * w + col]] ||
-			    alpha[bitmap[row * w + (col + 1)]] || alpha[bitmap[(row + 1) * w + (col + 1)]])
+			if (bitmap[row * w + col] == background_index || bitmap[(row + 1) * w + col] == background_index ||
+			    bitmap[row * w + (col + 1)] == background_index || bitmap[(row + 1) * w + (col + 1)] == background_index)
 			{
 				bitmap[row * w + col] = background_index;
 			}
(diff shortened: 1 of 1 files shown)


cfsmp3:
This seems reasonable. 

Hopefully we'll have a working test platform soon to verify :-( @canihavesomecoffee

hrideshmg:
> Hopefully we'll have a working test platform soon to verify :-( @canihavesomecoffee 

That is what I'm currently working on :)

https://ccextractor.zulipchat.com/#narrow/stream/478694-general/topic/Are.20the.20sample.20platform.20tests.20broken.3F

PR #1703, opened 2025-06-13
[FIX] escape special characters ( < > & ) to HTML context

ali-harmonicinc:

---

The `<`, `>`, `&` is not suppose to be appeared in SRT, because SRT supports HTML tags.
So, the content of SRT should be treated as HTML content with those characters being escaped.
As a result, this pull request intends to do the following:
- convert `&` to `&amp;`
- convert `<` to `&lt;`
- convert `>` to `&gt;`

### DIFF
--- a/.github/workflows/build_windows.yml
+++ b/.github/workflows/build_windows.yml
@@ -22,6 +22,7 @@ on:
       - "**.c"
       - "**.h"
       - "windows/**"
+      - "src/rust/**"
 
 jobs:
   build_release:
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,12 @@
 1.0 (to be released)
 -----------------
+- Fix: Regression failures on DVD files
+- Fix: Segmentation faults on MP4 files with CEA-708 captions
+- Refactor: Remove API structures from ccextractor
+- New: Add Encoder Module to Rust
+- Fix: Elementary stream regressions
+- Fix: Segmentation faults on XDS files
+- Fix: Clippy Errors Based on Rust 1.88
 - IMPROVEMENT: Refactor and optimize Dockerfile
 - Fix: Improved handling of IETF language tags in Matroska files (#1665)
 - New: Create unit test for rust code (#1615)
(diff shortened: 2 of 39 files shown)


cfsmp3 (changes requested):
Thank you for the contribution and for thinking about edge cases in subtitle output!

After researching the SRT format specification and testing player behavior, I have concerns about this change:

### The SRT Format Does Not Require HTML Entity Escaping

SRT supports a **limited set of HTML-like formatting tags** (`<b>`, `<i>`, `<u>`, `<font>`), but this does not mean SRT content should be treated as HTML requiring entity escaping.

From [Wikipedia's SubRip article](https://en.wikipedia.org/wiki/SubRip):
> "**Unofficially** the format has very basic text formatting... Formatting is derived from HTML tags"

The key word is "unofficially" - SRT has no formal specification, and the HTML-like tags are just for formatting, not full HTML parsing.

### Most Players Display Entities Literally

The bigger problem is that most SRT players (VLC, older Kodi, etc.) do **NOT** decode HTML entities. If we escape characters:

| Original | Escaped | What Players Show |
|----------|---------|-------------------|
| `Tom & Jerry` | `Tom &amp; Jerry` | "Tom &amp; Jerry" |
| `x < y` | `x &lt; y` | "x &lt; y" |

This would **break** subtitle display for most users.

Kodi only [added entity decoding in v18.7](https://github.com/xbmc/xbmc/issues/15365), and that was specifically for **WebVTT** subtitles from Netflix, not SRT.

### WebVTT vs SRT

You may be thinking of **WebVTT**, which DOES require escaping per its specification. But CCExtractor has separate handling for WebVTT output.

### Potential Buffer Issues

The `ascii_to_html()` function can expand 1 character to 5 (`&amp;`), but the calling code may not have allocated sufficient buffer space for the expansion.

---

If there's a specific use case where you need HTML-escaped output, we could consider:
1. Adding escaping only for WebVTT output format
2. Adding an optional `--escape-html-entities` flag

Could you share more about the specific problem you encountered that led to this PR?

ali-harmonicinc:
Hi @cfsmp3,

Thanks for putting effort to review my changes!
You are right that SRT only supports a limited set of HTML-like formatting tags which I overlooked.

My use case was feeding the SRT files to `gpac` for further conversion like outputing WEBVTT / TTML.
We have test stream includes special characters in the CC and `gpac` treating all `<anything>` as HTML tag and hide it.
I think rather than modifying `ccextractor`, I will do some post processing instead before feeding `gpac` for my case.

Anyway, I will close this pull request. Thanks for your input here!

PR #1707, opened 2025-06-28
fix: XDS segmentation faults

hrideshmg:

---

The regression tests under the XDS category have been failing on Windows due to segmentation faults for a while now. The cause for these errors stem from a small mistake that was made when the CEA-708 decoder functions were ported to Rust.

In specific, here is the offending [line](https://github.com/hrideshmg/ccextractor/blob/54dfce4c89245f71b4ffc71e099e6298550a1cd5/src/rust/src/decoder/service_decoder.rs#L1139):
```Rust
// Add symbol to window
window.rows[window.pen_row as usize] = Box::into_raw(Box::new(sym));
```
To contrast, here is the equivalent C [section](https://github.com/hrideshmg/ccextractor/blob/54dfce4c89245f71b4ffc71e099e6298550a1cd5/src/lib_ccx/ccx_decoders_708.c#L799):
```C
window->rows[window->pen_row][window->pen_column] = symbol;
```

`window.rows` has the following definition: 
```C
dtvcc_symbol *rows[15];
rows[i] = (dtvcc_symbol *)malloc(64 * sizeof(dtvcc_symbol));
```

As we can see, it is an array of pointers where each pointer points to a heap allocated row of `dtvcc_symbols`. In rust, this structure is incorrectly assumed to be an array of `dtvcc_symbols`. I've fixed this by using pointer arithmetic in rust to do the proper operation:
```Rust
unsafe {
    let ptr: *mut dtvcc_symbol =
        window.rows[window.pen_row as usize].add(window.pen_column as usize);
    *ptr = sym;
}
```

The unit tests which directly utilize this operation were also leaking memory due to this, and thus I've made helper functions to properly handle those as well.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 1.0 (to be released)
 -----------------
+- Fix: Segmentation faults on XDS files
 - Fix: Clippy Errors Based on Rust 1.88
 - IMPROVEMENT: Refactor and optimize Dockerfile
 - Fix: Improved handling of IETF language tags in Matroska files (#1665)
--- a/src/rust/src/decoder/service_decoder.rs
+++ b/src/rust/src/decoder/service_decoder.rs
@@ -1136,9 +1136,14 @@ impl dtvcc_service_decoder {
         window.is_empty = 0;
 
         // Add symbol to window
-        unsafe {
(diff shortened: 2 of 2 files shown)


hrideshmg:
@prateekmedia This PR is also ready, could you take a look? Thanks!

PR #1710, opened 2025-07-13
[Rust] Added Encoder Module

steel-bucket:

---
In this PR, I have migrated a large part of `ccx_encoders_common.c` and the library `ccx_encoders_g608.c` alongside it's helper functions. I have also fixed some issues(wrong functions) in the `encoding` module which were caught during regression testing. After this PR all the writing that is done on txt files is done by Rust.

### DIFF
--- a/src/rust/lib_ccxr/src/util/encoding.rs
+++ b/src/rust/lib_ccxr/src/util/encoding.rs
@@ -56,6 +56,8 @@ pub type Latin1Char = u8;
 
 /// Represents a character in UCS-2 encoding.
 pub type Ucs2Char = u16;
+/// Represents a character in UTF-8 encoding.
+pub type Utf8Char = u32;
 
 /// A String-like type containing a sequence of Line 21 encoded characters.
 #[derive(Clone, Debug, Eq, PartialEq, Default)]
@@ -339,7 +341,7 @@ impl From<&str> for Ucs2String {
 
 impl From<&Line21String> for String {
     fn from(value: &Line21String) -> String {
-        value.as_vec().iter().map(|&c| line21_to_utf8(c)).collect()
+        value.as_vec().iter().map(|&c| line21_to_char(c)).collect()
     }
 }
 
@@ -480,7 +482,7 @@ impl EncodedString {
     }
 
     /// Converts this [`EncodedString`] to a format provided by `encoding`, returning a new [`EncodedString`].
-    ///    
+    ///
     /// # Examples
     /// ```rust
     /// # use lib_ccxr::util::encoding::*;
@@ -653,7 +655,7 @@ fn latin1_to_line21(c: Latin1Char) -> Line21Char {
     }
 }
 
-fn line21_to_latin1(c: Line21Char) -> Latin1Char {
+pub fn line21_to_latin1(c: Line21Char) -> Latin1Char {
     if c < 0x80 {
         // Regular line-21 character set, mostly ASCII except these exceptions
         match c {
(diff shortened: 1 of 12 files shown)


src/rust/lib_ccxr/src/util/encoding.rs
```diff
@@ -764,118 +766,143 @@ fn line21_to_latin1(c: Line21Char) -> Latin1Char {
...
+            0xc7 => (0x7c, 1),     // Vertical bar
+            0xc8 => (0xc385, 2),   // Uppercase A, ring
+            0xc9 => (0xc3a5, 2),   // Lowercase A, ring
+            0xca => (0xc398, 2),   // Uppercase O, slash
+            0xcb => (0xc3b8, 2),   // Lowercase o, slash
+            0xcc => (0xe28c9c, 3), // Top left corner
+            0xcd => (0xe28c9d, 3), // Top right corner
+            0xce => (0xe28c9e, 3), // Bottom left corner
+            0xcf => (0xe28c9f, 3), // Bottom right corner
+            _ => (b'?' as u32, 1), // I'll do it eventually, I promise
```
prateekmedia: Can this comment be resolved by adding the remaining ones?

steel-bucket: That's not actually my comment, it is there in the C code, and it's also there in the Encoding module from last year. 
I could add more chars if I had a table, a paper or a reference. 
[This source](https://www.utf8-chartable.de/) seems to have some, but a lot of them are already made. If you can confirm that this source is the one, I will implement it.

prateekmedia:
@steel-bucket some tests are not passing.

steel-bucket:
@prateekmedia I think that's a bug, those tests are not passing in main either I assume. It's the exact same output in the latest PR 1721(not mine). I could make an empty PR to confirm it.

hrideshmg:
Yeah its a bug, its caused by some very minute timing differences in the SRT files. I'm currently looking into it.

src/rust/lib_ccxr/src/util/encoding.rs
```diff
@@ -764,118 +766,143 @@ fn line21_to_latin1(c: Line21Char) -> Latin1Char {
...
+                                    // This are weird chars anyway
         }
     }
 }
+pub fn line21_to_char(c: Line21Char) -> char {
+    let (utf8_packed, byte_count) = line21_to_utf8(c);
+
+    // Extract bytes and create a UTF-8 string
+    let mut bytes = Vec::new();
+    match byte_count {
```
prateekmedia: This logic can be changed to not return two things, like it can just return the u32 value and you can check like:
if utf8_packed <= 0xff {
	// 1 byte
} else if utf8_packed <= 0xffff {
	// 2 byte
} else if utf8_packed <= 0xffffff {
	// 3 byte
}

This way we won't need extra information to be passed back. Any feedback welcome.

src/rust/lib_ccxr/src/util/encoding.rs
prateekmedia: I have very less context around this.

@cfsmp3 @canihavesomecoffee any suggestions for resolving this comment or to keep it like this for now?

src/rust/lib_ccxr/src/util/encoding.rs
steel-bucket: resolved, I compressed the logic

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,483 @@
...
+    // Add null terminator
+    dest.push(0);
+
+    Ok(dest.len() - 1) // Return length without null terminator
+}
+
+fn utf8_to_latin1_map(character: u32) -> Latin1Char {
+    ucs2_to_latin1(char_to_ucs2(char::try_from(character).unwrap()))
+}
+pub fn get_str_basic(
```
prateekmedia: Should we keep pure rust function in rust/lib_ccxr?

I think lib_ccxr was meant to keep pure rust implementation of lib_ccx and the top level rust for interoperability between lib_ccxr and lib_ccx.

steel-bucket: Yes, the `encoder/common.rs` can be entirely moved to lib_ccxr, I just thought it would be more organised if it stayed in the encoder folder, as then we would have another encoder folder in lib_ccxr. @prateekmedia Should I do it?

src/rust/src/encoder/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        Self: Sized;
+}
+impl FromCType<ccx_encoding_type> for Encoding {
+    // Remove after demuxer, just import from ctorust
+    unsafe fn from_ctype(encoding: ccx_encoding_type) -> Option<Self> {
+        Some(match encoding {
+            0 => Encoding::Ucs2,   // CCX_ENC_UNICODE
+            1 => Encoding::Latin1, // CCX_ENC_LATIN_1
+            2 => Encoding::Utf8,   // CCX_ENC_UTF_8
+            3 => Encoding::Line21, // CCX_ENC_ASCII
```
prateekmedia, line 72: This should be ASCII & not Line21?

If anywhere else in rust you are using line21, it should be renamed to ASCII for Encoding::

steel-bucket: I haven't renamed it, that's the code in Main
https://github.com/CCExtractor/ccextractor/blob/b63a29cd2ee24c1d7f576f6b8b3dfb7b58b1a5af/src/rust/lib_ccxr/src/util/encoding.rs#L43
image

src/rust/src/encoder/simplexml.rs
```diff
@@ -0,0 +1,89 @@
...
+            data.characters[line_number].as_mut_ptr(),
+            context.trim_subs,
+            ccx_encoding_type_CCX_ENC_ASCII,
+            context.encoding,
+            CCX_DECODER_608_SCREEN_WIDTH as i32,
+        )
+    };
+
+    // Write opening caption tag
+    let _ret = unsafe { write_raw((*context.out).fh, cap.as_ptr() as *const c_void, cap.len()) };
```
prateekmedia, line 30: Can we have pure rust implementation here, you are modifying too much C, which is fine but it can be done in pure rust also.

src/rust/src/encoder/simplexml.rs
```diff
@@ -0,0 +1,89 @@
...
+    };
+}
+
+pub fn write_cc_buffer_as_simplexml(data: &mut eia608_screen, context: &mut encoder_ctx) -> c_int {
+    let mut wrote_something = 0;
+
+    for i in 0..15 {
+        if data.row_used[i] != 0 {
+            write_cc_line_as_simplexml(data, context, i);
+            wrote_something = 1;
```
prateekmedia, line 70: I saw the C code and it does wrong logic I think, since wrote_something is always 1 as it always go inside, can you confirm which logic to keep and what to update?

src/rust/src/encoder/headers_and_footers.rs
```diff
@@ -0,0 +1,545 @@
+#![allow(dead_code)]
```
prateekmedia: Any reason for not using file name similar to C?

headers_and_footers.rs seems weird to me.

steel-bucket: No, it's just for clarity, the `encoders_common` library has a ton of helper functions and all header and footer functions. I just moved them to a separate file. If needed I could put them in `common.rs`.

src/rust/src/encoder/simplexml.rs
steel-bucket: `write_raw` is a function I wrote, it's not imported from C. It's in line 226 of `headers_and_footers.rs`. It doesn't import from C at all, it uses the Rust `File::from_raw_fd`.

src/rust/src/encoder/simplexml.rs
steel-bucket: Let's discuss this in the meet.

src/rust/src/encoder/mod.rs
steel-bucket: @prateekmedia Ok I'll rename Line21 to ASCII, I thought that was weird too.

src/rust/src/encoder/headers_and_footers.rs
steel-bucket: renamed to `metadata.rs`

src/rust/src/encoder/mod.rs
steel-bucket: resolved

src/rust/src/encoder/common.rs
steel-bucket: Moved to `lib_ccxr`.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+    while remaining > 0 {
+        let written = write_raw(fd, current_buf as *const c_void, remaining);
+        if written == -1 {
+            return Err(io::Error::last_os_error());
+        }
+        let bytes_written = written as usize;
+        unsafe {
+            current_buf = current_buf.add(bytes_written);
+        }
+        remaining -= bytes_written;
```
prateekmedia: Why do you need to create another variable here, how about directly using written?
```suggestion
        unsafe {
            current_buf = current_buf.add(written);
        }
        remaining -= written;
```

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+    buffer: &mut [c_uchar],
+    line_num: usize,
+    data: &eia608_screen,
+) -> c_uint {
+    let mut buffer_pos = 0;
+    let line = &data.characters[line_num];
+
+    for i in line.iter().take(33) {
+        let bytes_written = match ctx.encoding {
+            ccx_encoding_type_CCX_ENC_UTF_8 => {
```
prateekmedia: You can resolve these into rust version and use them.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+                let utf8_packed = line21_to_utf8(*i);
+                let (byte_count, bytes) = if utf8_packed <= 0xff {
```
prateekmedia, line 47: We can optimize this a lot, I have DM'ed you on Zulip with more info.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+                        [
+                            ((utf8_packed >> 16) & 0xFF) as c_uchar,
+                            ((utf8_packed >> 8) & 0xFF) as c_uchar,
+                            (utf8_packed & 0xFF) as c_uchar,
+                        ],
+                    )
+                };
+
+                if buffer_pos + byte_count - 1 < buffer.len() {
+                    buffer[buffer_pos..(byte_count + buffer_pos)]
```
prateekmedia, line 70: This should be like this after you modify line21_to_utf8 to return vector of u8's, I have DM'ed you with more details:
```suggestion
                let bytes = line21_to_utf8(*i);
                byte_count = bytes.len();

                buffer[buffer_pos] = bytes[0] as c_uchar;
                buffer[buffer_pos + 1] = bytes[1] as c_uchar;
                buffer[buffer_pos + 2] = bytes[2] as c_uchar;

                byte_count
```

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+                        .copy_from_slice(&bytes[..byte_count]);
+                    byte_count
+                } else {
+                    0
+                }
```
prateekmedia: You can just use this:
```suggestion
		let utf8_char = line21_to_utf8(*i);
		
		let first_non_zero = (utf8_char.leading_zeros() / 8) as usize;
		let bytes = utf8_char.to_be_bytes(); // UTF-8, big-endian
		
                // Since we use u32, it would have leading zeroes
		byte_count = bytes.len() - first_non_zero;
		
		if buffer_pos + byte_count > buffer.len() { return 0; }

                buffer[buffer_pos..(byte_count + buffer_pos)]
                    .copy_from_slice(&bytes[first_non_zero..(first_non_zero + byte_count)]);

                byte_count
```

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+            }
+            ccx_encoding_type_CCX_ENC_ASCII => {
+                if buffer_pos < buffer.len() {
+                    buffer[buffer_pos] = *i;
+                    1
+                } else {
+                    0
+                }
+            }
+            _ => {
```
prateekmedia: If you convert it to rust then you wouldn't need this.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+                0 // This should never be reached
+            }
+        };
+
+        buffer_pos += bytes_written;
+
+        // Break if we've run out of buffer space
+        if bytes_written == 0 {
```
prateekmedia, line 91: If you really want this case, move it above line 111. Also should we really just break out of for loop without any error? Asking because C code doesn't cover this case.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+    let mut ms1: u32 = 0;
+
+    let mut h2: u32 = 0;
+    let mut m2: u32 = 0;
+    let mut s2: u32 = 0;
+    let mut ms2: u32 = 0;
+
+    unsafe {
+        ccxr_millis_to_time(data.start_time, &mut h1, &mut m1, &mut s1, &mut ms1);
+        ccxr_millis_to_time(data.end_time - 1, &mut h2, &mut m2, &mut s2, &mut ms2);
```
prateekmedia, line 169: Missing comment about why "-1" from C code (which seems important).

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+
+    // Increment counter
+    context.srt_counter += 1;
+
+    // Create timeline string for counter
+    let counter_line = format!("{}{}", context.srt_counter, unsafe {
+        std::str::from_utf8_unchecked(std::slice::from_raw_parts(
+            context.encoded_crlf.as_ptr(),
+            context.encoded_crlf_length as usize,
+        ))
```
prateekmedia, line 179: We can extract this to a variable since it's value is used more than once.

src/rust/src/encoder/g608.rs
```diff
@@ -0,0 +1,308 @@
...
+        // Get font encoded
+        let length = get_font_encoded(context, subline_slice, i, data);
+        if write_wrapped(
+            unsafe { (*context.out).fh },
+            &subline_slice[..length as usize],
+        )
+        .is_err()
+        {
+            return 0;
+        }
```
prateekmedia: We can simplify this one as well:
```suggestion
	let line_length = get_line_encoded(context, subline_slice, i, data);
	let color_length = get_color_encoded(context, subline_slice, i, data);
	let font_length = get_font_encoded(context, subline_slice, i, data);
		
	let lengths = [line_length, color_length, font_length];
	for j in 0..3 {
	    if write_wrapped(
	        unsafe { (*context.out).fh },
	        &subline_slice[..lengths[j] as usize],
	    ).is_err() {
        	return 0;
    	   };
  	}
```

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+
+    let mut bytes: c_uint = 0;
+    let mut buffer_pos = 0;
+    let mut text_pos = 0;
+    let text_len = text.iter().position(|&b| b == 0).unwrap_or(text.len());
+    while text_pos < text_len {
+        let current_byte = text[text_pos];
+
+        match ctx.encoding {
+            ccx_encoding_type_CCX_ENC_UTF_8 | ccx_encoding_type_CCX_ENC_LATIN_1 => {
```
prateekmedia: here also can use our rust version of this.

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    bytes
+}
+pub fn write_subtitle_file_footer(ctx: &mut encoder_ctx, out: &mut ccx_s_write) -> c_int {
+    let mut ret: c_int = 0;
+    let mut str_buffer = [0u8; 1024];
+
+    match ctx.write_format {
+        ccx_output_format::CCX_OF_SAMI => {
```
prateekmedia: Here also we can use our Rust version of write_format.

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+            let footer = b"</BODY></SAMI>\n\0";
+
+            // Bounds check for str_buffer
+            if footer.len() > str_buffer.len() {
+                return -1;
+            }
+
+            str_buffer[..footer.len()].copy_from_slice(footer);
```
prateekmedia, line 110: Why using footer.len() here?

prateekmedia: Ok I understood it!

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+
+            if ctx.encoding != ccx_encoding_type_CCX_ENC_UNICODE {
+                debug!(msg_type = DebugMessageFlag::DECODER_608; "\r{}\n",
+                    std::str::from_utf8(&str_buffer[..footer.len()-1]).unwrap_or(""));
```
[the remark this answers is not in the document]
prateekmedia: Ok I understood it!

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+
+    bytes
+}
+pub fn write_subtitle_file_footer(ctx: &mut encoder_ctx, out: &mut ccx_s_write) -> c_int {
+    let mut ret: c_int = 0;
+    let mut str_buffer = [0u8; 1024];
+
+    match ctx.write_format {
+        ccx_output_format::CCX_OF_SAMI => {
+            let footer = b"</BODY></SAMI>\n\0";
```
prateekmedia: I found out that for CCX_OF_SAMI, CCX_OF_SMPTETT, CCX_OF_SIMPLE_XML further logic is same only footer is different. So could you combine these cases into one and only change the value of footer based on which one we are in?

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    true
+}
+pub fn write_bom(ctx: &mut encoder_ctx, out: &mut ccx_s_write) -> c_int {
+    let mut ret: c_int = 0;
+
+    if ctx.no_bom == 0 {
+        match ctx.encoding {
+            ccx_encoding_type_CCX_ENC_UTF_8 => {
```
prateekmedia: Encoding can be converted to our rust one.

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+    }
+
+    ret
+}
+
+pub fn write_subtitle_file_header(ctx: &mut encoder_ctx, out: &mut ccx_s_write) -> c_int {
+    let mut used: c_uint;
+    let mut header_size: usize = 0;
+
+    match ctx.write_format {
```
prateekmedia: Can convert to rust one.

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+                header_size += header_line.len();
+            }
+
+            if !request_buffer_capacity(ctx, (header_size * 3) as c_uint) {
+                return -1;
+            }
+
+            for header_line in WEBVTT_HEADER {
+                let line_to_write = unsafe {
+                    if ccx_options.enc_cfg.line_terminator_lf == 1 && *header_line == "\r\n" {
```
prateekmedia, line 354: Add the comment like C one for making this clearer.

src/rust/src/encoder/common.rs
```diff
@@ -0,0 +1,545 @@
...
+                }
+
+                if write_raw(out.fh, ctx.buffer as *const c_void, used as usize) < used as isize {
+                    info!("WARNING: Unable to write complete Buffer\n");
+                    return -1;
+                }
+            }
+        }
+
+        ccx_output_format::CCX_OF_SAMI => {
```
prateekmedia: CCX_OF_SSA, CCX_OF_SAMI, CCX_OF_SMPTETT, CCX_OF_SIMPLE_XML have similar format just header value is different, you can combine them to single case.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
```diff
@@ -0,0 +1,483 @@
...
+                    vec![(utf8_packed >> 8) as u8, utf8_packed as u8]
+                } else if utf8_packed <= 0xffffff {
+                    vec![
+                        (utf8_packed >> 16) as u8,
+                        (utf8_packed >> 8) as u8,
+                        utf8_packed as u8,
+                    ]
+                } else {
+                    return Err(-1); // Invalid byte count
+                };
```
prateekmedia: Use new logic for this.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
```diff
@@ -0,0 +1,483 @@
...
+                *first_non_blank = i as i32;
+            }
+            *last_non_blank = i as i32;
+        }
+        // break on end‑of‑string or newline
+        if c == b'\0' || c == b'\n' || c == b'\r' {
+            break;
+        }
+    }
+}
```
prateekmedia: You may use rust logic to get last_non_blank and first_non_blank by positions iter and other things, making it native rust.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
```diff
@@ -0,0 +1,483 @@
...
+            c_len = 5;
+        } else {
+            c_len = 1; // Invalid UTF-8, treat as single byte
+        }
+
+        match out_enc {
+            Encoding::UTF8 => {
+                let to_copy = max;
+                if to_copy <= dest.len() {
+                    dest[..to_copy].copy_from_slice(&src[..to_copy]);
```
prateekmedia: Should len be used here or max is fine?
Either way to_copy seems redundant

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
```diff
@@ -0,0 +1,483 @@
...
+                    return to_copy as i32;
+                } else {
+                    return EncoderError::Unsupported as i32;
+                }
+            }
+            Encoding::Latin1 => {
+                if c_len == 1 {
+                    dest[dest_idx] = src[src_idx];
+                    dest_idx += 1;
```
prateekmedia: You can do this += operation at end since it seems common for all if-else ladders.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
```diff
@@ -0,0 +1,483 @@
...
+                }
+            }
+            Encoding::UCS2 => {
+                return EncoderError::Unsupported as i32;
+            }
+            Encoding::ASCII => {
+                if c_len == 1 {
+                    dest[dest_idx] = src[src_idx];
+                    dest_idx += 1;
```
prateekmedia: Same here, do dest_idx += 1 at end

src/rust/src/encoder/g608.rs
steel-bucket: Okay, but this needs a unsafe {}

src/rust/src/encoder/g608.rs
steel-bucket: I guess if the encoding type is somehow corrupted in memory(or someone wrote a function wrong and used this fn), this would be a problem. Otherwise the base case would never be out of the [0,3] range(it's impossible based on the rest of the code). I propose in that case, we just default to UTF8.

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/g608.rs
steel-bucket: resolved
Note - used unsafe as the from_ctype demands it. You could check the comment I left on `encoder/mod.rs`. Basically that function has to be unsafe as it's unsafe in demuxer, which is where I took the code from, when demuxer(or this PR) gets merged, I'll remove them from this mod.rs and import from demuxer's `ctorust.rs`. We could discuss this.

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/g608.rs
steel-bucket: resolved
Note - fallback is Encoding::default() which is UTF8
The only place where this could fail is if someone uses the function wrong, or some kind of weird memory corruption. I could add a different case if reqd.

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/g608.rs
steel-bucket: resolved

src/rust/src/encoder/common.rs
steel-bucket: resolved
Note - Similar to before I used Encoding::UTF8 as fallback, this will only problematic if someone uses the function wrong, I think.

src/rust/src/encoder/common.rs
steel-bucket: resolved
Note - Similar to before I used OutputFormat::Raw as fallback, this will only problematic if someone uses the function wrong, I think.

src/rust/src/encoder/common.rs
steel-bucket: resolved

src/rust/src/encoder/common.rs
steel-bucket: resolved
Note - Similar to before I used Encoding::UTF8 as fallback, this will only problematic if someone uses the function wrong, I think.

src/rust/src/encoder/common.rs
steel-bucket: resolved

src/rust/src/encoder/common.rs
steel-bucket: resolved

src/rust/src/encoder/common.rs
steel-bucket: resolved

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
steel-bucket: resolved

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
steel-bucket: I didn't quite understand this comment, I tried to make it more rusty, please take a look at the new code and see if it is ok.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
steel-bucket: It's tested for a real file and it works ok I think, I did remove the redundant to_copy as said.

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
steel-bucket: resolved

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
steel-bucket: resolved

src/rust/lib_ccxr/src/encoder/txt_helpers.rs
prateekmedia: You can ignore this comment!

src/rust/src/encoder/common.rs
prateekmedia: Thanks for simplifying SpuPng and transcript too!

PR #1722, opened 2025-08-03
refactor: remove api structures

hrideshmg:

---
Removes the API structures from `ccextractor.c`. This was causing a regression due to mismatch between the options maintained by `api_options` and `ccx_options`.

### DIFF
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -38,7 +38,7 @@ void print_end_msg(void)
 	mprint("https://github.com/CCExtractor/ccextractor/issues\n");
 }
 
-int api_start(struct ccx_s_options api_options)
+int start_ccx()
 {
 	struct lib_ccx_ctx *ctx = NULL;	      // Context for libs
 	struct lib_cc_decode *dec_ctx = NULL; // Context for decoder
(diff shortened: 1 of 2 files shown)


prateekmedia (approved):
LGTM!

prateekmedia:
@hrideshmg why doesn't that one case from wtv work on windows?

hrideshmg:
> @hrideshmg why doesn't that one case from wtv work on windows?

It seems to have timed out. WTV files currently take a long time to run and sometimes they hit the maximum allowed run duration. Will try to optimize them when I take an attempt at them later, its not caused by this PR at the very least.

hrideshmg:
@prateekmedia could you review this

src/ccextractor.c
```diff
@@ -430,23 +430,18 @@ int api_start(struct ccx_s_options api_options)
...
-
 int main(int argc, char *argv[])
 {
 	setlocale(LC_ALL, ""); // Supports non-English CCs
 			       // Use POSIX locale for numbers so we get "." as decimal separator and no
 			       // thousands' groupoing instead of what the locale might say
 	setlocale(LC_NUMERIC, "POSIX");
 
-	struct ccx_s_options *api_options = api_init_options();
-	parse_configuration(api_options);
```
prateekmedia, line 447: You removed this function, why?

hrideshmg: The only thing `api_init_options()` does is call `init_options()` and return a pointer to the `ccx_options` struct. Since we're now using the global `ccx_options` struct directly, I don't think this pattern is necessary.

PR #1733, opened 2025-08-19
fix: unicode encoding regression

hrideshmg:

---

CCextractor does not currently produce the correct output when we try to encode the subtitles in unicode (by passing `--unicode`). This is a regression currently reported by the sample platform under the options category. 
 
The problem seems to be caused by order differences between the C and Rust enums. The C [enum](https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_common_constants.h#L222) has the unicode entry at position 0 but the Rust [enum](https://github.com/CCExtractor/ccextractor/blob/master/src/rust/lib_ccxr/src/util/encoding.rs#L43) has it at position 3.

I'm not exactly sure *why* this fixes the issue because we are using explicit match by value statements when converting between Rust and C so this shouldn't make a difference but it does yield the correct output nonetheless.


prateekmedia (approved):
LGTM!

Issue #1743, opened 2025-09-02
[FEATURE] Add Default X-TIMESTAMP-MAP Header When No Subtitles Are Present


AbHash-RixE:
CCExtractor version: 0.94

# Necessary information

- Is this a regression (i.e. did it work before)? NO
- What platform did you use? Linux
- What were the used arguments? `./ccextractor --timestamp-map --out=webvtt output_clean.ts`

# Description

**Current Behavior**: When there are no subtitles in the .ts file, the tool (CCExtractor) produces an empty .webvtt file ,
i.e no X-TIMESTAMP-MAP header is produced . This is expected behavior, as no subtitles are present.

**Desired Behavior**: To add a default X-TIMESTAMP-MAP header in the .webvtt file even when no subtitles are present, 
i.e `X-TIMESTAMP-MAP=MPEGTS:0,LOCAL:00:00:00.000`.

This would ensure that the .webvtt file is consistently formatted, even when there are no subtitles present, which is necessary for HLS stream compatibility.

bbgdzxng1:
@AbHash-RixE... Proposed sharpening of your desired behavior statement...

**Desired Behavior:** _If `--timestamp-map` is specified_, add a X-TIMESTAMP-MAP header in the .webvtt file even when no subtitles are present,
i.e

```
WEBVTT
X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:900000

```

(I'm just nitting "default", because technically all `X-TIMESTAMP-MAP` headers are invalid WebVTT, they just happen to be valid HLS, due to a long-running saga around the exact definition of WebVTT metadata headers).  Background:
https://github.com/w3c/webvtt/issues/485#issuecomment-639553893
https://mailarchive.ietf.org/arch/msg/hls-interest/MzTHEqH3FVE5E3chn8ULjUon9b8/

And secondly, nitting that from a format perspective, HLS RFC defines that LOCAL is the first field, MPEGTS is the second.

> `X-TIMESTAMP-MAP=LOCAL:<cue time>,MPEGTS:<MPEG-2 time>`

AbHash-RixE:
Hi @bbgdzxng1, thank you for the clarification and pointing out the correct field order ( LOCAL before MPEGTS ) as
per the HLS spec — | wasn't aware of that .
I'll update the desired behavior.

AbHash-RixE:
Hi @bbgdzxng1,
[...]

PR #1744, opened 2025-09-03
[IMPROVEMENT]: Ensure WebVTT output includes minimal X-TIMESTAMP-MAP for HLS compatibility

AbHash-RixE:

**SUMMARY**

This PR ensures that when `--timestamp-map` & `--out=webvtt` is used and no captions are detected, CCExtractor still writes a minimal header containing : `X-TIMESTAMP-MAP=MPEGTS:0,LOCAL:00:00:00.000`
 This makes the generated .vtt file valid for HLS players, which often require this header even in empty subtitle files.

**IMPLEMENTATION DETAILS**

1. `CHANGES.TXT`

 -Added entry under Unreleased: 1.0
 IMPROVEMENT: Add default X-TIMESTAMP-MAP header for empty subtitle files in WebVTT for better compatibility
 of HLS streaming (#1743)

2. `src/ccextractor.c`

 -Call `webvtt_write_minimal_header()` in `start_ccx()` when no captions are found and `--out=webvtt` is set.

3. `src/lib_ccx/ccx_encoders_helpers.c`

 -Added new helper function `webvtt_write_minimal_header()`
 -Which:
     Initializes encoder context for WebVTT.
     Writes a default `X-TIMESTAMP-MAP=MPEGTS:0,LOCAL:00:00:00.000` line.
     Ensures line terminator respects user config (LF vs CRLF).
     Cleans up encoder context properly.

4. `src/lib_ccx/ccx_encoders_helpers.h`

 -Declared the new function: `void webvtt_write_minimal_header(void);`

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 1.0 (to be released)
 -----------------
+- IMPROVEMENT: Add default X-TIMESTAMP-MAP header for empty subtitle files in WebVTT for better compatibility of HLS streaming (#1743)
 - Fix: HardSubX OCR on Rust
 - Removed the Share Module
 - Fix: Regression failures on DVD files
--- a/src/ccextractor.c
+++ b/src/ccextractor.c
@@ -6,6 +6,7 @@ License: GPL 2.0
 #include "ccextractor.h"
 #include <stdio.h>
 #include <locale.h>
+#include <ccx_encoders_helpers.h>
 
 volatile int terminate_asap = 0;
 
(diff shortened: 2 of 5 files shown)


cfsmp3:
Thank you for working on this HLS compatibility feature! I tested it and found:

**Works:** Default output filename correctly includes `X-TIMESTAMP-MAP=MPEGTS:0,LOCAL:00:00:00.000`

**Bug:** When using `-o /path/to/output.vtt`, the X-TIMESTAMP-MAP header is NOT written. The `webvtt_write_minimal_header()` function creates a new encoder context that doesn't use the custom output filename from the command line.

**Rebase needed:** This PR is 4+ months old and contains unrelated `ENABLE_SHARING` code from before the share module was removed. Please rebase on current master.

To fix the `-o` bug, `webvtt_write_minimal_header()` needs to either:
1. Accept the output filename as a parameter, or
2. Use the existing encoder context instead of creating a new one

cfsmp3:
Hi @AbHash-RixE, this PR is over 4 months old and has some issues:
1. The `-o` flag doesn't work correctly
2. Needs rebase on current master

Are you still interested in completing this? If not, we may close it and implement the feature differently.

AbHash-RixE:
Hi, @cfsmp3  thanks for the review and for testing the PR.

You’re right about the issues you pointed out. The -o bug exists because webvtt_write_minimal_header() creates a new encoder context, so it doesn’t respect the output filename provided on the command line. 

I’ve reviewed PR #1772  , and it directly solves these problems by using the existing encoder context instead of creating a new one.

If it’s helpful, I’d be happy to assist by rebasing and testing PR #1772 against current master, or by validating it further. Otherwise, I’m also fine with closing my PR in favor of that implementation.

Thanks again for the feedback.

cfsmp3:
Closing due to inactivity — the -o flag bug reported in January was never addressed. If you'd like to pick this up again, feel free to open a new PR with the fix. Thanks.

PR #1745, opened 2025-09-06
[FIX] Hardsubx fixes + Mac OS build fixes

prateekmedia:

---

- Fix FFMpeg, tesseract compilation on Mac os.
- Fix Mac os build hardsubx workflow
- Fix Hardsubx


src/lib_ccx/ocr.c
```diff
@@ -103,6 +103,7 @@ char *probe_tessdata_location(const char *lang)
 	    "./",
 	    "/usr/share/",
 	    "/usr/local/share/",
+	    "/opt/homebrew/share/",
 	    "/usr/share/tesseract-ocr/",
 	    "/usr/share/tesseract-ocr/4.00/",
 	    "/usr/share/tesseract-ocr/5/",
```
canihavesomecoffee, line 109: Do we still support these? If not, we can probably remove them to keep the code base clean

prateekmedia: Nothing changed on that part from past, so I do think we support it.

PR #1746, opened 2025-09-08
fix: ocr luminance calculation fix

hrideshmg:

---
Hardsubx in rust currently does not give identical output as the C implementation, often providing inferior results. After debugging for a bit I noticed that there were very slight differences in the luminance masks between both versions. 

In the old C code, the RGB to LAB conversion was done [manually](https://github.com/hrideshmg/ccextractor/blob/ocr_fixes/src/lib_ccx/hardsubx_imgops.c#L69), in rust we [now](https://github.com/hrideshmg/ccextractor/blob/master/src/rust/src/hardsubx/imgops.rs#L15) use the `palette` crate for this purpose. 
The `Srgb` implementation from `palette` we were using till now assumes that the input image is [gamma corrected](https://en.wikipedia.org/wiki/Gamma_correction), and thus performs gamma decoding to make the RGB values linear.

This is unnecessary though, because our input is already linear RGB and was thus causing a faulty conversion. I've modified the code to use [`LinSrgb`](https://docs.rs/palette/latest/palette/type.LinSrgb.html) instead. The Rust implementation now provides identical output to the C code.

Tested on this [sample](https://drive.google.com/file/d/0B_61ywKPmI0TdlRWcVdnajVJUWs/view?resourcekey=0-oCZT9NnDCE9h0jMzcma9wg).

### DIFF
--- a/src/rust/src/hardsubx/imgops.rs
+++ b/src/rust/src/hardsubx/imgops.rs
@@ -1,8 +1,8 @@
-use palette::{FromColor, Hsv, Lab, Srgb};
+use palette::{FromColor, Hsv, Lab, LinSrgb};
 
 #[no_mangle]
 pub extern "C" fn rgb_to_hsv(R: f32, G: f32, B: f32, H: &mut f32, S: &mut f32, V: &mut f32) {
-    let rgb = Srgb::new(R, G, B);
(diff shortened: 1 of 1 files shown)


prateekmedia (approved):
LGTM!

Issue #1718, opened 2025-07-22
[feature request] allow linking to system libraries


apteryks:
Hi,

The build currently relies on  libraries bundled in source form under the src/thirdparty directory. It'd be nice if ccextractor's build system was able to detect compatible libraries from the system and use these instead, ideally being able to function even if  the `src/thirdparty` directory does not exist.

freetype, libpng, utf8proc, zlib, protobuf-c at least are all available from my system for example (GNU Guix).

cfsmp3:
We have vendored some libraries because it makes things much easier to build cross platform. Considering its size, it's just not worth the trouble of checking if there's a system version etc (and then dynamically link etc)

apteryks:
@cfsmp3 CMake should make this relatively easy. It'd be a great improvement for distributions, as freetype and libpng have had security vulnerabilities in the past, so it's important to be able to patch these quickly (and updating/patching them in a single place is easier/faster for distros). I'll try to review PR #1761 now.

cfsmp3:
> [@cfsmp3](https://github.com/cfsmp3) CMake should make this relatively easy. It'd be a great improvement for distributions, as freetype and libpng have had security vulnerabilities in the past, so it's important to be able to patch these quickly (and updating/patching them in a single place is easier/faster for distros). I'll try to review PR [#1761](https://github.com/CCExtractor/ccextractor/pull/1761) now.

This should be done already, see https://github.com/CCExtractor/ccextractor/commit/7075f6291dce80a01f1ee446dc5ffa934cfcefe6

Let us know if that's enough (or not)

cfsmp3:
Closing this this was done and we've tested it for homebrew

PR #1761, opened 2025-11-09
build: add CMake options for system library linking

DhanushVarma-2:
Partial fix for #1718

Adds CMake options for system library linking as foundation for full implementation:

- USE_SYSTEM_ZLIB - Use system zlib instead of bundled
- USE_SYSTEM_LIBPNG - Use system libpng instead of bundled  
- USE_SYSTEM_FREETYPE - Use system freetype instead of bundled

Current behaviour: Options are OFF by default, existing bundled libraries still used
Future work: Implement actual system library detection and linking

This is phase 1 of addressing #1718 - adding the configuration framework before implementing the full system library support.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,16 +2,15 @@ cmake_minimum_required (VERSION 3.24.0)
 project (CCExtractor)
 
 include (CTest)
-
-option (WITH_FFMPEG "Build using FFmpeg demuxer and decoder" OFF)
-option (WITH_OCR "Build with OCR (Optical Character Recognition) feature" OFF)
-option(USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
-option (WITH_HARDSUBX "Build with support for burned-in subtitles" OFF)
+# Build features
+option(WITH_FFMPEG "Build using FFmpeg demuxer and decoder" OFF)
+option(WITH_OCR "Build with OCR (Optical Character Recognition) feature" OFF)
+option(WITH_HARDSUBX "Build with support for burned-in subtitles" OFF)
 
 # Optional system libraries (OFF by default)
-option (USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
-option (USE_SYSTEM_LIBPNG "Use system libpng instead of bundled" OFF)
-option (USE_SYSTEM_FREETYPE "Use system freetype instead of bundled" OFF)
+option(USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
+option(USE_SYSTEM_LIBPNG "Use system libpng instead of bundled" OFF)
+option(USE_SYSTEM_FREETYPE "Use system freetype instead of bundled" OFF)
 
 # Version number
 set (CCEXTRACTOR_VERSION_MAJOR 0)
@@ -54,9 +53,6 @@ add_definitions(
     -DNO_GZIP
 )
 
-if (NOT USE_SYSTEM_FREETYPE)
-    add_definitions(-DFT2_BUILD_LIBRARY)
-endif ()
 
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   add_definitions(-DGPAC_64_BITS)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -6,6 +6,7 @@ include (CTest)
 option (WITH_FFMPEG "Build using FFmpeg demuxer and decoder" OFF)
 option (WITH_OCR "Build with OCR (Optical Character Recognition) feature" OFF)
 option (WITH_HARDSUBX "Build with support for burned-in subtitles" OFF)
+option(USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
```
apteryks: The commit messages mentions USE_SYSTEM_LIBPNG, and USE_SYSTEM_FREETYPE, but I only see USE_SYSTEM_ZLIB; I assume it's still a work in progress? Thanks for your efforts addressing this!

DhanushVarma-2: Thanks for pointing that out — you’re right. The options for libpng and freetype weren’t fully wired yet.
I’m cleaning up the CMake logic so each USE_SYSTEM_* flag properly switches between system and bundled sources (starting with removing the unconditional vendored freetype/libpng paths). I’ll update the PR mostly by today.

DhanushVarma-2:
Fixed the CMake conditional nesting and cleaned up how the existing USE_SYSTEM_* options are handled.
Defaults remain unchanged.

cfsmp3 (changes requested):
Thanks for working on this feature! System library linking is useful for distribution packagers.

However, there are several issues that need to be addressed before this can be merged:

## Issues to Fix

### 1. Duplicate Option Definition
`USE_SYSTEM_ZLIB` is defined twice (lines 8 and 13):
```cmake
option(USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
# ...
option (USE_SYSTEM_ZLIB "Use system zlib instead of bundled" OFF)
```
Remove the duplicate.

### 2. Incomplete FreeType Implementation
`USE_SYSTEM_FREETYPE` option exists but the implementation is incomplete:
- Only removes `FT2_BUILD_LIBRARY` definition
- Missing `find_package(Freetype REQUIRED)` call when enabled
- `FREETYPE_SOURCE` is still defined and included unconditionally

Please add the full implementation or remove the option until it's ready.

### 3. Misplaced Source Directories in ZLIB Block
Inside the `else()` block for `USE_SYSTEM_ZLIB`, there are unrelated source directories:
```cmake
aux_source_directory(${PROJECT_SOURCE_DIR}/thirdparty/lib_hash/ SOURCEFILE)
aux_source_directory(${PROJECT_SOURCE_DIR}/thirdparty/libpng/ SOURCEFILE)
```
These should not be conditional on zlib selection.

### 4. Duplicate macOS arm64 Handling
The macOS arm64 libpng handling appears twice - once at the top level and once inside the `USE_SYSTEM_LIBPNG` else block.

### 5. Formatting
Mixed indentation (spaces vs tabs). Please run through the project's formatting.

## Summary
The concept is good but the implementation needs cleanup. Please address these issues and I'll be happy to approve.

DhanushVarma-2:
Thanks for the detailed review — all points addressed
Removed duplicate USE_SYSTEM_ZLIB option
Fixed misplaced source directories in ZLIB logic
Completed USE_SYSTEM_FREETYPE with proper system/bundled switching
Deduplicated macOS arm64 libpng handling
Cleaned up formatting and conditionals
Please take another look

cfsmp3:
Thank you for working on this! Unfortunately, the CMake build is currently broken - it fails to compile with:

```
fatal error: ft2build.h: No such file or directory
```

The issue is that the freetype include directories are set in the main CMakeLists.txt, but `lib_ccx` is built as a separate library and doesn't receive these includes. The freetype handling happens after `add_subdirectory(lib_ccx)` is processed.

To fix this, you'll need to:
1. Ensure the freetype include directories are added globally before lib_ccx is built, OR
2. Pass the include directories through `EXTRA_INCLUDES` which lib_ccx can use

Also, this PR needs a rebase on current master.

cfsmp3:
Hi @DhanushVarma-2, the build is still broken - freetype includes aren't being propagated to lib_ccx. Any progress on fixing this?

cfsmp3:
Closed due to inactivity, feel free to submit a new one (make sure it's based off master) when you are ready to resume work.

PR #1768, opened 2025-11-21
bump MSRV from 1.54.0 to 1.87.0 (rust)

DhanushVarma-2:
This PR updates CCExtractor’s Rust Minimum Supported Rust Version (MSRV) from 1.54.0 to 1.87.0.
The change is applied consistently across:
Linux build and configure scripts
macOS build and configure scripts
CMake Rust version checks
Clippy configuration (clippy.toml)
With this update, the build system explicitly enforces Rust ≥ 1.87.0, ensuring the declared MSRV matches the Rust features and tooling currently used in the codebase and preventing mismatches during builds and CI


cfsmp3:
## Review Feedback

Thanks for working on this! We've decided to go with bumping the MSRV to 1.87.0 (the code already uses `is_multiple_of()` which requires it).

However, **the PR title and description don't match what the code actually does**:

- **Title says**: "fix(rust): revert is_multiple_of to maintain MSRV 1.54.0"
- **PR actually does**: Bumps MSRV from 1.54.0 to 1.87.0

This happened because the second commit reversed the first commit's intent.

**Please update:**
1. **Title** → Something like: `build: bump MSRV from 1.54.0 to 1.87.0`
2. **Description** → Update to reflect that we're bumping MSRV to support modern Rust features like `is_multiple_of()`

Once the title/description are updated to match the actual changes, this should be good to merge.

PR #1780, opened 2025-12-07
remove label -without-rust

Wvidit:
<**[IMPROVEMENT]**>

**In raising this pull request, I confirm the following (please check boxes):**

- [X] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [X] I have considered, and confirmed that this submission will be valuable to others.
- [X] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [X] I give this submission freely, and claim no ownership to its content.
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [X] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [ ] I am an active contributor to CCExtractor.

---

{minor fix in docs/COMPILATION.md}

PR #1785, opened 2025-12-08
Fix #1779: Update COMPILATION.md to reflect correct Linux build flags

### LINKED ISSUE #1779: A small error in COMPILATION.md
In docs/COMPILATION.md
the -without-rust tag doesn't work for any executable in linux.
one such instance is given below.

```bash
#navigate to linux directory and call the build script

cd ccextractor/linux

 #compile without debug flags
./build

 #compile without rust
./build -without-rust  #<-this one here
...
```

Rahul-2k4:
✅ Problem
The Linux build script (ccextractor/linux/build) does not support the -without-rust flag, yet the documentation suggested it could be used:
     ./build -without-rust
     ./build -debug -without-rust
    
This caused confusion because the build script only supports:
-debug
-hardsubx

Meanwhile, the --without-rust option does exist, but only in the autoconf-based workflow (./configure --without-rust).

🔧 Changes Made

- Removed all incorrect references to ./build -without-rust
- Removed the incorrect combined example ./build -debug -without-rust
- Clarified that --without-rust is a valid option only for autoconf builds
- Updated COMPILATION.md to accurately reflect supported flags

### DIFF
--- a/docs/COMPILATION.MD
+++ b/docs/COMPILATION.MD
@@ -66,9 +66,6 @@ cd ccextractor/linux
 # compile without debug flags
 ./build
 
-# compile without rust
-./build -without-rust
-
 # compile with debug info
 ./build -debug            # same as ./builddebug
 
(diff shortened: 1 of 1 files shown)


prateekmedia:
Duplicate of #1780

PR #1818, opened 2025-12-14
fix: Enable stdout output for CEA-708 captions on Windows

### LINKED ISSUE #1693: ccextractorwinfull.exe can't print captions to stdout
Dear,
Thank you for your good application.
I have found out that ccextractorwinfull.exe cannot make captions printed to stdout at the latest release.
The core .rs file wants the output filename to be provided.
Please check it out.
Thank you.

cfsmp3:
## Summary

Fixes #1693 - ccextractorwinfull.exe can't print captions to stdout

The CEA-708 decoder crashed on Windows when using `--stdout` because the `dtvcc_writer` was not properly initialized for stdout output.

## Changes

1. **Fixed Windows stdout handle initialization** (`ccx_encoders_common.c`):
   - Use `GetStdHandle(STD_OUTPUT_HANDLE)` instead of `NULL` for `fhandle`
   - This allows the Rust writer to detect stdout mode properly

2. **Changed env_logger target from Stdout to Stderr** (`lib.rs`):
   - Debug messages no longer pollute stdout when using `--stdout`
   - This prevents mixing debug output with subtitle content

3. **Removed redundant debug statement** (`service_decoder.rs`):
   - The bare `debug!("{}", self.current_window)` was noisy and duplicated by a more detailed debug statement below it

## Test plan

- [x] All 270 Rust tests pass
- [x] Added `test_writer_output_with_valid_fd` - verifies stdout mode works when fd is pre-set
- [x] Added `test_writer_output_missing_filename_and_fd` - verifies proper error handling (not panic)
- [x] Verified subtitles extract correctly to stdout on Linux
- [x] Windows fix uses proper `GetStdHandle(STD_OUTPUT_HANDLE)` API


syehoonkim:
@cfsmp3 Thanks for fixing this issue. It looks like there are still some conflicts, and I’m wondering if you plan to resolve them. Thanks again.

PR #1824, opened 2025-12-14
fix(timing): Set pts_set to MinPtsSet after PTS jump (fixes #1277)

### LINKED ISSUE #1277: BUG - not all closed captions are found
CCExtractor version: 0.85 - 0.88 (I checked 0.85, 0.85b, 0.87, 0.88)

# In raising this issue, I confirm the following:

I've read all the stuff and did my best to check that this issue is not a duplicate

# Necessary information

I confirmed that this problem does not happen in version 0.84 and earlier (I checked 0.79, 0.80, 0.82)
I used Windows 10 64 bit.
I used the following arguments:
--gui_mode_reports  -autoprogram  -out=srt -bom -latin1 [+input files]

CCExtractor will find the closed captions at the beginning a VOB stream, but at a certain point (e.g. ~6 minute in), it will stop seeing any closed captions.  This doesn't happen with all VOBs, but it will happen every time with certain VOBs.

# Video links

https://drive.google.com/drive/folders/1gb6zOfPrFQJlCOLslUfTYfaIVEy5kEMK?usp=sharing
Included are sequential vob files of an individual episode of an old TV show and the srt files generated by both version 0.84 and 0.88.

# Additional information

If you look at the diff between the 2 srt files.  0.88 produces the same output until entry 213 where there is a discrepancy, and then misses all the entries from 214 to 618
diff_screenshot

cfsmp3:
## Summary

- Fixed issue where DVD VOB files with PTS discontinuities stopped extracting captions after ~6 minutes
- When a PTS jump is detected, the code now properly sets `pts_set = MinPtsSet` after updating `min_pts`, enabling continued `fts_now` calculation
- This was a regression introduced in v0.85 - version 0.84 worked correctly

## Root Cause

In `set_fts()` when handling PTS jumps:
1. The code updated `fts_offset` and `min_pts` for the new timeline
2. But it set `pts_set = Received` without ever setting it back to `MinPtsSet`
3. The `fts_now` calculation only runs when `pts_set == MinPtsSet`, so after a PTS jump, all timestamps would be stuck

## Fix

Set `pts_set = MinPtsSet` after setting `min_pts` in the PTS jump handling code. This allows `fts_now` to continue being calculated correctly after the discontinuity.

## Test Results

Before fix: Only 213 captions extracted, timing stuck at ~7 minutes
After fix: All 618 captions extracted, timing correctly continues to ~21 minutes (matching v0.84 reference output)

## Test plan

- [x] Tested with issue #1277 sample files (DVD VOB files with PTS discontinuities)
- [x] Verified output matches v0.84 reference (618 captions, timing to ~21 minutes)
- [x] All 265 Rust unit tests pass
- [x] Clippy lint check passes

### DIFF
--- a/src/rust/lib_ccxr/src/time/timing.rs
+++ b/src/rust/lib_ccxr/src/time/timing.rs
@@ -391,18 +391,22 @@ impl TimingContext {
                         .as_timestamp(timing_info.current_fps);
                 self.fts_max = self.fts_offset;
 
-                // Start counting again from here
-                self.pts_set = PtsSet::Received; // Force min to be set again
-                self.sync_pts2fts_set = false; // Make note of the new conversion values
+                // Reset sync_pts2fts tracking for the new timeline
+                self.sync_pts2fts_set = false;
 
-                // Avoid next async test - the gap might have occured on
-                // current_tref != 0.
+                // Set new sync_pts accounting for any temporal reference offset.
+                // The gap might have occurred on current_tref != 0.
                 self.sync_pts = self.current_pts
                     - self
                         .current_tref
                         .as_mpeg_clock_tick(timing_info.current_fps, timing_info.mpeg_clock_freq);
-                // Set min_pts = sync_pts as this is used for fts_now
+
+                // Set min_pts to sync_pts and mark as set so fts_now calculation works.
+                // This is essential - without setting pts_set to MinPtsSet, fts_now would
+                // stop being updated after a PTS jump, causing all subsequent timestamps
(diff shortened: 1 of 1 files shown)


cfsmp3:
@canihavesomecoffee can you add this https://drive.google.com/file/d/1KefZPcygUimX5SmuFnvfd1zvAO-3HjIu/view?usp=sharing

to the SP? This is an interesting sample because it was a PTS wrap - not many of those.

This PR fixes the problem.

PR #1832, opened 2025-12-15
[FIX] Prevent creation of empty per-stream subtitle files

### LINKED ISSUE #1282: empty 608 cc3 caption generating with option -12
I am using 
cmd:  /usr/bin/ccextractor -nobom --service all[ISO88591] -nofc -ts   -12 -s 180 /home/dn/ramesh/191505_20201123020000.ts

its generating caption like

-rw------- 1 root root     205930 Nov 23 11:23 191505_20201123020000_1.srt
-rw------- 1 root root          0 Nov 23 11:04 191505_20201123020000_2.srt
-rw------- 1 root root     132090 Nov 23 11:04 191505_20201123020000.p1.svc01.srt
-rw------- 1 root root      76040 Nov 23 11:04 191505_20201123020000.p1.svc03.srt

why  191505_20201123020000_2.srt is empty, its should generate 608-cc3 caption.

VyankateshD206:

---
#### This pull request implements deferred output file creation to avoid producing zero-byte `.srt` files for caption streams that contain no actual data.
- Track per-output writes via wrote_data on ccx_s_write.
- When no data is written to an output, skip footer writing and delete the file on teardown to avoid empty cc3 artifacts.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -709,9 +709,17 @@ void dinit_encoder(struct encoder_ctx **arg, LLONG current_fts)
 		return;
 	for (i = 0; i < ctx->nb_out; i++)
 	{
-		if (ctx->end_credits_text != NULL)
-			try_to_add_end_credits(ctx, ctx->out + i, current_fts);
-		write_subtitle_file_footer(ctx, ctx->out + i);
+		struct ccx_s_write *out = ctx->out + i;
+		if (out->wrote_data)
+		{
+			if (ctx->end_credits_text != NULL)
+				try_to_add_end_credits(ctx, out, current_fts);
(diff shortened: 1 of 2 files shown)


cfsmp3:
Thank you for working on this issue. Unfortunately, the Sample Platform CI shows 47 test failures across multiple categories (Broken, CEA-708, General, Hardsubx, Hauppauge, Options, WTV, XDS), indicating that this approach causes significant regressions.

The `wrote_data` tracking approach is interesting, but it's interfering with the normal output flow in ways that break existing functionality.

We have an alternative approach in PR #1877 that handles this more simply by deleting empty files after the encoding process completes, rather than trying to track writes during encoding. This avoids the complexity and regressions.

I'm closing this PR in favor of #1877. Thank you for your contribution - the effort is appreciated!

PR #1880, opened 2025-12-22
[FIX]: Prevent CEA-708 decoder panic with -stdout on Windows

### LINKED ISSUE #1478: [BUG] -stdout .ts crashes ccextractorwinfull.exe 094
CCExtractor version: 094

# Necessary information

- Is this a regression (i.e. did it work before)? {YES/NO} yes, in 085
- What platform did you use? {Window/Linux/Mac} Windows 11
- What were the used arguments? '-stdout`

# Additional information
- .ts files created by HDHomerun
- ccextractorwinfull.exe is launched in a c# Process
- the only argument is -stdout
- ccextractor emits one line of captions and then crashes
- link to c# project and example files that reproduce the error:
- https://1drv.ms/u/s!Ai_hpL0GElr6o-huZrnjSdk7hmW8Wg?e=Sv0PD6
- click on select and choose cctest1.ts or cctest2.ts. A few seconds later the outputs from stdout and stderror will be displayed in list boxes.

x15sr71:

---

## Issue Summary

Issue #1478 reports a panic in CCExtractor 0.94 when using `-stdout` with HDHomeRun-recorded `.ts` files on Windows:

**Original crash reported:**
```
thread '' panicked at 'index out of bounds: the len is 1 but the index is 1', 
src\decoder\service_decoder.rs:275:25
```
## Reproduction

I reproduced a similar panic on Windows 0.94 portable build using `-stdout`and [sample file from #1478](https://1drv.ms/u/s!Ai_hpL0GElr6o-huZrnjSdk7hmW8Wg?e=Sv0PD6):

**My reproduction:**
```powershell
PS> $env:RUST_BACKTRACE=1
PS> .\ccextractorwinfull.exe -stdout cctest1.ts > C:\Users\Public\ccStdOut.txt

CCExtractor 0.94, Carlos Fernandez Sanz, Volker Quetschke.
Teletext decoder: Petr Kutalek
...
Opening file: cctest1.ts
thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: "Filename missing"', 
src\decoder\service_decoder.rs:897:43
```

## Root Cause Analysis

- Although the panics are reported at different line numbers across builds, they stem from the same underlying issue in the CEA-708 service decoder.

- On Windows, when running with `-stdout`, transcript output is effectively disabled or not fully initialized (no output filename, different writer setup). However, the decoder still executes code paths that assume a fully configured transcript writer.

- This results in panics in different places depending on build and execution path:

- Failed `Writer::new()` being unconditionally used (e.g. “Filename missing”)
- Accessing a non-existent writer context (index out of bounds)

- While the exact panic location varies, both failures indicate that writer-related state is accessed when transcript output is not valid.

- This fix adds an explicit check for `transcript_settings` and exits early when transcript output is disabled, preventing invalid writer usage. Behavior when transcript output is enabled remains unchanged.

## Changes Made

Modified `src/rust/src/decoder/service_decoder.rs` to add defensive checks in two methods:

### 1. `screen_print()` method:
- Check if `transcript_settings` exists before attempting Writer creation
- Return early if transcript output is disabled (instead of panicking)
- Remove `.unwrap()` call that caused the panic

### 2. `flush()` method:
- Add same defensive check for `transcript_settings`
- Return early if transcript output is not configured
- Prevent similar panic scenarios during cleanup

## Testing

**Cross-platform tested:**
- **macOS (M4)**: Successfully built and tested with the sample file provided in the issue; -stdout generates valid SRT output with no regressions observed. 
- **Linux (Ubuntu 22.04)**: Built and tested using the same approach as macOS; output is correct and unchanged.

## **Request for Windows testing**

- I was unable to test whether this fix resolves the crash on Windows, as I currently don't have access to a Windows build environment. Testing on Linux and macOS shows no regressions, but Windows-specific validation via community testing would be valuable.
- If anyone with a Windows machine can test with the [sample file from #1478](https://1drv.ms/u/s!Ai_hpL0GElr6o-huZrnjSdk7hmW8Wg?e=Sv0PD6), that would be greatly appreciated!

## Why This Fix Should Work
Even though the panic line numbers differ between the original report (275) and my reproduction (897), this fix addresses what appears to be the root cause based on available evidence:

The Fix Strategy:

- Before any Writer operations, check if transcript output is properly configured

- If not configured, skip transcript output gracefully instead of attempting to use an uninitialized Writer

- This should prevent panics at any point where Writer is accessed (creation, usage, or cleanup)

Even if this doesn't address the exact root cause of the line 275 panic (due to build differences or additional factors), the fix provides important safety guarantees:

- Prevents panics from null/missing transcript_settings

- Fails gracefully instead of crashing

At minimum, this fix makes the codebase more robust against Writer initialization issues on Windows.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -34,10 +34,10 @@ jobs:
         LLVM_CONFIG_PATH: "C:\\Program Files\\LLVM\\bin\\llvm-config"
         CARGO_TARGET_DIR: "..\\..\\windows"
         BINDGEN_EXTRA_CLANG_ARGS: -fmsc-version=0
-      run: msbuild ccextractor.sln /p:Configuration=Release-Full /p:Platform=Win32
+      run: msbuild ccextractor.sln /p:Configuration=Release-Full /p:Platform=x64
       working-directory: ./windows
     - name: Copy files to directory for installer
-      run: mkdir installer; cp ./Release-Full/ccextractorwinfull.exe ./installer; cp ./Release-Full/*.dll ./installer
+      run: mkdir installer; cp ./x64/Release-Full/ccextractorwinfull.exe ./installer; cp ./x64/Release-Full/*.dll ./installer
       working-directory: ./windows
     - name: install WiX
       run: dotnet tool install --global wix --version 4.0.0-preview.0 &&  wix extension -g add WixToolset.UI.wixext
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,107 +1,38 @@
-0.96 (2025-12-21)
+0.96 (2025-12-23)
 -----------------
 - Fix: Prevent CEA-708 decoder panic when using --stdout on Windows by safely handling writer output errors
-- New: Added --list-tracks (-L) option to list all tracks in media files without processing
-- Fix: Garbled captions from HDHomeRun and I/P-only H.264 streams (#1109)
-- Fix: Enable stdout output for CEA-708 captions on Windows (#1693)
-- Fix: McPoodle DVD raw format read/write - properly handle loop markers (#1524)
-- Fix: Variable shadowing in general_loop causing false "premature end of file" messages
-- Fix: Double-free crash in teletext cleanup when processing multiple files
-- Fix: Uninitialized memory and memory leaks found by Valgrind testing
-- Fix: Dangling pointers in Rust FFI copy_from_rust functions
-- New: Improve -out=report to show detected Teletext subtitle pages (#1034)
-- FIX: Include ATSC VCT virtual channel numbers and call signs in XMLTV output
-- FIX: Restore ATSC XMLTV generation with ETT parsing for extended descriptions, multi-segment handling, extended table ID's (EIT/VCT), corrected <programme> XMLTV formatting, buffer bounds fixes
-- Fix: DVB subtitle extraction improvements for Chinese broadcasts (#224):
-  - Fix crash in parse_PMT() due to missing bounds checks
-  - Fix negative timestamps in DVB subtitle output
-  - Fix crash in ignore_alpha_at_edge() OCR cropping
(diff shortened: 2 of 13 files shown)


cfsmp3 (approved):
## Review Summary

Thanks for this fix! The code changes look good and testing confirms the fix works correctly.

### ✅ What's Good
- Proper use of `as_ref()` for null-pointer safety
- Error handling with `if let Err(e)` instead of `.unwrap()` panic
- Consistent fix in both `screen_print()` and `flush()` methods
- All 299 Rust tests pass
- Tested with CEA-708 and CEA-608 samples - no regressions

### 📋 Required Before Merge
1. **Rebase on master** - Your branch is 12 commits behind. Run:
   ```bash
   git fetch origin
   git rebase origin/master
   ```

2. **Update CHANGES.TXT** - After rebase, move your changelog entry to the `0.96` section (version was bumped while your PR was open)

Once rebased, this is ready to merge! 🎉

x15sr71:
Thanks for the review. I’ve rebased the branch on `master` and pushed the updated commits. Waiting for CI to complete.

cfsmp3:
I can't reproduce this bug with 0.96.2 - can you still trigger it?

x15sr71:
Just tested this with `0.96.2` using `-stdout` and the [sample](https://onedrive.live.com/?id=FA5A1206BDA4E12F!586862&resid=FA5A1206BDA4E12F!586862&e=Sv0PD6&migratedtospo=true&redeem=aHR0cHM6Ly8xZHJ2Lm1zL3UvcyFBaV9ocEwwR0VscjZvLWh1WnJualNkazdobVc4V2c_ZT1TdjBQRDY&cid=fa5a1206bda4e12f), and I can’t reproduce the crash anymore.

cfsmp3:
Let's close this PR since the issue is no longer reproducible.

PR #1885, opened 2025-12-23
[FEATURE] Add guarded ASS/SSA \pos positioning for CEA-608 captions

### LINKED ISSUE #1726: [Feature Request] Precise positioning when exporting to Substation Alpha (ASS/SSA)
The current method used for exporting SubStation Alpha format uses a combination of white space characters and line breaks to position caption text. However, both SSA and ASS are capable of far more precise positioning for text, both through Margin overrides and the "\posX,Y" manual placement tag.

The current WebVTT export includes more precise positioning information using the "line" tag, but player support for placement tags is spotty. Comparatively, SubStation Alpha is supported by far more software and devices. Using similar logic to what is present in the WebVTT export code it should be possible to add precise positioning information to SSA/ASS exported subtitles.

x15sr71:

---
## Summary
Implements precise positioning for ASS/SSA subtitle export using `\pos(x,y)` tags, replacing whitespace-based layout where sufficient positioning information is available.

Fixes #1726

## Changes
- Added `ass_position_from_row_col()` function to map CEA-608 grid position (row and column) to ASS coordinates and alignment anchor
- Applied `{\anN\pos(x,y)}` positioning tags for single-region captions (1–2 adjacent rows), with alignment derived from CEA-608 PAC indent zones
- Added PlayResX/PlayResY declarations to Script Info header (384×288)
- Maintained backward compatibility via conservative guards

## Design Decisions

**Positioning is intentionally conservative:**
- Applied only when 1–2 adjacent caption rows are active to avoid layout conflicts
- Falls back to legacy whitespace behavior when layout is ambiguous (>2 rows)

**Geometry is spec-grounded:**
- Safe-area margins use approximately 3/32 of the active picture on each side (≈9.375%), reflecting historical CEA-608 hardware display geometry
- Horizontal alignment anchor (`\an1`/`\an2`/`\an3`) is derived from the CEA-608 PAC indent level (`first_col`), approximating native left-, center-, and right-aligned caption placement
- Column extents are computed as a union across all used rows for consistent 2-line anchoring
- Vertical placement uses 15 equal row slots with slot-center mapping

**Resolution choice:**
- Uses SSA default 384×288 resolution (standard for libass/Aegisub/VSFilter)
- Explicitly declared in header for spec compliance

## Testing
Verified output using a [sample](https://sampleplatform.ccextractor.org/sample/27fab4dbb603753d76e8b7701bbf0a5b1b9d41d2748759bb22f7fe6afc75597b) from the CCExtractor sample platform, comparing generated SSA against embedded 608 captions in mpv. Left-, center-, and right-aligned captions now render at approximately the correct screen position. Legacy layout is preserved for multi-region captions.

## Future Enhancements
The remaining small offset between SSA and native 608 rendering is due to libass font glyph bearing and is not resolvable at this level.

### DIFF
--- a/src/lib_ccx/ccx_encoders_ssa.c
+++ b/src/lib_ccx/ccx_encoders_ssa.c
@@ -5,6 +5,25 @@
 #include "ccx_encoders_helpers.h"
 #include "ocr.h"
 
+static void ass_position_from_row(
+    int row,
+    int play_res_x,
+    int play_res_y,
+    int *out_x,
+    int *out_y)
+{
+	// Center horizontally
+	*out_x = play_res_x / 2;
+
+	// Map CEA-608 row (0–14) to ASS Y coordinate
+	// SSA default PlayResY is 288
+	int top = play_res_y * 60 / 100; // start of lower third
+	int bottom = play_res_y * 95 / 100;
+
+	int y = top + (row * (bottom - top) / 14);
+	*out_y = y;
+}
+
 int write_stringz_as_ssa(char *string, struct encoder_ctx *context, LLONG ms_start, LLONG ms_end)
 {
 	int used;
@@ -168,11 +187,18 @@ int write_cc_buffer_as_ssa(struct eia608_screen *data, struct encoder_ctx *conte
 	unsigned h1, m1, s1, ms1;
 	unsigned h2, m2, s2, ms2;
 	int wrote_something = 0;
+	int used_row_count = 0;
 
 	int prev_line_start = -1, prev_line_end = -1;	    // Column in which the previous line started and ended, for autodash
 	int prev_line_center1 = -1, prev_line_center2 = -1; // Center column of previous line text
 	int empty_buf = 1;
-	for (int i = 0; i < 15; i++)
+
+	int first_row = -1;
+	int x, y;
+	char pos_tag[64];
+	int i;
+
+	for (i = 0; i < 15; i++)
 	{
 		if (data->row_used[i])
 		{
@@ -182,6 +208,12 @@ int write_cc_buffer_as_ssa(struct eia608_screen *data, struct encoder_ctx *conte
 	}
 	if (empty_buf)
 		return 0;
+
+	for (i = 0; i < 15; i++)
+	{
+		if (data->row_used[i])
+			used_row_count++;
+	}
 
 	millis_to_time(data->start_time, &h1, &m1, &s1, &ms1);
 	millis_to_time(data->end_time - 1, &h2, &m2, &s2, &ms2); // -1 To prevent overlapping with next line.
@@ -194,8 +226,59 @@ int write_cc_buffer_as_ssa(struct eia608_screen *data, struct encoder_ctx *conte
 	dbg_print(CCX_DMT_DECODER_608, "%s", timeline);
 
 	write_wrapped(context->out->fh, context->buffer, used);
+
+	/*
+	 * ASS precise positioning note:
+	 * We emit {\an2\pos(x,y)} using ASS script resolution coordinates.
+	 * PlayResX/PlayResY are explicitly declared in the SSA header (384x288),
+	 * which is the SSA/libass default resolution and ensures consistent
+	 * positioning across players.
+	 *
+	 * Positioning is intentionally guarded to avoid regressions when
+	 * caption layout information is ambiguous.
+	 */
+
+	/* ---- ASS precise positioning ---- */
+
+	first_row = -1;
+
+	/*
+	 * Only apply ASS positioning when:
+	 * - At least one row is present
+	 * - AND there is a single logical caption region
+	 * Otherwise, fall back to legacy SSA behavior.
+	 */
+
+	if (used_row_count > 0 && used_row_count <= 2)
+	{
+		for (i = 0; i < 15; i++)
+		{
+			if (data->row_used[i])
+			{
+				first_row = i;
+				break;
+			}
+		}
+
+		if (first_row >= 0)
+		{
+			// SSA default resolution (used by libass / Aegisub)
+			ass_position_from_row(first_row, 384, 288, &x, &y);
+
+			snprintf(
+			    pos_tag,
+			    sizeof(pos_tag),
+			    "{\\an2\\pos(%d,%d)}",
+			    x, y);
+
+			write_wrapped(context->out->fh, pos_tag, strlen(pos_tag));
+		}
+	}
+
+	/* ---- end ASS positioning ---- */
+
 	int line_count = 0;
-	for (int i = 0; i < 15; i++)
+	for (i = 0; i < 15; i++)
 	{
 		if (data->row_used[i])
 		{
--- a/src/lib_ccx/ccx_encoders_ssa.c
+++ b/src/lib_ccx/ccx_encoders_ssa.c
@@ -17,7 +17,7 @@ static void ass_position_from_row(
 
 	// Map CEA-608 row (0–14) to ASS Y coordinate
 	// SSA default PlayResY is 288
-	int top = play_res_y * 60 / 100; // start of lower third
+	int top = play_res_y * 10 / 100; // start of lower third
 	int bottom = play_res_y * 95 / 100;
 
 	int y = top + (row * (bottom - top) / 14);
--- a/src/lib_ccx/ccx_encoders_ssa.c
+++ b/src/lib_ccx/ccx_encoders_ssa.c
@@ -5,23 +5,23 @@
 #include "ccx_encoders_helpers.h"
 #include "ocr.h"
 
-static void ass_position_from_row(
+static void ass_position_from_row_col(
     int row,
+    int col,
     int play_res_x,
     int play_res_y,
     int *out_x,
     int *out_y)
 {
-	// Center horizontally
-	*out_x = play_res_x / 2;
+	/* vertical: matches FFmpeg's 0.1 + 0.0533*i formula (0.80/15 = 0.05333) */
+	double top = play_res_y * 0.10;
+	double row_step = (play_res_y * 0.80) / 15.0;
+	*out_y = (int)(top + row * row_step + 0.5);
 
-	// Map CEA-608 row (0–14) to ASS Y coordinate
-	// SSA default PlayResY is 288
-	int top = play_res_y * 10 / 100; // start of lower third
-	int bottom = play_res_y * 95 / 100;
-
-	int y = top + (row * (bottom - top) / 14);
-	*out_y = y;
+	/* horizontal: matches FFmpeg's 0.1 + 0.025*j (0.025 = 0.80/32, 32 cells) */
+	double left = play_res_x * 0.10;
+	double col_step = play_res_x * 0.025;
+	*out_x = (int)(left + col * col_step + 0.5);
 }
 
 int write_stringz_as_ssa(char *string, struct encoder_ctx *context, LLONG ms_start, LLONG ms_end)
@@ -187,11 +187,9 @@ int write_cc_buffer_as_ssa(struct eia608_screen *data, struct encoder_ctx *conte
 	unsigned h1, m1, s1, ms1;
 	unsigned h2, m2, s2, ms2;
 	int wrote_something = 0;
-	int used_row_count = 0;
 
 	int prev_line_start = -1, prev_line_end = -1;	    // Column in which the previous line started and ended, for autodash
 	int prev_line_center1 = -1, prev_line_center2 = -1; // Center column of previous line text
-	int empty_buf = 1;
 
 	int first_row = -1;
 	int last_row = -1;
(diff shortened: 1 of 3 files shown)


src/lib_ccx/ccx_encoders_ssa.c
```diff
@@ -168,11 +187,18 @@ int write_cc_buffer_as_ssa(struct eia608_screen *data, struct encoder_ctx *conte
...
 
 	int prev_line_start = -1, prev_line_end = -1;	    // Column in which the previous line started and ended, for autodash
 	int prev_line_center1 = -1, prev_line_center2 = -1; // Center column of previous line text
 	int empty_buf = 1;
-	for (int i = 0; i < 15; i++)
+
+	int first_row = -1;
+	int x, y;
+	char pos_tag[64];
+	int i;
```
cfsmp3: Why move the loop counter outside each loop, is there a reason for this old school style?

cfsmp3 (approved):
## Review Summary

Thanks for this feature! The implementation is well-designed with good safeguards. Testing confirms it works correctly.

### ✅ What's Good
- Conservative guard: only applies positioning for 1-2 row captions
- Falls back gracefully to legacy behavior for complex layouts (>2 rows)
- Uses standard SSA resolution (384×288) with explicit header declaration
- Positioning math is correct: Row 0 → y=172 (60%), Row 14 → y=273 (95%)
- Doesn't affect SRT or other output formats
- All CI checks pass, all 248 Rust tests pass

### Testing Results
Tested with sample `56c9f345...` from Sample Platform:
- 118 dialogues correctly received `{\an2\pos(x,y)}` tags
- 1 dialogue with >2 rows correctly fell back to legacy behavior
- Compared with master: changes are purely additive

### 📋 One Minor Fix Needed

**CHANGES.TXT entry is truncated:**
```
- New: Added ASS/SSA \pos-based positioning for CEA-608 captions when layout
```

Please complete the sentence, e.g.:
```
- New: Added ASS/SSA \pos-based positioning for CEA-608 captions when layout is simple (1-2 rows) (#1726)
```

Once fixed, this is ready to merge! 🎉

cfsmp3:
**Note:** The approval above is conditional on the Sample Platform CI tests passing. The tests were still queued at the time of review:

- CI - linux: pending
- CI - windows: pending

Please wait for these to complete before merging. If any tests fail, the issues should be addressed first.

src/lib_ccx/ccx_encoders_ssa.c
x15sr71: That’s on me, I initially moved the counter outside while reshuffling the loops, but there’s no good reason to keep it that way. I’ll update the PR to scope the loop counters locally and match the surrounding style.

x15sr71:
Thanks for the review. I’ve fixed the truncated changelog entry and updated the loops to use scoped variables, and pushed the updated commit.

cfsmp3 (changes requested):
The implementation looks good! The positioning is working correctly and the guard for 1-2 row captions is a sensible approach to avoid regressions.

The Windows CI failure is expected since the ASS output format changed - we'll need to update the test baselines.

One small fix needed: the CHANGES.TXT entry appears incomplete:
```
- New: Added ASS/SSA \pos-based positioning for CEA-608 captions when layout
```
Please complete this sentence (e.g., "...when layout information is available" or similar).

Once that's fixed, we can merge and update the test baselines.

x15sr71:
@cfsmp3, I've completed the `CHANGES.TXT` entry as discussed and resolved the merge conflict by keeping both entries. Please let me know if anything else needs adjustment.

cfsmp3:
Hi @x15sr71, the CHANGES.TXT entry appears to be cut off mid-sentence. Could you complete it?

x15sr71:
Hi @cfsmp3, thanks for checking again.
The current `0.96.2` changelog entry now reads:

- New: Added ASS/SSA \pos-based positioning for CEA-608 captions when layout is simple (1–2 rows) (#1726)

Please let me know if you’d prefer any wording changes, as I might be misunderstanding something.

x15sr71:
Hi @cfsmp3, quick ping in case this got buried, no rush on my side. Happy to help with any adjustments or test updates if that helps move this forward.

cfsmp3 (changes requested):
## Deep Review of PR #1885

I've thoroughly analyzed the math, built the PR, and tested with multiple CEA-608 samples.

### What Works

1. **PlayResX/PlayResY header**: Correctly added (384×288, SSA default)
2. **Single-line captions**: Positioning is correct
3. **Guard for 3+ rows**: Falls back to legacy behavior appropriately

### Issue 1: Math Error for 2-Line Captions

The code uses `first_row` with `\an2` (bottom-center alignment), which is incorrect for multi-row captions.

**The problem:**
- `\an2` places the **bottom** of the text block at the specified Y coordinate
- For a 2-line caption at rows 13-14:
  - `first_row = 13`, `y = 265`
  - The bottom line (row 14's text) ends at y=265
  - But row 14 should be at y=273
  - **Error: 8 pixels (2.8% of screen)**

**Evidence from test output:**
```
Dialogue: 0,00:00:21.98,00:00:24.23,Default,,0000,0000,0000,,{\an2\pos(192,265)}(both)\NWOW!
```
With `\an2`, "WOW!" (the bottom line, row 14) has its bottom at y=265, but row 14's calculated position is y=273.

**Fix:** Use the **last** used row instead of the first:
```c
// Current (lines 254-260):
for (i = 0; i < 15; i++)
    if (data->row_used[i]) { first_row = i; break; }

// Should be:
int last_row = -1;
for (i = 14; i >= 0; i--)
    if (data->row_used[i]) { last_row = i; break; }
```

### Issue 2: Non-Adjacent Rows Would Be Catastrophically Wrong

The code assumes `used_row_count <= 2` means a "single logical caption region" but doesn't verify the rows are adjacent.

**Worst case:** Rows 3 and 14 both used (rare but possible):
- `first_row = 3`, `y = 193`
- BOTH lines positioned with bottom at y=193
- Row 14 text would be misplaced by **80 pixels (27.8% of screen)**

**Suggested fix:** Either:
1. Check that the two rows are adjacent before applying positioning, OR
2. Only apply positioning when `used_row_count == 1`

### Issue 3: Changelog Entry in Wrong Version

The changelog entry is added under `0.96.2 (2025-12-26)` which is a past release. It should be moved to the current development version section (0.96.5 or whatever is current at merge time).

### Row-to-Y Mapping Reference

For reviewers, here's the complete mapping with PlayResY=288:
```
row 0  → y=172 (59.7%)    row 8  → y=229 (79.5%)
row 1  → y=179 (62.2%)    row 9  → y=236 (81.9%)
row 2  → y=186 (64.6%)    row 10 → y=244 (84.7%)
row 3  → y=193 (67.0%)    row 11 → y=251 (87.2%)
row 4  → y=200 (69.4%)    row 12 → y=258 (89.6%)
row 5  → y=208 (72.2%)    row 13 → y=265 (92.0%)
row 6  → y=215 (74.7%)    row 14 → y=273 (94.8%)
row 7  → y=222 (77.1%)
```

### Summary

The feature concept is good and addresses issue #1726, but the positioning math needs correction before merging. The error is small for adjacent rows (~3%) but could be severe for edge cases.

x15sr71:
Thanks for the detailed review and for the helpful fix suggestions.

I’ve applied the suggested fixes to the ASS positioning logic (anchor correction + row adjacency guard) and did a small cleanup around variable initialization. I validated the changes using this [sample](https://sampleplatform.ccextractor.org/sample/56c9f345482c635f20340d13001f1083a7c1913c787075d6055c112fe8e2fcaa) from the Sample Platform. I’ve attached the newly generated .ass file [here](https://drive.google.com/file/d/1oBWcQ24iv7kWOgz34A1xh69b_DeuQnVm/view?usp=sharing) for review and moved the changelog entry to the correct section. Let me know if anything else needs adjusting!

cfsmp3 (changes requested):
Thanks for the updates! I tested this again and found a positioning bug.

## Test Setup

Using sample file `27fab4dbb603753d76e8b7701bbf0a5b1b9d41d2748759bb22f7fe6afc75597b.ts` (Friends episode):

```bash
# Generate SSA
./ccextractor 27fab4dbb603753d76e8b7701bbf0a5b1b9d41d2748759bb22f7fe6afc75597b.ts -out=ssa -o test.ssa

# Play with embedded captions
mpv 27fab4dbb603753d76e8b7701bbf0a5b1b9d41d2748759bb22f7fe6afc75597b.ts

# Play with generated SSA
mpv 27fab4dbb603753d76e8b7701bbf0a5b1b9d41d2748759bb22f7fe6afc75597b.ts --sub-file=test.ssa
```

## Issue

The positioning is totally different between mpv playing the embedded captions in the .ts file and the generated .ssa captions:

- **Embedded captions**: Appear at the **top** of the screen
- **Generated SSA**: Appear **vertically centered** in the screen

## Root Cause

The positioning formula maps all rows to the "lower third" (60-95% of screen height):

```c
int top = play_res_y * 60 / 100;    // 172 (60% down from top)
int bottom = play_res_y * 95 / 100;  // 273 (95% down)
```

This means CEA-608 row 0 maps to y=172, which is 60% down the screen (vertical center). But row 0 in CEA-608 represents the **top** of the screen.

## Fix

CEA-608 rows 0-14 span the full screen height, so the mapping should cover the full vertical range:

```c
int top = play_res_y * 5 / 100;     // ~14 (5% from top)
int bottom = play_res_y * 95 / 100;  // ~273 (95% from top)
```

This correctly maps:
- Row 0 → y ≈ 14 (near top)
- Row 14 → y ≈ 273 (near bottom)

x15sr71:
Thanks for the review and for suggesting the fix.

After expanding the row-to-Y mapping, I tested 5%, 10%, and 12% top margins. With bottom-anchored ASS positioning (`\an2`), 5% caused clipping for top-row captions; 10% avoids clipping while preserving top-caption intent. I also verified that captions with 3+ rows continue to fall back to the legacy layout. Screenshots are attached [here](https://drive.google.com/drive/folders/1OORkUOtqP9nYXmyBvXMQalARlowfaww4?usp=sharing) for reference.

yukichigai:
I'd say 10% would be the most "correct" margin to use, at least if you're trying to recreate the positioning as it would be displayed by most EIA-608 capable display devices. Older devices using chips like the Zilog Z86229 had a top and bottom margin of 9.375% each (3/32), while newer authoring software seems to have simplified that to 10% top and bottom each.

yukichigai:
On that note, horizontal positioning should probably also have some padding if you want to recreate native 608 display methods, but which margin to use there is a little less easy to decide. Older chips like the Zilog left a full 1/6th margin on each side (16.6666...%), but newer devices and authoring software use a uniform 10% margin all four sides. Even more maddening is the fact that different captions seem to be created with different horizontal margins in mind: some are clearly positioned with the 1/6th margin in mind, others for 10%.

x15sr71:
Hi @cfsmp3, just a gentle ping in case this slipped through. I’ve applied the suggested fixes and updated the branch accordingly. No rush at all, whenever you have time, I’d really appreciate your feedback. Thanks!

cfsmp3 (changes requested):
Thanks for the work on this, and for the 10% top margin fix which @yukichigai validated.

However, after testing with `mpv` and comparing the SSA output against the embedded 608 captions, there's a **horizontal positioning problem**: the PR always centers captions horizontally, but the actual 608 captions are often left-aligned.

## The issue

The `ass_position_from_row()` function hardcodes `*out_x = play_res_x / 2` (center), and all positioning tags use `\an2` (bottom-center anchor). This means every caption is horizontally centered regardless of where it actually appears in the 608 grid.

CEA-608 captions are positioned on a 32-column grid and can start at any column — they're frequently left-aligned, not centered.

## Visual comparison

I tested with sample `27fab...7b.ts` using `mpv --sub-file=<file>` and pressing `j` to switch between embedded 608 and the SSA output.

**Embedded 608 — captions are bottom-left:**

mpv-shot0003

mpv-shot0004

**SSA output from this PR — same scenes, captions are centered:**

mpv-shot0005

mpv-shot0006


## What needs to change

The vertical (Y) positioning looks good with the 10% top margin. The problem is horizontal (X) positioning: the code needs to account for the CEA-608 column position rather than always centering. Options:

1. Use `\an1` (bottom-left anchor) when text starts on the left side of the 608 grid, `\an3` (bottom-right) for right-aligned, `\an2` for centered — based on the actual column offset in the row data
2. Compute X from the actual starting column of the text in the 32-column grid

The goal is that the SSA output should approximate the same screen position as the embedded 608 captions.

x15sr71:
Thanks for the deep review and for suggesting the fixes. I've updated the code so that the SSA output better approximates the same screen position as the embedded CEA-608 captions.

The horizontal coordinate is now calculated from the actual starting column of the caption text. This starting column is determined by scanning the CEA-608 character buffer using `find_limit_characters()` across the used rows.

A few related adjustments were made along with this:

- `\an7` (top-left anchor) is used so that `\pos(x,y)` represents the top-left corner of the caption block, rather than the default bottom-center alignment which previously introduced a horizontal offset
- Vertical row mapping added to match FFmpeg's formula (`0.1 + 0.0533*i`, equivalent to `PlayResY * 0.80 / 15` per row step)

**Note on drift** — A small rightward drift can still be observed for captions positioned toward the far right of the 608 grid. This appears to be a side effect of mapping the fixed 32-column CEA-608 grid into the ASS safe area using constant step sizes. The same behavior can also be observed when generating SSA output with FFmpeg (`ffmpeg -f lavfi -i "movie=27fab........97b.ts[out+subcc]" -map s -c:s ass ffmpeg.ssa`), so this seems to be an inherent limitation of the grid → ASS coordinate mapping rather than something specific to this implementation.

Screenshots comparing native 608, CCExtractor SSA, and FFmpeg SSA output at the same timestamp, [here.](https://drive.google.com/drive/folders/130E8sUuUx4EQiUKs4RQpEOsZ2ScvKh7i?usp=sharing)

**Note on font size:** The current `Fontsize 20` in the SSA style header can cause glyphs to extend slightly beyond the nominal row slots derived from the 608 grid mapping, depending on font metrics and renderer behavior. Changing it would affect all SSA output including the legacy fallback path, so it's left unchanged in this PR.

PR #1905, opened 2025-12-25
[FIX] Fix MSVC warning C4098: void function returning a value in networking.c

VS7686:

- [x ] I have checked that another pull request for this purpose does not exist.

- [x ] I have used CCExtractor just a couple of times.

---

**Description**
This PR fixes multiple instances of compiler warning C4098 in `src/lib_ccx/networking.c`.
Several functions (`connect_to_srv`, `net_send_header`, etc.) were declared as `void` but were attempting to return values from internal function calls.

**Changes**
- Casted the return values of the internal Rust functions to (void) to suppress MSVC warning C4098.

- Removed the explicit return; statements to ensure the subsequent C implementation is correctly executed (fixing a logic error   where the C fallback was being skipped).

**Environment**
- Windows 11
- MSVC / CMake

### DIFF
--- a/src/lib_ccx/networking.c
+++ b/src/lib_ccx/networking.c
@@ -104,7 +104,7 @@ void connect_to_srv(const char *addr, const char *port, const char *cc_desc, con
 {
 #ifndef DISABLE_RUST
 	(void)ccxr_connect_to_srv(addr, port, cc_desc, pwd);
-
+	return;
 #endif
 	if (NULL == addr)
 	{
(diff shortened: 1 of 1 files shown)


cfsmp3 (changes requested):
Thank you for the fix! However, there's a logic issue.

The original `return ccxr_xxx();` was intentional - it calls Rust and returns early so the C fallback doesn't execute. Your fix for `net_check_conn` is correct:
```c
ccxr_net_check_conn();
return;
```

But the other 3 functions are missing the `return;`, so both Rust AND C code will execute when Rust is enabled.

Please add `return;` after the Rust calls in:
- `connect_to_srv`
- `net_send_header`
- `net_send_epg`

VS7686:
Thanks for the review! I've added the missing return; statements to connect_to_srv, net_send_header, and net_send_epg as requested. Ready for re-review.

PR #1921, opened 2025-12-28
IMPROVEMENT: 1. simplified verify_parity function. 2.Improved documentat…

ChubbyChipmunk77:
…ion for public function validate_cc_pair. 3. Added constant for 0x7F.

**[IMPROVEMENT]**

**In raising this pull request, I confirm the following**

**My familiarity with the project is as follows:**

---
##Changes Made

1. Replaced magic number 0x7F with named constant `CC_BLANK_SPACE`.
2. Added comprehensive documentation to function `validate_cc_pair`.
3. Simplified verify_parity function.

##Tests
✅ All existing tests passed.

### DIFF
--- a/src/rust/src/lib.rs
+++ b/src/rust/src/lib.rs
@@ -273,7 +273,6 @@ extern "C" fn ccxr_process_cc_data(
 ///      If this fails, the entire pair is deemed corrupt, and the function returns false.
 ///    - Sanitization: Validates parity for the first data byte (cc_block[1]).
 ///      If this fails (but byte 2 was valid), cc_block[1] is overwritten with CC_SOLID_BLANK (0x7F).
-
 const CC_SOLID_BLANK: u8 = 0x7F;
 
 pub fn validate_cc_pair(cc_block: &mut [u8]) -> bool {


ChubbyChipmunk77:
> Thank you for the improvements! The `verify_parity` simplification and the documentation are helpful.
> 
> However, there's a semantic issue with the constant usage. The value `0x7F` is used for two different purposes in this code:
> 
>     1. **As the solid blank character** (0x7F = DEL/solid blank in CEA-608) - your constant is correct here
> 
>     2. **As a parity bit mask** (`& 0x7F` strips bit 7) - this should NOT use `CC_SOLID_BLANK`
> 
> 
> In `do_cb_dtvcc`, the expression `(cc_block[1] & 0x7F) == 0` is masking off the parity bit to check if the 7 data bits are zero. This isn't related to the "solid blank" character - it's just that the mask happens to have the same value.
> 
> Please revert the changes in `do_cb_dtvcc` (lines ~315-316 in your diff) to keep `& 0x7F` as a literal, or create a separate `PARITY_MASK` constant. The constant usage in `validate_cc_pair` is correct.
> 
> Also, consider using `///` for Rust doc comments instead of `//`.

That makes sense. It is definitely confusing to use the blank character constant when performing bit masking. I’ll define a separate PARITY_MASK constant for this usage. Resolving the documentation comments too.

PR #1928, opened 2025-12-29
[FIX] Silence unused-but-set variable warning in zvbi decoder

AS-pngg:

---

{pull request content here}
This PR fixes a compiler warning (`-Wunused-but-set-variable`) by explicitly
marking an intentionally unused local variable in `zvbi/decoder.c`.

There are no functional changes.


cfsmp3 (changes requested):
If it's unused why not just remove it?
Clean up is fine, but just hiding warnings by adding unexplainable code is not a good idea :-)

AS-pngg:
> If it's unused why not just remove it? Clean up is fine, but just hiding warnings by adding unexplainable code is not a good idea :-)

Thanks for pointing this out.

I initially considered removing the variable, but it is still required for the assignment call (vbi3_raw_decoder_set_sampling_par(...)) where the side effect is needed, even though the returned value is not used later.

I agree that removing unused code is preferable where possible — if there’s a better approach you’d recommend here (e.g., refactoring the call or handling the return value differently), I’m happy to update the patch.

cfsmp3 (changes requested):
Instead of silencing the warning with `(void)`, please remove the variable entirely:

```c
// Remove this line:
vbi_service_set service_set;

// And change this:
service_set = vbi3_raw_decoder_set_sampling_par(rd3, (vbi_sampling_par *)rd, /* strict */ 0);
(void)service_set;

// To just:
vbi3_raw_decoder_set_sampling_par(rd3, (vbi_sampling_par *)rd, /* strict */ 0);
```

The function is called for its side effects; there's no need to capture the return value if we're not using it.

AS-pngg:
Done I’ve removed the unused variable entirely and updated the call to rely only on the function’s side effects, as suggested.
Thanks for the clear explanation!

cfsmp3:
Hi @AS-pngg, any update on this? To clarify: rather than using `(void)val;` to silence the warning, please remove the variable entirely since it's not used. This is cleaner than keeping unused variables around.

PR #1949, opened 2026-01-01
Prevent integer overflow in EIA-608 screen buffer reallocation

### LINKED ISSUE #1948: Integer overflow in EIA-608 screen buffer allocation may lead to heap buffer overflow
### Description 

Component: EIA-608 decoder
src/lib_ccx/ccx_decoders_608.c
Function: write_cc_buffer, write_cc_line

### The Problem

The EIA-608 decoder dynamically grows its screen buffer using realloc based on the number of subtitle screens processed:

```c
(sub->nb_data + 1) * sizeof(struct eia608_screen)
```

However, there is no check for integer overflow. If a malformed or crafted input triggers a very high number of screen updates, this calculation can wrap around, resulting in a very small buffer allocation. Subsequent writes then cause a heap buffer overflow, potentially crashing the program or corrupting memory.

### The Proposed Fix

Added an overflow guard:
```c
if (sub->nb_data + 1 > SIZE_MAX / sizeof(struct eia608_screen))
```
Calculated allocation size explicitly using size_t new_size
Ensured safe handling of realloc failure (log and return without corrupting memory)
This prevents the heap overflow caused by integer overflow while keeping behavior unchanged for normal input.

### Impact

Processing malformed EIA-608 subtitle data will no longer result in:
- Heap memory corruption
- Program crashes

THE-Amrit-mahto-05:

- [x] I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).

---

[FIX] Prevent integer overflow in EIA-608 screen buffer allocation

### Description

This PR fixes a potential integer overflow in the EIA-608 decoder, which could lead to a heap buffer overflow when processing malformed subtitle input.

**Component:** EIA-608 decoder  
**File:** `src/lib_ccx/ccx_decoders_608.c`  
**Functions:** `write_cc_buffer`, `write_cc_line`  

### Problem

The screen buffer was being reallocated using:

```c
(sub->nb_data + 1) * sizeof(struct eia608_screen)
```
without checking for integer overflow. If sub->nb_data is extremely large (due to a malformed or crafted subtitle), this multiplication could wrap around, causing a tiny buffer allocation and heap overflow on subsequent writes.

### Fix

- Added an overflow guard before allocation:
```c
if (sub->nb_data + 1 > SIZE_MAX / sizeof(struct eia608_screen))
```
- Calculated allocation size explicitly using size_t new_size
- Ensured safe handling of realloc() failure (log error and return safely)

### Impact
- Prevents heap memory corruption
- Prevents program crashes
- Keeps normal behavior unchanged

Fixes #1948

### DIFF
--- a/src/lib_ccx/ccx_decoders_608.c
+++ b/src/lib_ccx/ccx_decoders_608.c
@@ -316,10 +316,20 @@ int write_cc_buffer(ccx_decoder_608_context *context, struct cc_subtitle *sub)
 
 	if (!data->empty && context->output_format != CCX_OF_NULL)
 	{
-		struct eia608_screen *new_data = (struct eia608_screen *)realloc(sub->data, (sub->nb_data + 1) * sizeof(*data));
+		size_t new_size;
+
+		if (sub->nb_data + 1 > SIZE_MAX / sizeof(struct eia608_screen))
+		{
+			ccx_common_logging.log_ftn("Too many screens, cannot allocate more memory.\n");
+			return 0;
+		}
+
+		new_size = (sub->nb_data + 1) * sizeof(struct eia608_screen);
+
+		struct eia608_screen *new_data = (struct eia608_screen *)realloc(sub->data, new_size);
 		if (!new_data)
 		{
-			ccx_common_logging.log_ftn("No Memory left");
+			ccx_common_logging.log_ftn("Out of memory while reallocating screen buffer\n");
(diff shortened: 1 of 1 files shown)


cfsmp3:
Closing as superseded by #1953, which includes all the commits from this PR.

PR #1953, opened 2026-01-01
Fix/ts heap overflow

THE-Amrit-mahto-05:
**[FIX]**

- [x] I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).

---

### Description

Component: Transport Stream (TS) handling  
src/lib_ccx/ts_functions.c  
Function: copy_payload_to_capbuf

### The Problem

The function `copy_payload_to_capbuf` grows a capture buffer using:

```c
newcapbuflen = cinfo->capbuflen + payload->length;
```
However, there was no check for integer overflow before this addition.
If a very large payload->length is combined with capbuflen, the sum can wrap around, resulting in a very small allocation passed to realloc.
This can cause a heap buffer overflow, potentially crashing the program or corrupting memory.

### The Proposed Fix
- Promoted the newcapbuflen variable to int64_t to handle large sums safely.
- Added an explicit overflow guard:
```c
if (payload->length > INT64_MAX - cinfo->capbuflen)
{
    mprint("Error: capbuf size overflow\n");
    return -1;
}
```

- Ensured safe allocation using size_t in the realloc call.
- Logged errors and returned safely if allocation fails.

### Impact

- Prevents heap memory corruption.
- Prevents crashes caused by malformed TS data.
- Maintains normal functionality for valid input.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -6,6 +6,7 @@
 - Fix: 32-bit build failures on i686 and armv7l architectures
 - Fix: Legacy command-line argument compatibility (-1, -2, -12, --sc, --svc)
 - Fix: Prevent heap buffer overflow in Teletext processing (security fix)
+- Fix: Prevent integer overflow leading to heap buffer overflow in Transport Stream handling (security fix)
 - Fix: Lazy OCR initialization - only initialize when first DVB subtitle is encountered
 - Build: Optimized Windows CI workflow for faster builds
 - Fix: Updated GUI with version 0.7.1. A blind attempt to fix a hang on start on some Windows.


THE-Amrit-mahto-05:
@cfsmp3 

Thanks for the detailed review!

I’ve addressed all the requested changes:

Removed the incorrect (int) casts and now assign newcapbuflen directly to capbufsize and capbuflen (both int64_t as defined in ccx_demuxer.h).

Added the security fix entries to docs/CHANGES.TXT documenting the Teletext and Transport Stream overflow fixes.

The remaining checks are currently running. Please let me know if you’d like any further adjustments once CI completes.

cfsmp3 (approved):
Thanks for the quick fix! The casts are removed and CHANGES.TXT is updated.

Build passes, tests show no regression. 

Minor note: The comments still say "capbufsize is int" but the struct uses `int64_t` - not a blocker, just FYI.

Approving and merging.

cfsmp3:
@THE-Amrit-mahto-05 I've merged this, but TBH, since we're migrating to Rust, I don't think it's worth spending time on these things, particularly if they are purely theoretical problems...

Issue #1965, opened 2026-01-02
Critical Teletext Decoder Vulnerabilities: Out-of-Bounds Read/Write and Loop Overflow


THE-Amrit-mahto-05:
I have discovered a cluster of critical and previously unreported vulnerabilities in CCExtractor's Teletext decoder (telxcc.c).

**Vulnerabilities Identified:**

1. **Out-of-Bounds Reads in tlt_process_pes_packet**
   - Cause: Minimal PES packet size checks allow truncated packets to trigger reads past the buffer end.
   - Impact: Heap/stack corruption, undefined behavior.
   - Trigger: Malformed DVB-TS Teletext streams with truncated PES packets.

2. **Out-of-Bounds Write in payload reversal loop**
   - Cause: `data_unit_len` can exceed remaining PES packet size.
   - Impact: Heap corruption, memory overwrite.
   - Trigger: Malformed Teletext data unit length.

3. **Potential Infinite Loop / Loop Counter Overflow**
   - Cause: `i` was declared as `uint16_t` and can wrap-around for large PES packets.
   - Impact: Infinite loop or memory corruption.
   - Trigger: Large or malformed PES packets.

**Affected Files/Functions:**
- `telxcc.c` → `tlt_process_pes_packet`

**Proposed Fixes (ready to implement in a PR):**
- Added minimum PES packet size checks before processing headers.
- Ensure `data_unit_len` does not exceed remaining packet length.
- Changed loop counter `i` to `uint32_t` to prevent wrap-around.
- Added debug messages for oversized data units for safer decoding.

**Impact if Unpatched:**
- Heap or stack corruption
- Decoder crashes or undefined behavior
- Security vulnerabilities for malicious Teletext streams

cfsmp3:
This is not critical unless you provide an example file that shows the problem (fine if you create it yourself, but it must hit the code path that shows the problem).

I'm going to close all these small issues with theoretical problems - they add a lot of overhead to my workload.

PR #1966, opened 2026-01-02
fix: Critical Teletext Decoder Vulnerabilities: Out-of-Bounds Read/Write and Loop Overflow 

THE-Amrit-mahto-05:

---

### **Description**

**Issues Fixed:**
1. **Out-of-Bounds Read** in `tlt_process_pes_packet`: minimal size checks on PES header allowed truncated packets to trigger reads past buffer end.
2. **Out-of-Bounds Write** in payload reversal loop: a malformed packet could specify `data_unit_len` causing a write past buffer end.
3. **Potential Infinite Loop / Buffer Overflow** due to `uint16_t` wrap-around of the loop index.

**Fixes Implemented:**
- Changed loop index `i` from `uint16_t` → `uint32_t` to avoid wrap-around.
- Added size checks to ensure PES packet is long enough before reading header fields.
- Added boundary check for `data_unit_len` to prevent OOB writes.
- Logging added for truncated or malformed packets.

**Impact:**
- Prevents crashes, buffer over-reads, and memory corruption.
- Ensures Teletext decoder robustness against malformed DVB-TS streams.
- Normal Teletext streams continue to work with no regression.

**Testing:**
- Verified with truncated PES packets: safely ignored.
- Verified payload length exceeding buffer: OOB writes prevented.
- Verified loop termination even with max uint16_t values.

**Issue Reference:** #1965

### DIFF
--- a/src/lib_ccx/telxcc.c
+++ b/src/lib_ccx/telxcc.c
@@ -1392,7 +1392,7 @@ int tlt_process_pes_packet(struct lib_cc_decode *dec_ctx, uint8_t *buffer, uint1
 	uint8_t pes_ext_flag;
 	// extension
 	uint32_t t = 0;
-	uint16_t i;
+	uint32_t i;
 	struct TeletextCtx *ctx = dec_ctx->private_data;
 	ctx->sentence_cap = sentence_cap;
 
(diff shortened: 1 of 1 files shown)


cfsmp3:
Closing - all commits are included in #1968.

PR #1984, opened 2026-01-05
fix: add NULL checks to Rust FFI exports

### LINKED ISSUE #1983: Potential memory safety issues in Rust FFI exports (NULL pointer handling)
### Description

While going through the Rust FFI layer, I noticed a few places where raw pointers coming from C are used without NULL checks. This can lead to undefined behavior if these functions are ever called with invalid inputs.

I’m still learning Rust FFI, so please correct me if I’m wrong, but based on my understanding these look like genuine safety issues.

### What I found

ccxr_verify_crc32
- buf is used without a NULL check
- len is not checked for negative values before creating a slice

ccxr_levenshtein_dist and ccxr_levenshtein_dist_char
- Pointer arguments are assumed to be non-NULL

Functions in libccxr_exports/bitstream.rs
- Multiple exported FFI functions dereference *mut bitstream without checking for NULL (via copy_bitstream_c_to_rust)

THE-Amrit-mahto-05:

- [ ] I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).

---

### Description

While reviewing the Rust FFI layer, I noticed several exported functions that
dereference raw pointers coming from C without checking for `NULL`.  
If these functions are ever called with invalid inputs, this can lead to
undefined behavior.

I’m still learning Rust FFI, so please correct me if I’m misunderstanding
anything, but based on my understanding these look like genuine safety issues.

### What this PR changes

- Adds defensive `NULL` checks to Rust FFI exports in:
  - `libccxr_exports/mod.rs`
  - `libccxr_exports/bitstream.rs`
- Adds a negative-length check to `ccxr_verify_crc32`
- Ensures FFI functions return safe defaults (`0` or `NULL`) instead of causing UB
- Adds tests to verify safe behavior when called with `NULL` pointers

### Testing

- Ran `cargo test` in `src/rust`
- All existing and newly added tests are passing

fixed #1983

### DIFF
--- a/src/rust/src/libccxr_exports/bitstream.rs
+++ b/src/rust/src/libccxr_exports/bitstream.rs
@@ -848,7 +848,10 @@ mod bitstream_copying_tests {
             assert!(super::ccxr_next_bytes(std::ptr::null_mut(), 1).is_null());
             assert!(super::ccxr_read_bytes(std::ptr::null_mut(), 1).is_null());
             assert_eq!(super::ccxr_bitstream_get_num(std::ptr::null_mut(), 1, 0), 0);
-            assert_eq!(super::ccxr_read_exp_golomb_unsigned(std::ptr::null_mut()), 0);
+            assert_eq!(
+                super::ccxr_read_exp_golomb_unsigned(std::ptr::null_mut()),
+                0
+            );
             assert_eq!(super::ccxr_read_exp_golomb(std::ptr::null_mut()), 0);
             assert_eq!(super::ccxr_read_int(std::ptr::null_mut(), 8), 0);
         }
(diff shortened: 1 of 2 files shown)


cfsmp3:
I don't think this is needed.

- 1) If you don't have a bitstream that functions won't be called, the whole thing will fall apart way before getting there.
- 2) Because the unsafely derives from C and the C code is going away, all this FFI stuff is going to be removed anyway.

Also, we can't really have safeguards at the entry point of every function. There's invariants that are guaranteed by the logic (and verified with unit tests), so checking everything everywhere just adds a lot of bloat but doesn't really improve anything.

Issue #1993, opened 2026-01-07
Memory Leak: File buffer not freed, leaking 16MB per demuxer instance


THE-Amrit-mahto-05:
### Description
Discovered a memory leak in CCExtractor: the file buffer allocated by the demuxer is never freed. This results in approximately 16MB of memory being leaked for each demuxer instance.

### Steps to Reproduce
1. Run CCExtractor on a video file.
2. Observe memory usage of the process.
3. Repeat with multiple files or instances.
4. Memory usage continues to grow and is not released.

### Impact
- Long-running or batch operations can consume large amounts of memory.
- Could cause crashes or slow performance on low-memory systems.
- This is a real, current issue affecting stability, not a futuristic concern.

### Suggestion
- Free the allocated file buffer after demuxing completes.
- Add tests to verify that buffers are properly released.

cfsmp3:
Please don't open a ticket for these things - PRs are OK, but issues are for users to report problems. If a developer sees a problem he/she just sends a PR.

PR #1994, opened 2026-01-07
Free filebuffer to prevent 16MB memory leak in ccxr_demuxer_close

THE-Amrit-mahto-05:

---

### Description

fixes a critical memory leak in CCExtractor where the file buffer allocated for each demuxer instance is never freed, causing  16MB of memory to leak per instance. This occurs because Rust allocated the buffer and then "forgot" it so C code could use it, leaving it unmanaged.

### Changes made
- In `ccxr_demuxer_close`, the raw file buffer pointer is reclaimed as a boxed slice and dropped to allow Rust to free the memory.
- The pointer is then set to `NULL` to prevent double-free issues if C code tries to clean it up.
- Updated comments to explain why this step is necessary.

Fixes: #1993

### Verification
- Verified with multiple demuxer instances that memory usage no longer grows by 16MB per instance.
- Ensured normal demuxer functionality is unaffected.


THE-Amrit-mahto-05:
@cfsmp3  pls review this too

cfsmp3:
> @cfsmp3 pls review this too

There's no need to tag me in every PR, I review them when I can... tagging me just adds noise, no need.

THE-Amrit-mahto-05:
@cfsmp3 
Sorry for the delay in raising this PR. I spent a bit of extra time understanding how Rust memory management works and how it interacts with C, as well as getting familiar with the CI/testing workflow.

The PR now makes sure that Rust-allocated memory is properly freed in ccxr_demuxer_delete for all code paths, which prevents memory leaks and avoids undefined behavior when Rust and C interact as you suggested

Issue #1583, opened 2023-12-14
[PROPOSAL]: Snap package ready for `ccextractor`


soumyaDghosh:
I have created a snap package for the `ccextractor`. The main enhancement by this, is the fact that it can be used in any distro now, even if it's a non-LTS Ubuntu release, or an Arch based distro or anything else. 

Also, if I can get some help in testing the package, that'd also be helpful.

# In raising this issue, I confirm the following:

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that the bug-fix I am reporting can be replicated, or that the feature I am suggesting isn't already present.
- [x] I have checked that the issue I'm posting isn't already reported.
- [x] I have checked that the issue I'm porting isn't already solved and no duplicates exist in [closed issues](https://github.com/CCExtractor/ccextractor/issues?q=is%3Aissue+is%3Aclosed) and in [opened issues](https://github.com/CCExtractor/ccextractor/issues)
- [x] I have checked the pull requests tab for existing solutions/implementations to my issue/suggestion.
- [x] I have used the latest available version of CCExtractor to verify this issue exists.
- [x] I have ticked all the boxes in this section and to prove it I'm deleting the section completely to remove boilerplate text.

SparklinStar:
I can help you with testing the package in Fedora 39.

SP-XD:
Create a appImage its better than a snap package

cfsmp3:
@soumyaDghosh Did you send a PR? What do you need from us? 
In any case this should be, I suppose, a GitHub workflow that creates the snap package automatically as needed.

Akki2005:
Hi, I’m interested in helping with this.

From the discussion, it sounds like the preferred direction would be:
- an automated GitHub Actions workflow
- that builds and publishes a Snap package
- rather than a manually maintained Snap

Before proceeding, I wanted to confirm:
1) Is Snap packaging still desired long-term?
2) Should this target stable releases only, or every commit?
3) Would a similar workflow for AppImage also be welcome, or should we focus on one format?

Happy to draft a proposal once direction is confirmed.

x15sr71:
Hi all, just a quick follow-up on this proposal.

I’ve opened PR [#1998](https://github.com/CCExtractor/ccextractor/pull/1998), which implements Snap packaging for CCExtractor based on this discussion.
[...]

PR #1998, opened 2026-01-08
[FEATURE]: Add Snap packaging support with Github workflow

x15sr71:

- [ ] I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).

**My familiarity with the project:**

---

## Summary

As proposed in #1583, this PR adds Snapcraft-based packaging for CCExtractor, along with a GitHub Actions workflow that builds the Snap and uploads the resulting `.snap` artifact to GitHub Releases.

**Benefits:**  
- Cross-distribution Linux support  
- Self-contained runtime dependencies  
- No impact on existing build systems or packaging flows  

---

## Implementation Details

- **`snap/snapcraft.yaml`**  
  Snapcraft configuration using the existing CMake build system. Runtime dependencies (FFmpeg, GPAC, Tesseract, etc.) are bundled explicitly via `stage-packages`.

- **`snap/local/run-ccextractor.sh`**  
Lightweight runtime wrapper that ensures bundled shared libraries are resolved from within the Snap at execution time, avoiding accidental linkage against host system libraries. While the Snap may work without this wrapper on some environments, it makes runtime behavior deterministic and consistent across distributions.

- **`.github/workflows/build_snap.yml`**  
GitHub Actions workflow completes successfully and produces a `.snap` package, which is published as a [CI artifact](https://github.com/x15sr71/ccextractor/actions/runs/20799849300) (zipped by GitHub Actions).
---

## Testing

- GitHub Actions workflow completes successfully and produces a `.snap` package ([CI run artifact](https://github.com/x15sr71/ccextractor/actions/runs/20799849300)).

- Local runtime verification performed on:
  - Ubuntu 22.04 (Jammy)
  - Ubuntu 24.04 (Noble)

```bash
   sudo snap install ./ccextractor_*.snap --classic --dangerous
   ccextractor --version
```
- The release-triggered workflow uses the same job definition and artifact upload logic as the manually verified CI run, so the generated `.snap` artifact should be attached to GitHub Releases as well.


x15sr71:
@cfsmp3, Thanks for the thorough testing and for flagging this issue.

I’ve updated the runtime wrapper to explicitly shift the first argument passed by Snap, as suggested in option 2, so the binary path is no longer forwarded as an input file. With this change, subtitle extraction works as expected. I rebuilt and installed the snap locally and verified it using:

```
sudo snap install ./ccextractor_*.snap --classic --dangerous
ccextractor /path/to/sample.ts -o output.srt

```
The input is now parsed correctly and the output is generated as expected. Please let me know if you’d like me to test additional cases or make any further adjustments.

cfsmp3:
## Build Issue: `libgpac-dev` not available in core22

When building the snap with `sudo snapcraft --destructive-mode`, the build fails:

```
Cannot find package listed in 'build-packages': libgpac-dev
```

### Root Cause

Ubuntu 22.04 (core22) doesn't have `libgpac-dev` in the standard repositories. Your local build worked because you likely have GPAC installed system-wide or from a PPA, but the snap build environment is clean.

### Suggested Fix

Build GPAC from source as a snap part. Add a `gpac` part that builds before ccextractor:

```yaml
parts:
  gpac:
    plugin: cmake
    source: https://github.com/gpac/gpac.git
    source-tag: v2.4.0
    cmake-parameters:
      - -DCMAKE_INSTALL_PREFIX=/usr
      - -DGPAC_STATIC=OFF
    build-packages:
      - zlib1g-dev
      - libssl-dev
    prime:
      - usr/lib/*/libgpac*
      - usr/include/gpac

  ccextractor:
    after: [gpac]  # Ensures gpac is built first
    plugin: cmake
    source: .
    source-subdir: src
    # ... rest of your config, but remove libgpac-dev from build-packages
    # ... and remove libgpac11 from stage-packages (gpac part provides it)
```

This is the standard approach for snap packages when a dependency isn't available in the base repos.

Please rebuild and test after making this change.

x15sr71:
@cfsmp3, Thanks for the review and the detailed feedback on the GPAC integration. I followed the suggested snap-part approach for GPAC, but wasn’t able to get a build in `core22`, below is a detailed summary of the investigation.

Before attempting `GPAC` integration, I fixed multiple build and environment mismatches to ensure the Snap pipeline itself was sound:

1. **Rust toolchain mismatch**
   The default Rust available in `core22` was insufficient for the current codebase. I resolved this by installing and pinning a newer Rust toolchain via `rustup` during the Snap build.

2. **CMake version mismatch**
   The system CMake in the base image was outdated. I fixed this by using the `cmake/latest/stable` snap, which allowed the CMake configuration to proceed correctly.

3. **Source layout / CMake entry point**
   I adjusted the Snap build configuration to correctly locate and invoke `src/CMakeLists.txt`, which resolved initial configuration and path errors.

With these fixes in place, the CCExtractor build pipeline itself is stable and reproducible.

---

I then attempted to satisfy the `GPAC` dependency using several approaches, but each one fails for several reasons.

#### 1. Building `GPAC` from source (shared and static)

* **Attempt:** Build `GPAC` `v2.4.0` from source inside the Snap `core22` environment.
* **Result:**

  * **Shared build:** Fails at link time with unresolved symbols (e.g. `gf_mixer_*`). GPAC’s audio/filter components appear to assume system audio backends (ALSA/OSS) that do not link cleanly in this environment.
  * **Static build (`--static-bin`):** Compiles successfully, but does not install the required development headers (e.g. `gpac/isomedia.h`) or generate a `gpac.pc` file. As a result, CCExtractor cannot detect or link against GPAC.
  * **Mocking / manual symbol stubbing:** I evaluated this as a last resort, but rejected it because it would require injecting fake symbols or bypassing internal GPAC logic, producing a fragile binary with undefined runtime behavior.

#### 2. Disabling `GPAC`

* **Attempt:** Build CCExtractor without `GPAC` to ship a stable but feature-reduced Snap.
* **Result:** The build fails immediately with:

  ```
  The following required packages were not found: - gpac
  ```
* **Root cause:** The upstream `CMakeLists.txt` appears to treat `GPAC` as a `REQUIRED` dependency on Linux.
  I considered patching the build scripts to remove this requirement, but decided against it, as forcing the build in this way could result in broken references or runtime crashes and would not be safe to ship.

---

At this point, it appears to be:

- `GPAC` cannot be built and consumed cleanly in the Snap `core22` environment without invasive or unsafe hacks.
- `GPAC` cannot be disabled because the current CCExtractor CMake logic treats it as mandatory.

Possible options I see are:

1. Make `GPAC` truly optional in the CMake logic, with consistent feature guards and header usage across CCExtractor, so the project can be built cleanly without `GPAC` when required (e.g. for distros or Snap), while preserving full MP4 support when it is available.
2. Pause Snap work until GPAC’s build/install behavior becomes more modular.

If there’s a recommended GPAC configuration for Snap that I’ve overlooked, I’m happy to try that. I may still be missing something, but after tracing this through the build system and multiple `GPAC` configurations, I haven’t found a clean or safe path forward.

Please let me know how you’d like me to continue. Thanks for your time.

x15sr71:
Update:

As a follow-up, I also tried integrating GPAC 2.5 in the core22 Snap environment. I built GPAC from source in both static and shared configurations, including reduced/headless builds, and wired it into the Snap using staged artifacts and explicit include/library paths.

GPAC itself builds successfully, but CCExtractor is still unable to consume it cleanly in core22. In practice, this shows up as missing installed headers (for example gpac/isomedia.h), or, n the case of static builds—linker failures related to X11/XCB dependencies. Addressing these would likely require fairly invasive build workarounds or maintaining a custom GPAC fork, which doesn’t feel like the right direction for Snap packaging.

I wanted to updated that GPAC 2.5 was also tested. Please let me know how you’d like me to proceed.

cfsmp3:
## Update: We've identified and fixed the issues blocking this PR

Thanks for working on this! We've done some investigation and found the issues that were blocking the snap build. Here's what needs to happen:

### 1. Rebase on PR #2042 (or wait for it to merge)

PR #2042 fixes a linking issue in the CMakeLists.txt files that affects all Linux builds (not just snap). Without this fix, the build fails with:
```
undefined reference to `decode_vbi'
```

**Option A:** Wait for #2042 to be merged, then rebase your branch on master.

**Option B:** Rebase your branch on the `fix/cmake-rust-linking` branch now:
```bash
git fetch origin
git rebase origin/fix/cmake-rust-linking
```

### 2. Update `snap/snapcraft.yaml` to build GPAC from source

Since `libgpac-dev` isn't available in Ubuntu 22.04 (core22) repositories, we need to build GPAC from source as part of the snap. Add this `gpac` part **before** the `ccextractor` part:

```yaml
parts:
  gpac:
    plugin: make
    source: https://github.com/gpac/gpac.git
    source-tag: abi-16.4
    build-packages:
      - build-essential
      - pkg-config
      - zlib1g-dev
      - libssl-dev
      - libfreetype6-dev
      - libjpeg-dev
      - libpng-dev
    override-build: |
      set -eux
      # Standard GPAC build - same as normal CCExtractor builds
      ./configure --prefix=/usr
      make -j$(nproc)
      make DESTDIR=$SNAPCRAFT_PART_INSTALL install-lib
      # Fix gpac.pc prefix to point to stage directory for build time
      sed -i "s|^prefix=.*|prefix=$SNAPCRAFT_STAGE/usr|" $SNAPCRAFT_PART_INSTALL/usr/lib/pkgconfig/gpac.pc
    stage:
      - usr/lib/libgpac*
      - usr/lib/pkgconfig/gpac.pc
      - usr/include/gpac

  ccextractor:
    after: [gpac]    # <-- ADD THIS LINE to ensure GPAC builds first
    # ... rest of your existing ccextractor part ...
```

**Key points:**
- Use GPAC tag `abi-16.4` (corresponds to version 2.5-DEV)
- Use `make install-lib` (not `make install`) - this installs headers and the pkg-config file
- The `sed` command fixes the gpac.pc prefix so pkg-config can find the headers during the ccextractor build
- Add `after: [gpac]` to the ccextractor part

### 3. Add build environment for pkg-config

In your `ccextractor` part, add this build environment so CMake can find GPAC:

```yaml
  ccextractor:
    after: [gpac]
    plugin: nil
    source: .
    build-environment:
      - PKG_CONFIG_PATH: "$SNAPCRAFT_STAGE/usr/lib/pkgconfig:$PKG_CONFIG_PATH"
```

### 4. Update `snap/local/run-ccextractor.sh` to include GPAC library path

The wrapper script needs to include `$SNAP/usr/lib` in `LD_LIBRARY_PATH` because that's where GPAC's libraries are installed. Change the `LD_LIBRARY_PATH` line to:

```bash
export LD_LIBRARY_PATH="$SNAP/usr/lib:\
$SNAP/usr/lib/$LIB_TRIPLET:\
$SNAP/usr/lib/$LIB_TRIPLET/blas:\
$SNAP/usr/lib/$LIB_TRIPLET/lapack:\
$SNAP/usr/lib/$LIB_TRIPLET/pulseaudio:\
${LD_LIBRARY_PATH:+:$LD_LIBRARY_PATH}"
```

Note the addition of `$SNAP/usr/lib:\` at the beginning.

### 5. Use a fixed version string

Change `version: git` to a fixed version like `version: '0.96.5'` because the git-generated version string can exceed 32 characters, which causes snap packaging to fail.

---

### Summary of all changes needed:

1. Rebase on #2042 (or wait for merge)
2. Add `gpac` part to snapcraft.yaml (builds GPAC from source)
3. Add `after: [gpac]` and `build-environment` with `PKG_CONFIG_PATH` to ccextractor part
4. Add `$SNAP/usr/lib` to LD_LIBRARY_PATH in run-ccextractor.sh
5. Change `version: git` to `version: '0.96.5'`

We've tested this configuration and it builds and runs successfully. Let us know if you have any questions!

x15sr71:
Thank you so much for taking the time to identify the root causes and provide such clear, actionable steps. I've implemented all the changes you outlined. The snap now builds cleanly on `core22` and runs correctly in my local testing.

Thanks again for fixing this and for helping identify the right approach. Happy to make any further changes if needed.

PR #1999, opened 2026-01-08
fix prevent FFI memory leaks in demuxer sync

THE-Amrit-mahto-05:

---

### Summary

 fixes multiple memory leaks and unsafe pointer overwrites in the Rust–C FFI demuxer synchronization logic.

The issues occur when Rust demuxer state is copied into C structures repeatedly (e.g. during resets), causing existing C-owned pointers to be overwritten without being freed.

---

### Problem

1. In `copy_demuxer_from_rust_to_c`, existing pointers in `PID_buffers` and `PIDs_programs` were overwritten without freeing the previous allocations, leading to cumulative memory leaks during demuxing.

2. `PSIBuffer::default()` initialized `buffer` with a dummy heap allocation instead of `NULL`, which:
   - Creates a small but persistent leak
   - Makes ownership and validity ambiguous on the C side

Both issues become more visible when the demuxer is reset or reused multiple times.

---

### Fixes

- **demuxer.rs**
  - Free existing C pointers before overwriting them
  - Add NULL checks before calling `Box::from_raw` to avoid undefined behavior

- **common_types.rs**
  - Initialize `PSIBuffer::buffer` with `std::ptr::null_mut()` instead of a 1-byte allocation
  - Clarify Rust ↔ C ownership semantics

---

### Impact
- Prevents repeated memory leaks during long-running demuxing
- Makes FFI ownership rules explicit and safer
- No functional behavior change expected

---

### Verification
- Built CCExtractor successfully with the changes
- `cargo test` passes in `src/rust`
- Manual demuxing on sample input shows no regression

### DIFF
--- a/src/rust/src/demuxer/common_types.rs
+++ b/src/rust/src/demuxer/common_types.rs
@@ -109,7 +109,9 @@ impl Default for PSIBuffer {
     fn default() -> Self {
         PSIBuffer {
             prev_ccounter: 0,
-            buffer: Box::into_raw(Box::new(0u8)),
+            // Initialize with null to avoid unnecessary heap allocations and
+            // signal that the buffer is currently empty.
+            buffer: std::ptr::null_mut(),
             buffer_length: 0,
             ccounter: 0,
         }
--- a/src/rust/src/libccxr_exports/demuxer.rs
+++ b/src/rust/src/libccxr_exports/demuxer.rs
@@ -98,59 +98,73 @@ pub unsafe fn copy_demuxer_from_rust_to_c(c_demuxer: *mut ccx_demuxer, rust_demu
     c.global_timestamp_inited = rust_demuxer.global_timestamp_inited.millis() as c_int;
 
     // PID buffers - extra defensive version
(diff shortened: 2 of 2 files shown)


cfsmp3 (approved):
## Review: ✅ Approved

Fixes multiple memory leaks in Rust-C FFI demuxer synchronization.

**Changes:**
1. **demuxer.rs**: Frees existing C pointers before overwriting them in `copy_demuxer_from_rust_to_c()`
2. **common_types.rs**: Initializes `PSIBuffer::buffer` with `null_mut()` instead of dummy heap allocation

**Analysis:**
The fix is correct. When syncing state from Rust to C repeatedly, the old code would overwrite C-side pointers without freeing them first, causing cumulative memory leaks. The new code properly frees existing allocations before creating new ones.

The POISON_PTR_PATTERN check (0xcdcdcdcd) is a defensive measure against uninitialized memory in debug builds.

**Testing:**
- ✅ Builds successfully
- ✅ Tested with sample - works correctly

**Verdict:** Ready to merge.

PR #2008, opened 2026-01-10
fix(matroska): abort parsing on invalid EBML ID to prevent infinite loop

pulk17:
[FIX] matroska: Abort parsing on invalid element ID (0xFFFFFFFF) to prevent infinite loops

---

### Summary
This PR fixes a critical regression where the legacy C Matroska parser (`matroska.c`) enters an infinite loop when encountering invalid EBML IDs (specifically `0xFFFFFFFF`) or EOF conditions inside segment/cluster loops.

### The Issue
In `parse_segment`, `parse_segment_cluster`, and related functions, the loop structure was:
1. Read EBML ID.
2. If ID is unknown, log warning and `skip_bytes`.
3. If the ID is `0xFFFFFFFF` (EOF/Error), `skip_bytes` skips 0 bytes.
4. The loop repeats indefinitely, consuming 100% CPU and flooding logs.

### The Fix
Added an explicit check for `code == 0xFFFFFFFF` in the parsing loops. The parser now detects this "Invalid ID" state and aborts the loop gracefully with a specific error message.

### Verification
Tested against corrupted/truncated MKV samples that previously caused hangs.
* **Before:** Infinite loop, process hangs.
image

* **After:** Logs "Invalid EBML ID... Aborting segment parsing" and exits successfully.
image
image

### DIFF
--- a/src/lib_ccx/matroska.c
+++ b/src/lib_ccx/matroska.c
@@ -126,11 +126,11 @@ void parse_ebml(FILE *file)
 
 		// Safety check to prevent infinite loop on invalid EBML ID
 		if (code == 0xFFFFFFFF)
-    	{
-	        mprint(MATROSKA_WARNING "Invalid EBML ID 0xFFFFFFFF encountered at position " LLD ". Aborting segment parsing to prevent infinite loop.\n", 
-            	   get_current_byte(file) - 4);
-        	break;
-    	}
+		{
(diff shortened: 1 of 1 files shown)


cfsmp3:
Already fixed (note though that this PR is fixing the symptom but not the underlying problem)

PR #2009, opened 2026-01-10
fix(ts): allow manual streamtype override for private streams

pulk17:
[FIX] ts: Allow manual streamtype override for private streams causing fatal errors

---

### Summary
Fixed a logic bug in `ts_tables.c` where the parser would trigger a fatal error ("I can't tell the stream type") for private streams (0x80-0xFF), even if the user explicitly provided `--streamtype`.

### The Issue
When parsing a PMT with private streams (common in M2TS files with PGS subtitles), the code checks if the stream type is private and immediately exits with a fatal error requesting manual selection.

The error logic failed to check if `ccx_options.demux_cfg.ts_forced_streamtype` was already set by the user via the CLI arguments. This made it impossible to force-read private streams, as the application would crash before applying the user's override.

### The Fix
Added a check to verify if the user has already manually set the stream type (via `--streamtype`).
* If the user **has** provided a type, the fatal error is skipped, a verbose message is logged ("User manually set stream type..."), and the parser proceeds using the user-supplied value.
* If the user **has not** provided a type, the fatal error triggers as intended.

### Verification
Tested with `hitman_cut.m2ts` (a sample with missing PMT/Private Stream) and `00000.m2ts` (PGS streams).

**Command:**
`ccextractor input.m2ts --datapid 4608 --streamtype 6`

* **Before:**
   
Screenshot 2026-01-10 192702

* **After:**
    
Screenshot 2026-01-10 194020


cfsmp3:
I see "no captions were found" after.

Each PR needs to show a real life improvement.

Do you have any sample in which by doing this we are now able to extract captions?
A private stream is private for a reason, it's not hidden captions (or at least, I haven't seen that ever, but maybe).

cfsmp3:
Closing — no response to the January request for a real-life use case. If you have a concrete scenario where this is needed, feel free to reopen with that context.

PR #2015, opened 2026-01-10
[FIX] rust: add defensive length check to validate_cc_pair

Harsh-Sahu43:

---

### Summary

Adds a defensive guard to `validate_cc_pair` to ensure the function only
processes CC blocks of the expected length (3 bytes).

This prevents potential out-of-bounds access if the function is ever
called with malformed input. Behavior for valid CC blocks is unchanged.

### Details

- Returns `false` early when the CC block slice length is not exactly 3
- Adds unit tests covering invalid slice lengths
- No changes to existing logic for valid CEA-608 / CEA-708 data

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,3 +1,10 @@
+0.96.6 (unreleased)
+-------------------
+- Fix: Clear status line output on Linux/WSL to prevent text artifacts (#2017)
+- Fix: Prevent infinite loop on truncated MKV files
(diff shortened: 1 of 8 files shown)


cfsmp3:
The `format_rust` CI check is failing due to formatting issues. Please run:

```bash
cd src/rust && cargo fmt
```

This will fix the missing spaces before `{` in the code:
- `if cc_block.len() != 3{` → `if cc_block.len() != 3 {`
- `fn test_validate_cc_pair_invalid_length(){` → `fn test_validate_cc_pair_invalid_length() {`

Once the formatting is fixed, this should be ready to merge.

PR #2019, opened 2026-01-14
fix(matroska): prevent out-of-bounds NAL parsing in AVC/HEVC blocks

Akki2005:
This patch hardens Matroska AVC/HEVC parsing by making NAL length extraction safe on malformed or truncated input.

Previously, NAL length was read via a direct cast to a long pointer, which could result in:
- unaligned memory access
- undefined behavior on some architectures
- out-of-bounds reads when handling truncated frames

The updated logic:
- verifies that enough bytes are available before reading the NAL length
- validates the parsed NAL size before consuming the payload
- exits parsing gracefully on malformed data

Behavior for valid Matroska streams remains unchanged.


Akki2005:
Applied clang-format to address the formatting check failure.
No functional changes.

Issue #2023, opened 2026-01-16
[BUG]: Segmentation fault in report-only(-out=report) mode when processing AVC/H.264 streams with Rust decoder


x15sr71:
### Summary
CCExtractor crashes with a segmentation fault when running in report-only mode (`-out=report`) on files containing `AVC/H.264` video streams when the Rust AVC decoder is enabled (default in recent builds).

### Command used
```
./ccextractor -out=report input.ts
```
Where `input.ts` is any transport stream file containing [H.264 video](https://sampleplatform.ccextractor.org/sample/7236304cfcfce141c7cec31647c1268a3886063390ce43c2f71188c70f5494c4).

### Segmentation fault
```
Opening file: 7236304cfcfce141c7cec31647c1268a3886063390ce43c2f71188c70f5494c4.ts
Analyzing data in general modetream
[1]    97365 segmentation fault  ccextractor  -out=null
```

### Expected behavior
CCExtractor should generate a report without crashing, as report generation does not require video decoding.

### Actual behavior
Segmentation fault (SIGSEGV) at address `0x50` when attempting to dereference `encoder_ctx->timing` inside the Rust AVC decoder.

### Root cause
The crash occurs because:

1. **Report-only mode** (`-out=report`) does not initialize video decoding structures, specifically:
   - `encoder_ctx->timing` remains `NULL`
   - `lib_cc_decode->avc_ctx` may remain `NULL`

2. **The Rust AVC decoder** (`ccxr_process_avc`) assumes these structures are always initialized and dereferences them unconditionally

3. **The C→Rust boundary** in `process_avc()` does not validate these preconditions before calling into Rust

### Technical details

**Crash location(lldb):**
```
Thread 1: EXC_BAD_ACCESS (code=1, address=0x50)
Frame: ccx_rust::avc::core::process_avc
Instruction: ldr x0, [x8, #0x50]  ; loading enc_ctx.timing
```

**Call stack:**
```
process_avc (avc_functions.c)
  └─ ccxr_process_avc (Rust FFI)
       └─ ccx_rust::avc::core::process_avc
            └─ ccxr_set_fts(enc_ctx.timing)  ← NULL dereference
```
## Proposed Solution

Add a guard at the C-Rust boundary in `process_avc()` that skips AVC processing when the decoder isn't fully initialized (like in report-only mode).

PR #2024, opened 2026-01-16
Fix: Prevent AVC segfault in report-only mode (`-out=report`)

x15sr71:

---

### Summary

This PR fixes a segmentation fault that occurs (as reported in issue #2023) when running CCExtractor in report-only mode (`-out=report`) on files containing AVC (H.264) video streams with the Rust AVC decoder enabled.

---

### Problem

When running:

```bash
./ccextractor -out=report input.ts
```
on transport streams containing AVC/H.264 video, CCExtractor crashes with a segmentation fault.

### Solution
Add a defensive guard at the C → Rust FFI boundary in `process_avc()` to skip AVC processing when the decoder context isn't fully initialized (such as in report-only mode).

### Guard Conditions
Skip calling the Rust AVC decoder if any of the following are true:

- `encoder_ctx` is NULL
- `encoder_ctx->timing` is NULL
- `lib_cc_decode` is NULL
- `lib_cc_decode->avc_ctx` is NULL

In these cases, `process_avc()` simply returns without invoking the Rust decoder.

## Rationale

- The fix belongs at the C → Rust boundary, where assumptions about initialization must be enforced
- Only affects `process_avc()`; no changes to Rust internals
- Report-only mode does not require AVC frame decoding.
- Prevents similar crashes if decoding state isn't initialized elsewhere

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 0.96.6 (unreleased)
 -------------------
+- Fix: Prevent segmentation fault when running in report-only mode (-out=report) on AVC/H.264 streams.
 - Fix: Prevent infinite loop on truncated MKV files
 - Fix: Various memory safety and stability fixes in demuxers (MP4, PS, MKV, DVB)
 - Fix: Delete empty output files instead of leaving 0-byte files (#1282)


cfsmp3:
Thank you for working on this fix! However, I'm closing this PR in favor of #2025, which takes a simpler approach that doesn't break existing tests.

## Why this PR breaks tests

The guard conditions are too aggressive:

```c
if (!enc_ctx || !enc_ctx->timing || !dec_ctx || !dec_ctx->avc_ctx)
```

This checks 4 conditions, but only `!enc_ctx` is needed for the report-only mode fix. The additional checks (`!enc_ctx->timing`, `!dec_ctx->avc_ctx`) reject legitimate processing scenarios:

1. **`!enc_ctx->timing`** - This can be NULL in certain initialization states during normal processing, not just report mode
2. **`!dec_ctx->avc_ctx`** - Same issue; this blocks AVC processing even when it should proceed

The CI shows 10+ test failures including `--autoprogram`, `--startcreditstext`, and other options that have nothing to do with report mode.

## The simpler fix (#2025)

PR #2025 adds the guard at the Rust FFI boundary with minimal checks:

```rust
if enc_ctx.is_null() {
    return avcbuflen;
}
```

This precisely targets the report-only mode case (where `enc_ctx` is NULL because no encoder is created) without affecting any other code paths.

## Location consideration

The fix belongs at the Rust FFI boundary (inside `ccxr_process_avc`) rather than in the C wrapper, because:
- The Rust code is what dereferences the pointers
- Rust's safety guarantees should be enforced at the Rust boundary
- It keeps the C code clean and unchanged

Thank you again for identifying this issue and submitting a fix! Your contribution helped highlight the problem.

x15sr71:
Thanks for the review and explanation. I see the issue now, I was checking too many things when only the `enc_ctx NULL` case mattered for report mode. The extra checks on timing and `avc_ctx` ended up blocking legitimate scenarios during normal processing also putting the guard in the Rust FFI function is cleaner too, since that's where the crash actually happens. Appreciate you fixing it properly and explaining why.

PR #2027, opened 2026-01-17
Fix unsafe temporary Vec pointer usage in userdata.rs

THE-Amrit-mahto-05:

---

### Description
While working on the Rust FFI code in userdata.rs, I noticed that a few calls were passing pointers created from temporary Vecs (to_vec().as_mut_ptr()) directly into C functions.

This is unsafe because the temporary Vec is dropped immediately after the statement, which can leave the C side with a dangling pointer and result in undefined behavior.

### Fix
- Keep copied Vecs alive for the duration of the corresponding FFI calls where needed
- Reuse existing slice pointers when the C code immediately copies the data

This change does not alter behavior and is limited to fixing lifetime safety at the FFI boundary.

### Proposed Changes
- Store the Vec in a variable so it remains alive until after the dump() call completes
- Since `cc_data` is already a slice returned from `read_bytes()`, use its pointer directly.  
  The C function `store_hdcc()` copies the data using `memcpy()` (see `sequencing.c:70`), so the pointer only needs to be valid during the call
- Store the Vec in a variable so it remains alive until after the decode_vbi() call completes

### DIFF
--- a/src/rust/src/es/userdata.rs
+++ b/src/rust/src/es/userdata.rs
@@ -278,7 +278,8 @@ pub unsafe fn user_data(
 
                 if !proceed {
                     debug!(msg_type = DebugMessageFlag::VERBOSE; "\rThe following payload is not properly terminated.");
-                    dump(cc_data.to_vec().as_mut_ptr(), (cc_count * 3 + 1) as _, 0, 0);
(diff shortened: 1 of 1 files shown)


cfsmp3:

PR #2028, opened 2026-01-17
Fix GXF VBI out-of-bounds panic in userdata.rs

THE-Amrit-mahto-05:

---

### Desciption

critical out-of-bounds panic when processing malformed GXF VBI data.

Previously, the code attempted to slice exactly 720 bytes from the stream, even if fewer bytes were available:
```rust
let vbi_data = &ustream.data[ustream.pos..ustream.pos + 720];
```
This caused the program to panic if the GXF packet was truncated, corrupted, or attacker-crafted. The check for udatalen < 720 only logged a message but did not prevent execution.

### Fix
- Return early (Ok(1)) when insufficient data (<720 bytes) is detected.
- This prevents out-of-bounds panics and allows the decoder to continue processing subsequent packets.
- Existing functionality for valid packets is unchanged.

We could consider returning a specific error variant instead of Ok(1) to indicate that the packet is malformed. This would allow higher level code to handle malformed packets explicitly (e.g., loging, skipping, or applying custom processing). The current approach is safe and ensures decoder stability.

### DIFF
--- a/src/rust/src/es/userdata.rs
+++ b/src/rust/src/es/userdata.rs
@@ -532,6 +532,7 @@ pub unsafe fn user_data(
 
         if udatalen < 720 {
             info!("MPEG:VBI: Minimum 720 bytes in luma line required");
+            return Ok(1); // Early return - not enough data
         }
 
         let vbi_data = &ustream.data[ustream.pos..ustream.pos + 720];


cfsmp3:

PR #2030, opened 2026-01-17
Fix Dish Network caption out-of-bounds panics in userdata.rs

THE-Amrit-mahto-05:

---

### Description 

Prevents critical crashes when processing malformed or truncated Dish Network caption streams.
Previously, the code performed unchecked slicing based on pattern_type, which could panic if insufficient data was available.

The old comment “4 to 6 bytes” was outdated. Based on the actual patterns, the required range is now 4 to 12 bytes.

This fix ensures the decoder continues gracefully with subsequent packets without crashing.

### Fix
- Added a single bounds check before the match pattern_type block.
- Determine the number of bytes required for each pattern:
  - Pattern 0x02 → 4 bytes
  - Pattern 0x04 → 5 bytes
  - Pattern 0x05 → 12 bytes (accesses indices 0–11)
- If insufficient data is available:
  -  Log a message indicating the insufficient bytes
  - Return early with Ok(1) to skip the malformed packet
- Updated the slice for pattern 0x05 from 10 → 12 bytes to safely access all required indices.

Returning Ok(1) skips malformed packets while preserving decoder stability, and all valid packets continue processing as before.

### DIFF
--- a/src/rust/src/es/userdata.rs
+++ b/src/rust/src/es/userdata.rs
@@ -335,9 +335,25 @@ pub unsafe fn user_data(
         debug!(msg_type = DebugMessageFlag::PARSE; "DN  ID: {:02X}  Count: {:5}  Unknown: {:04X}  Pattern: {:X}",
                   id, dishcount, something, pattern_type);
 
-        // The following block needs 4 to 6 bytes starting from the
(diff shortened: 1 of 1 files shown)


cfsmp3:

PR #2031, opened 2026-01-18
Fix use-after-free bugs in Rust userdata handling

THE-Amrit-mahto-05:

---

### Description
While reviewing userdata.rs, I found another instance of the same use-after-free pattern in the Unrecognized user data handling.

The current code creates a temporary Vec using .to_vec() and immediately extracts a raw pointer from it using .as_mut_ptr().
Because the temporary Vec is dropped right after .as_mut_ptr(), the pointer passed to dump() becomes dangling, resulting in use-after-free undefined behavior.

This is the same issue that was fixed earlier in this file.

### Fix
Store the Vec in a local variable so its lifetime extends until after the dump() call:
```rust
let mut data_copy =
    ustream.data[ustream.pos..ustream.pos + dump_len].to_vec();
dump(data_copy.as_mut_ptr(), dump_len as _, 0, 0);
```
This guarantees the backing memory remains valid for the duration of the call.

### DIFF
--- a/src/rust/src/es/userdata.rs
+++ b/src/rust/src/es/userdata.rs
@@ -342,6 +342,10 @@ pub unsafe fn user_data(
         let dcd_pos = ustream.pos; // dish caption data position
         match pattern_type {
             0x02 => {
+                if ustream.data.len() - ustream.pos < 4 {
+                    info!("Dish Network caption: insufficient data");
+                    return Ok(1);
+                }
                 // Two byte caption - always on B-frame
                 // The following 4 bytes are:
                 // 0  :  0x09
@@ -389,6 +393,10 @@ pub unsafe fn user_data(
                 // Ignore 3 (0x0A, followed by two unknown) bytes.
             }
             0x04 => {
+                if ustream.data.len() - ustream.pos < 5 {
+                    info!("Dish Network caption: insufficient data");
+                    return Ok(1);
+                }
                 // Four byte caption - always on B-frame
                 // The following 5 bytes are:
                 // 0  :  0x09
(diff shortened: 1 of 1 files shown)


cfsmp3:
Thanks for finding these issues! However, I noticed you have 4 open PRs that all modify the same file (`src/rust/src/es/userdata.rs`):

- #2027 - Fix unsafe temporary Vec pointer usage
- #2028 - Fix GXF VBI out-of-bounds panic
- #2030 - Fix Dish Network caption out-of-bounds panics
- #2031 - Fix use-after-free bugs (this PR)

**These will conflict with each other** when we try to merge them sequentially.

Could you please **consolidate all 4 fixes into a single PR**? This will:
1. Avoid merge conflicts
2. Make it easier to review all the `userdata.rs` security fixes together
3. Ensure all fixes work correctly with each other

You can either:
- Update one of these PRs with all the fixes and close the others, OR
- Create a new PR that combines everything

Also note: #2027 and #2031 fix the same use-after-free issues but with slightly different approaches. When consolidating, please pick one consistent approach.

Thanks!

THE-Amrit-mahto-05:
@cfsmp3 

I’ve consolidated all the related fixes into a single PR on fix/rust-userdata-uaf as suggested. This PR now includes:
a) Use-after-free fixes at the Rust–FFI boundary using a consistent lifetime safe approach
b) Dish Network caption out-of-bounds guards
c) GXF VBI length checks to prevent panics

I’ve kept one consistent approach across all changes and will close the other PRs to avoid conflicts.

I want to clarify that I never intended to submit separate PRs for the same file. While going through the code, I noticed issues at different points, so I raised PRs as I found them. I apologize for the confusion, and I’ll make sure to consolidate fixes properly in a single PR next time.

PR #2040, opened 2026-01-18
fix SEI payload size handling and type correctness

THE-Amrit-mahto-05:

---
### Description

`payload_type` and `payload_size` were previously inferred or explicitly defined as `i32`, which required repeated casting when indexing buffers or comparing against slice lengths. This added unnecessary casts and could theoretically lead to incorrect behavior for large payload values.

### Fix
- Changed `payload_type` and `payload_size` to explicit `usize`.
- Removed unnecessary `as i32` casts during accumulation.
- Removed redundant `as usize` casts during indexing and bounds checks.
- Ensured `seibuf_idx` arithmetic remains type-safe and consistent.


THE-Amrit-mahto-05:
@cfsmp3 
Thanks 
I’ve updated  followings:
Changed payload_type and payload_size to u32
Kept explicit as usize casts only where required for indexing and slicing.
Left the logic unchanged, focusing purely on type correctness

THE-Amrit-mahto-05:
@cfsmp3 
Some Rust tests are already failing on master. Since my change only adjusts type correctness (i32 → u32) without altering logic or control flow, it shouldn’t affect test behavior.

PR #2046, opened 2026-01-19
fix(rust): harden ccxr_process_cc_data against excessive cc_count

Harsh-Sahu43:

---

{pull request content here}
### Summary

Adds a defensive upper bound on `cc_count` in `ccxr_process_cc_data` to prevent
excessive allocations or misuse at the FFI boundary.

### Details
- Rejects calls where `cc_count` exceeds a sane limit
- Logs a warning instead of attempting allocation
- Adds a unit test to validate the guard
- No behavior change for valid inputs

### DIFF
--- a/src/rust/src/lib.rs
+++ b/src/rust/src/lib.rs
@@ -394,8 +394,8 @@ pub extern "C" fn ccxr_dtvcc_set_active(dtvcc_ptr: *mut std::ffi::c_void, active
 /// data should point to cc_data of length cc_count
 /// dec_ctx.dtvcc_rust must point to a valid DtvccRust instance
 
+/// Maximum number of CC blocks accepted per call to prevent excessive allocation
 const MAX_CC_BLOCKS_PER_CALL: c_int = 10_000;
-
 #[no_mangle]
 extern "C" fn ccxr_process_cc_data(
     dec_ctx: *mut lib_cc_decode,


cfsmp3 (changes requested):
const MAX_CC_BLOCKS_PER_CALL: c_int = 10_000;

Where did you get this number from? 
Do you have any sample in which this is an actual problem?

Harsh-Sahu43:
Good point — you’re right to question the constant.

Per CEA-708/ATSC, CC bandwidth is very small (~9.6 kbps total), so under normal conditions only a few bytes of CC data appear per frame [CTA-708 on Wikipedia](https://en.wikipedia.org/wiki/CTA-708). That said, this PR wasn’t based on a failing sample, and the 10_000 value is not spec-derived.

The motivation was defensive hardening: ccxr_process_cc_data sits at the C→Rust boundary and allocates proportional to cc_count, which in some demuxer paths is derived as len / 3 without validation.

I agree a fixed numeric limit is arbitrary. I can switch this to overflow / buffer-size guarding instead of a fixed limit.
Happy to adjust based on what you think is the right direction here.

cfsmp3:
God, I'm definitely not going to be dealing with this AI slop.

Harsh-Sahu43:
Understood — thanks for the feedback. I agree this change isn’t justified without a concrete failure case, and I’ll adjust my approach accordingly.

I’ll focus future contributions on well-scoped, test-backed issues aligned with existing CCExtractor patterns.

PR #2063, opened 2026-01-26
[FIX] Guard against NULL timing context when reporting decoder statistics

hridyasadanand:
[FIX] Guard against NULL timing context when reporting decoder statistics

In raising this pull request, I confirm the following (please check boxes):

- [x] I have read and understood the contributors guide.
- [ ] I have mentioned this change in the changelog.

My familiarity with the project is as follows (check one):

---

### Summary

This PR adds a defensive null check for dec_ctx->timing in the decoder reporting loop at the end of start_ccx().

Under certain execution paths, decoder contexts present in ctx->dec_ctx_head may not have an associated timing context. The current code assumes timing is always present and unconditionally dereferences dec_ctx->timing, which can lead to undefined behavior or crashes.

This change safely skips such decoder entries during reporting only.

### Problem Description

The reporting loop in start_ccx() assumes that every decoder in ctx->dec_ctx_head has a valid timing structure.

While decoders created via init_cc_decode() always initialize timing, decoder contexts created via copy_decoder_context() explicitly set timing = NULL. These copied decoder contexts may still be present in dec_ctx_head, making the assumption invalid.

### Fix

Add a simple guard before accessing timing fields:

```c
if (!dec_ctx->timing)
    continue;

### CODE REVIEW

**Comment by hridyasadanand** [comment]
Hi, I checked the CI logs on this PR.

The failures seem to come from the test runner, not from the code change itself. In both Linux and Windows runs I’m seeing repeated errors like:

Empty path name is not legal / Path is empty
followed by exit code 10 for some entries.

On Windows the tester is invoked with --tempfolder "", which looks like it’s passing an empty path to the tester. The CCExtractor binary builds fine and most tests run successfully.

Since this PR only adds a NULL check in the reporting loop and doesn’t touch decoding or file output, it doesn’t look like the change is causing this.

Let me know if you want me to dig more into the CI/tester side.

**Comment by cfsmp3** [comment]
Thank you for the defensive programming effort, but after code analysis, this fix appears to address a scenario that cannot actually occur.

## Analysis

**All decoders in `dec_ctx_head` come from `init_cc_decode()`:**

Looking at every place where decoders are added to `dec_ctx_head` in `lib_ccx.c`:

```c
// Line 322-325:
dec_ctx = init_cc_decode(ctx->dec_global_setting);
list_add_tail(&(dec_ctx->list), &(ctx->dec_ctx_head));

// Line 376-379: same pattern
// Line 384-387: same pattern
```

**`init_cc_decode()` always initializes timing (lines 291-293 in ccx_decoders_common.c):**

```c
ctx->timing = init_timing_ctx(&ccx_common_timing_settings);
if (!ctx->timing)
    fatal(EXIT_NOT_ENOUGH_MEMORY, "In init_cc_decode: Out of memory initializing timing.");
```

If `init_timing_ctx()` fails, the program exits fatally. So any decoder created via `init_cc_decode()` will always have a non-NULL `timing`.

**`copy_decoder_context()` does set `timing = NULL`, but those copies are NOT added to `dec_ctx_head`:**

```c
// dvb_subtitle_decoder.c:1820
dec_ctx->prev = copy_decoder_context(dec_ctx);
```

The copied context is stored in `dec_ctx->prev`, not added to the decoder list.

## Conclusion

There is no code path where a decoder with `timing = NULL` can end up in `dec_ctx_head`. The reporting loop in `start_ccx()` will only ever see decoders with valid timing contexts.

If you have a crash log or reproduction case showing this null pointer dereference actually occurring, please share it and we can revisit this.

PR #2078, opened 2026-01-31
Fix FFI allocator mismatch by properly handling Rust/C ownership

THE-Amrit-mahto-05:

---
### Description 

As discussed on Zulip, I had a question about memory ownership in stream_functions.rs. Some pointers coming from C (like the demuxer pointer) were being wrapped in Box::from_raw, which assumes Rust owns them and will free them. The allocation actually happens on the C side, which could cause undefined behavior (UB) and allocator mismatches.

So proper handling is needed, especially as the platform is moving towards Rust-managed memory for safety and maintainability.

### Problem
- Allocator mismatch / UB: Some C-allocated memory was being freed by Rust using Box::from_raw, which is unsafe.
- Miri test failures: Isolated operations like open and drop(Box::from_raw) caused errors in Miri.
- Potential crashes / undefined behavior: If Rust tries to free C-allocated memory incorrectly, it can panic or corrupt memory.

### Changes Made
- Transferred ownership to Rust for certain allocations that Rust should manage.
- Replaced drop(Box::from_raw(ptr)) with a safe free(ptr) where the memory is still C-managed.
- Used Rust’s dealloc for Rust-allocated memory, ensuring ownership and deallocation match.
- Exposed free in libccxr_exports::demuxer to correctly free C-allocated memory where needed.
- Made CCX_DTVCC_MAX_COLUMNS public for safe Rust usage when calculating layouts in deallocations.
- Ensured sub.data in simplexml.rs is freed with the correct allocator (free) instead of Box::from_raw.

### Flowchart / Old vs New Behavior

- Old Behavior (Unsafe / UB Risk)

```rust
C allocates memory -> Pointer passed to Rust -> Rust does Box::from_raw -> Rust frees
```

Problems
Allocator mismatch
UB if C malloc and Rust dealloc mismatch
Miri and test errors

- New Behavior (Safe Ownership / Correct Free)
```rust 
C allocates memory -> Pointer passed to Rust ->Rust uses pointer but calls `free(ptr)` from C (no mismatch)
      
Rust allocates memory -> Pointer stays in Rust -> Rust deallocs with `dealloc`
```

### Benefits
- Memory safely freed according to its allocator
- Ownership matches Rust or C context
-  Eliminates UB and test failures

### Why Ownership Changed to Rust
- The platform is moving towards Rust-first memory management.
- Centralizing ownership in Rust reduces complexity and makes future maintenance safer.
- Rust’s compile-time guarantees prevent memory leaks and runtime crashes.

### Test Plan
- cargo test (all decoder, demuxer, and encoder tests passed)
- cargo miri test (Miri no longer reports UB in ownership paths)

### DIFF
--- a/src/rust/src/parser.rs
+++ b/src/rust/src/parser.rs
@@ -1663,6 +1663,7 @@ pub mod tests {
         common::{MkvLangFilter, OutputFormat, SelectCodec, StreamMode, StreamType},
         util::{encoding::Encoding, log::DebugMessageFlag},
     };
+    use serial_test::serial;
 
     /// # Safety
     ///
@@ -2318,18 +2319,21 @@ pub mod tests {
     }
 
     #[test]
+    #[serial]
     fn test_buffersize_with_k_suffix() {
         let (_, _) = parse_args(&["--buffersize", "64K"]);
         assert_eq!(get_file_buffer_size(), 64 * 1024);
     }
 
     #[test]
+    #[serial]
     fn test_buffersize_with_m_suffix() {
         let (_, _) = parse_args(&["--buffersize", "2M"]);
         assert_eq!(get_file_buffer_size(), 2 * 1024 * 1024);
     }
 
     #[test]
+    #[serial]
     fn test_buffersize_with_numeric_value() {
         let (_, _) = parse_args(&["--buffersize", "8192"]);
         assert_eq!(get_file_buffer_size(), 8192);


THE-Amrit-mahto-05:
@cfsmp3 
Thanks for the detailed review
I’ve refactored the implementation to fully address the points you raised
1) Centralized row allocation/deallocation via alloc_row/dealloc_row
2) Removed duplicated Layout calculations
3) Introduced a central ffi_alloc module for C allocator interactions
4) Consolidated free declarations and removed demuxer-specific exposure
5) Kept CCX_DTVCC_MAX_COLUMNS crate-private

All allocation/deallocation paths now match ownership correctly
Please let me know if required any new changes

PR #2079, opened 2026-02-01
Fix empty WebVTT files for HLS compatibility (Issue #1743)

Atul-Chahar:
This fixes issue #1743 where CCExtractor generates empty .webvtt files when processing content with no subtitles, breaking HLS compatibility.

Changes:
- Fixed X-TIMESTAMP-MAP field order (LOCAL before MPEGTS) per HLS specification 
- Added WebVTT case in write_subtitle_file_footer() to ensure header is always written
- Added function declaration for write_webvtt_header in header file

Fixes: #1743

---

{pull request content here}

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -176,6 +176,14 @@ int write_subtitle_file_footer(struct encoder_ctx *ctx, struct ccx_s_write *out)
 		case CCX_OF_CCD:
 			ret = write(out->fh, ctx->encoded_crlf, ctx->encoded_crlf_length);
 			break;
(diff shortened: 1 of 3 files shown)


cfsmp3:
Reviewed and verified:

1. **Field order change** (`MPEGTS,LOCAL` → `LOCAL,MPEGTS`): Confirmed correct per [RFC 8216](https://datatracker.ietf.org/doc/html/rfc8216) which specifies `X-TIMESTAMP-MAP=LOCAL:<cue time>,MPEGTS:<MPEG-2 time>`.

2. **Empty file header fix**: Ensures proper WebVTT format even when no captions are extracted, fixing HLS compatibility (Issue #1743).

Merging.

PR #2089, opened 2026-02-03
Fix/webvtt always include timestamp map

rajj28:
## Summary

This PR ensures that every WebVTT file generated by CCExtractor contains the `X-TIMESTAMP-MAP` header, which is required for proper HLS (HTTP Live Streaming) playback per [RFC 8216 §3.5](https://datatracker.ietf.org/doc/html/rfc8216#section-3.5).

## Problem

Previously, the `X-TIMESTAMP-MAP` header was only written when:
1. The `--timestamp-map` flag was explicitly passed, AND
2. Timing information was available in the stream, AND  
3. Subtitles were actually found

This caused HLS players to incorrectly synchronize subtitles because the header was often missing.

## Solution

- **Always** write `X-TIMESTAMP-MAP` in WebVTT output
- Use actual MPEGTS timestamp when timing info is available
- Fall back to `MPEGTS:0,LOCAL:00:00:00.000` when no timing exists
- Ensure header is written even when no subtitles are found (via footer fallback)
- Reset header flag for multi-segment and multi-track scenarios

## Changes

| File | Change |
|------|--------|
| `ccx_encoders_webvtt.c` | Core fix: always write X-TIMESTAMP-MAP |
| `ccx_encoders_common.c` | Footer fallback + flag resets for multi-segment/multi-track |
| `ccx_encoders_common.h` | Added `write_webvtt_header()` declaration |
| `params.c` | Marked `--timestamp-map` as deprecated |
| `args.rs`, `options.rs` | Rust documentation updates |

## Backward Compatibility

The `--timestamp-map` flag is preserved but now effectively a no-op. Existing scripts using this flag will continue to work.

## Testing

Tested with multiple scenarios:
- ✅ Real video with timing info → Uses actual MPEGTS timestamp (e.g., `MPEGTS:129003`)
- ✅ Empty/zeroed TS file → Uses fallback `MPEGTS:0`
- ✅ Invalid `--datapid` → Still produces header with fallback
- ✅ HLS segment → Correct timestamp mapping
- ✅ All output files have exactly ONE `X-TIMESTAMP-MAP` header

## Evidence

See attached screenshots/outputs showing:
After: WebVTT always contains X-TIMESTAMP-MAP
image

### DIFF
--- a/tests/webvtt_timestamp_map/README.md
+++ b/tests/webvtt_timestamp_map/README.md
@@ -1,0 +1,50 @@
+# WebVTT X-TIMESTAMP-MAP Test Suite
+
+This directory contains tests to verify that CCExtractor always includes the `X-TIMESTAMP-MAP` header in WebVTT output for HLS compliance.
+
+## What This Tests
+
+1. **Real video with timing info** → Uses actual MPEGTS timestamp
+2. **Empty/no subtitles** → Uses fallback `MPEGTS:0,LOCAL:00:00:00.000`
+3. **Invalid datapid** → Still produces header with fallback
+4. **All outputs have exactly ONE header** (no duplicates)
+
+## Running the Tests
+
+### Linux/WSL
+```bash
+cd tests/webvtt_timestamp_map
+chmod +x run_tests.sh
+./run_tests.sh /path/to/ccextractor
+```
+
+### Windows (PowerShell)
+```powershell
+cd tests\webvtt_timestamp_map
+.\run_tests.ps1 -CCExtractorPath "C:\path\to\ccextractor.exe"
+```
+
+## Expected Results
+
+All generated `.vtt` files should have:
+- Line 1: `WEBVTT`
+- Line 2: `X-TIMESTAMP-MAP=MPEGTS:<value>,LOCAL:<time>`
+- Exactly ONE `X-TIMESTAMP-MAP` header per file
+
+## Sample Output
+
+With timing info available:
+```
+WEBVTT
+X-TIMESTAMP-MAP=MPEGTS:129003,LOCAL:00:00:00.000
+
+1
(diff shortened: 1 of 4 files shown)


bbgdzxng1:
`X-TIMESTAMP-MAP` should remain disabled by default.

The primary task of ccextractor is to generate valid webVTT, first and foremost.

If `X-TIMESTAMP-MAP` is included by default, ccextractor would be producing invalid, out of spec webVTT. Support for webVTT headers was deprecated in 2016. https://mailarchive.ietf.org/arch/msg/hls-interest/4vmLpEsV-EnmkEwMQZkzbGQai_4/

Here is the official webVTT validator, which will fail validation if `X-TIMESTAMP-MAP` is present.
https://w3c.github.io/webvtt.js/parser.html

image


webVTT is the _canonical_ specification, HLS is the _derivative_ specification. Therefore, ccextractor's default options should follow the canonical standard - and any extra settings for HLS compatibility should be _optional_.

Otherwise, by default, ccextractor will be poisoning the web with invalid webvtt.

Reference to the PR where it was already changed to off by default: https://github.com/CCExtractor/ccextractor/commit/492f0d51976323cd8acbfef57a62a1285a295505

cfsmp3:
Closing for the same reason as #2093.

**Decision:** X-TIMESTAMP-MAP should remain **disabled by default** to maintain W3C WebVTT spec compliance.

**Rationale:**
1. X-TIMESTAMP-MAP is not part of the W3C WebVTT specification - it's an HLS-specific extension
2. This was intentionally disabled by default in commit 492f0d51 (Dec 2022)
3. Enabling it by default produces files that fail W3C WebVTT validation

**For HLS users:** The `--timestamp-map` flag is available to enable this feature when needed.

See discussion in #2093 for full context.

PR #2093, opened 2026-02-04
Feature/timestamp map control

pranavshar223:

---

As discussed in the issue, the `X-TIMESTAMP-MAP` header is required for strict HLS compliance but was previously disabled by default, creating invalid WebVTT files unless the user manually added a flag.

This PR makes CCExtractor HLS-compliant out of the box by flipping the default behavior to **Enabled**. To satisfy the requirement for flexibility (for non-HLS use cases), I have also added a `--no-timestamp-map` flag to explicitly disable the header.

### Technical Changes
This implementation bridges the C core and the Rust argument parser:
* **`src/lib_ccx/ccx_common_option.c`**: Changed the default value of `timestamp_map` from `0` to `1`.
* **`src/lib_ccx/params.c`**: Updated the help text to reflect that the header is now default.
* **`src/rust/src/args.rs`**: Added the new `--no-timestamp-map` argument definition.
* **`src/rust/src/lib.rs`**: Added logic to override the default C option when the Rust negative flag is present.

### How to Use / Testing Commands
You can verify the behavior with any TS or MP4 file:

**1. Default Behavior (Header Enabled)**
*Command:*
`./ccextractor input.ts -out=webvtt -o default.vtt`
*Result:* The `X-TIMESTAMP-MAP` header is automatically included.

**2. Disable Header (New Flag)**
*Command:*
`./ccextractor input.ts -out=webvtt --no-timestamp-map -o disabled.vtt`
*Result:* The `X-TIMESTAMP-MAP` header is omitted.

### Evidence of Fix
*(See attached comparison screenshot below)*
image

### DIFF
--- a/src/lib_ccx/ccx_common_option.c
+++ b/src/lib_ccx/ccx_common_option.c
@@ -41,7 +41,7 @@ void init_options(struct ccx_s_options *options)
 	options->live_stream = 0;     // 0 -> A regular file
 	options->messages_target = 1; // 1=stdout
 	options->print_file_reports = 0;
-	options->timestamp_map = 0; // Disable X-TIMESTAMP-MAP header by default
+	options->timestamp_map = 1; // Enable  X-TIMESTAMP-MAP header by default
 
 	/* Levenshtein's parameters, for string comparison */
 	options->dolevdist = 1;		  // By default attempt to correct typos
--- a/src/lib_ccx/params.c
+++ b/src/lib_ccx/params.c
@@ -268,7 +268,8 @@ void print_usage(void)
 	mprint("                       less or equal than the max allowed..\n");
 	mprint("	   --analyzevideo  Analyze the video stream even if it's not used for\n");
 	mprint("                       subtitles. This allows to provide video information.\n");
-	mprint("  --timestamp-map      Enable the X-TIMESTAMP-MAP header for WebVTT (HLS)\n");
+	mprint("  --timestamp-map       Enable X-TIMESTAMP-MAP (default)\n");
+	mprint("  --no-timestamp-map    Disable X-TIMESTAMP-MAP header\n");
 	mprint("Levenshtein distance:\n");
 	mprint("           --no-levdist: Don't attempt to correct typos with Levenshtein distance.\n");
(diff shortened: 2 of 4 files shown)


bbgdzxng1:
Strong disagree.  `X-TIMESTAMP-MAP` should be disabled by default.

The primary task of ccextractor is to generate valid webVTT _first and foremost_.

If `X-TIMESTAMP-MAP` is included by default, ccextractor would be producing invalid, out of spec webVTT.  Support for webVTT headers was deprecated in 2016.  https://mailarchive.ietf.org/arch/msg/hls-interest/4vmLpEsV-EnmkEwMQZkzbGQai_4/

Here is the official webVTT validator.  Any webVTTs containing `X-TIMESTAMP-MAP` will fail validation.
https://w3c.github.io/webvtt.js/parser.html

webVTT is the _canonical_ specification, HLS is the _derivative_ specification.  Therefore, default options should follow the canonical standard - and any extra settings for HLS uses should be optional.

Otherwise, ccextractor will be poisoning the web with invalid webvtt by default.

Reference to where it was already changed to off by default: https://github.com/CCExtractor/ccextractor/commit/492f0d51976323cd8acbfef57a62a1285a295505

cfsmp3:
Thank you for the PR, but after review we're keeping X-TIMESTAMP-MAP disabled by default.

**Reasoning:**

1. **W3C WebVTT compliance**: X-TIMESTAMP-MAP is not part of the W3C WebVTT specification - it's an HLS-specific extension. Enabling it by default produces files that fail W3C validation.

2. **Previous maintainer decision**: This was intentionally disabled by default in commit 492f0d51 (Dec 2022) with the rationale that "X-TIMESTAMP-MAP is only part of the HLS spec, and is not valid WebVTT."

3. **User control already exists**: Users who need HLS compliance can use `--timestamp-map` to enable it.

@bbgdzxng1's points are well-taken - they have prior experience with this specific feature (Issue #1463) and the standards considerations are valid.

For HLS users: the `--timestamp-map` flag remains available. We also just merged #2079 which fixes the field order to `LOCAL,MPEGTS` per RFC 8216.

PR #2106, opened 2026-02-10
Fix fragmented teletext subtitles by delaying flush until sentence completion

Apprentice2907:
**[FIX]** Fix fragmented teletext subtitles by delaying flush until sentence completion

---

## Problem

In some teletext broadcasts, subtitles are transmitted as incremental fragments where only a few characters change between consecutive pages. CCExtractor currently flushes the previous subtitle whenever the fuzzy comparison detects a difference between pages, causing each fragment to be written as a separate subtitle entry.

**Example of incorrect output:**
```
Ra
Pe
wee
aan !
```

**Expected output:**
```
Rape wee aan!
```

## Root Cause

The issue occurs in `process_page()` inside `telxcc.c`. When `fuzzy_memcmp()` detects a difference between the previous and current teletext page, the code immediately flushes the previous subtitle via `telxcc_dump_prev_page(ctx, sub)`. However, in some broadcasts, teletext pages are updated gradually, causing small differences that trigger premature flushing.

## Solution

This patch modifies the flushing logic so that the previous subtitle is only flushed if it ends with a sentence-terminating punctuation mark (`.`, `?`, `!`, `:`). Otherwise, the current fragment is merged into the previous buffer and the subtitle is not flushed yet.

This allows incremental teletext fragments to be combined into a complete subtitle before being written.

## Changes

- **File modified:** `src/lib_ccx/telxcc.c`
- **Function:** `process_page()`
- Only the flushing condition after `fuzzy_memcmp()` was adjusted

## Testing

Tested with:
1. Teletext streams producing fragmented subtitles → Now merged into complete sentences
2. Normal teletext subtitle streams → Behavior unchanged

## Benefits

- Prevents fragmented subtitle output
- Preserves existing timing and buffering logic
- Does not modify teletext decoding or encoder behavior
- Minimal and safe change limited to flush logic

### DIFF
--- a/src/lib_ccx/telxcc.c
+++ b/src/lib_ccx/telxcc.c
@@ -884,11 +884,48 @@ page_is_empty:
 			}
 			else
 			{
-				// OK, the old and new buffer don't match. So write the old
-				telxcc_dump_prev_page(ctx, sub);
-				ctx->prev_hide_timestamp = page->hide_timestamp;
-				ctx->prev_show_timestamp = page->show_timestamp;
+				// Instead of dumping immediately, check if sentence ended
+				if (ctx->page_buffer_prev && ctx->page_buffer_cur)
+				{
+					int len = strlen(ctx->page_buffer_prev);
+
+					// Skip trailing spaces/newlines
+					while (len > 0 &&
+						(ctx->page_buffer_prev[len - 1] == ' ' ||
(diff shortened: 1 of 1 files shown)


cfsmp3:
Closing — duplicate of #2108, no activity since February.

PR #2107, opened 2026-02-10
[FIX]Fix panic when using --mp4/--mkv without explicit input format

gaurav02081:
[FIX]Fix panic when using --mp4/--mkv without explicit input format

---
Found a issue --

if  the user use --mp4 ya --mkv with invalid / broken file ,then the programs gets crashes instead of handling that with error message .

it is happeing because of the   unwrap()  METHOD         
**command used for reproduce the issue** 

```
dd if=/dev/urandom of=bad.mp4 bs=1024 c
ount=1 && ./mac/ccextractor bad.mp4 --mp4
```

image

FIX -- (src/rust/src/parser.rs)

and added flags  rgs.mp4 and args.mkv in the set_input_format  function
replace unwrap with map with closures.

After the fix, the program no longer crashes

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,22 @@ The core functionality is written in C. Other languages used include C++ and Pyt
 
 Downloads for precompiled binaries and source code can be found [on our website](https://ccextractor.org/public/general/downloads/).
 
+### WebVTT Output Options
+
+CCExtractor supports optional WebVTT-specific headers for advanced use cases
+such as HTTP Live Streaming (HLS).
+
+#### `--timestamp-map`
+
+Enable writing the `X-TIMESTAMP-MAP` header in WebVTT output.
+
+This header is required for HLS workflows but is **disabled by default**
+to preserve compatibility with standard WebVTT players.
+
+Example:
+```bash
+ccextractor input.ts --timestamp-map -o output.vtt
+```
 
 ### Windows Package Managers
 
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,7 +1,11 @@
 0.96.6 (unreleased)
 -------------------
+- New: Add optional machine-readable JSON output for -out=report via --report-format json
+- Fix: Incorrect strlen argument when writing end timestamps in MKV subtitle extraction (WebVTT, SRT, ASS/SSA)
+- Fix: File descriptor leak and missing open() error check in MKV subtitle track saving
 - Fix: DVB EIT start time BCD decoding in XMLTV output causing invalid timestamps (#1835)
 - New: Add Snap packaging support with Snapcraft configuration and GitHub Actions CI workflow. 
+- New: Implement dictionary-based capitalization and censorship for transcripts 
 - Fix: Clear status line output on Linux/WSL to prevent text artifacts (#2017)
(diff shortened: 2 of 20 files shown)


gaurav02081:
Previously:

set_input_format() was called conditionally through a guard in parse_parameters(), meaning it was skipped during most normal (autodetection) runs.

Additionally, the fatal! fallback inside set_input_format() would terminate execution if no explicit format matched, preventing autodetection from taking place.

This caused unintended behavior:

Normal runs could bypass initialization logic.

If triggered without a matching format, the function would exit instead of allowing autodetection.

Final Fix

Removed the fatal! fallback so that when no explicit format flag matches, the function simply returns and preserves autodetection behavior.

Removed the guard in parse_parameters() so that set_input_format() is always called as part of initialization.

Explicit handling for --mp4 and --mkv remains intact.

The unsafe unwrap() has been eliminated.

cfsmp3 (changes requested):
## Review

The bug is real — confirmed that `--mp4` (and `--mkv`) cause an abort on master because `set_input_format()` doesn't handle these flags, falls through to the `fatal!` macro, and `args.input.unwrap()` panics on `None`.

However, the fix is over-engineered. You're commenting out the `fatal!` and commenting out the guard condition, when the actual fix is just 2 lines.

### What's wrong with the current approach:

1. **Commenting out code instead of removing it** — commented-out code is dead code. Either remove it or keep it.
2. **Removing the guard condition** — the guard at line 661-672 correctly prevents calling `set_input_format()` when no format flags are set. Removing it is unnecessary.
3. **Removing the fatal** — the `fatal!` for unknown formats is a useful error message. It should stay, just with the `unwrap()` fixed.

### The correct fix (3 lines changed):

In `set_input_format()`, add the missing handlers before the `else` branch:

```rust
        } else if args.wtv {
            self.set_input_format_type(InFormat::Wtv);
+       } else if args.mp4 {
+           self.set_input_format_type(InFormat::Mp4);
+       } else if args.mkv {
+           self.set_input_format_type(InFormat::Mkv);
        } else {
            fatal!(
                cause = ExitCause::MalformedParameter;
-              "Unknown input file format: {}\n", args.input.unwrap()
+              "Unknown input file format: {}\n", args.input.unwrap_or("unknown")
            );
        }
```

That's it. Keep the guard condition as-is (it already includes `args.mp4` and `args.mkv`). Keep the `fatal!` for genuinely unknown formats. Just add the missing handlers and fix the `unwrap()`.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

gaurav02081:
@cfsmp3 Thanks for the review I have applied the suggested changes and kept the original guard and fatal! logic intact.

While implementing -

`args.input.unwrap_or("unknown")
`
the build failed because args.input is of type Option<InFormat>, whereas "unknown" is a &str, which resulted in a type mismatch.

To keep the behavior the same while ensuring type safety, I updated the fatal! call to:

```
fatal!(
    cause = ExitCause::MalformedParameter;
    "Unknown input file format: {}\n",
    args.input
        .map(|i| i.to_string())
        .unwrap_or_else(|| "unknown".to_string())
);
```
This preserves the intended error message while avoiding the panic .

PR #2088, opened 2026-02-03
[RUST] Added XDS module 

vatsalkeshav:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [ ] I am an active contributor to CCExtractor.

---
**_(This PR Supersedes https://github.com/CCExtractor/ccextractor/pull/1890; rebased on latest `master` with fixes from review.)_**

This pull request migrates `ccx_decoders_xds.c` and `ccx_decoders_xds.h` to `/src/rust/xds` rust module.
 - `mod.rs` : exposes the Rust implementation to the C codebase
 - `types.rs` : has types and structures for decoding extended data
 - `handlers.rs` : has handler functions for processing extended data packets

Tested against stream : https://sampleplatform.ccextractor.org/sample/b22260d065ab537899baaf34e78a5184671f4bcb2df0414d05e6345adfd7812f

PR #2109, opened 2026-02-13
[RUST] Added ISDB module

vatsalkeshav:

---

This pull request migrates `src/lib_ccx/ccx_decoders_isdb.c, ccx_decoders_isdb.h` to `/src/rust/src/isdb` rust module.
 - `types.rs` - ISDB-specific types, enums, and constants etc.
 - `leaf.rs`.   - Low-level helper functions; independent of functions in `mid/high/mod.rs`
 - `mid.rs`    - Mid-level processing functions
 - `high.rs`   - High-level parser functions
 - `mod.rs`   - ffi from rust side

Tested against streams : 
 1. https://drive.google.com/file/d/0B_61ywKPmI0TMzlwZGw0MzdIc0U/view?usp=drive_link&resourcekey=0-UkWAUJcBIl_6uJL3nzCTcA
  2. https://drive.google.com/file/d/0B_61ywKPmI0TM3dKRlJ6UjI0STQ/view?usp=drive_link&resourcekey=0-zwyV8fkI_xMeM3_1zUZ_RQ


vatsalkeshav:
Please note :
 - this pr introduces slight output parity when tested against both the streams above
    Check [diff](https://github.com/vatsalkeshav/ccx-regression-testing/blob/master/tests/isdb/0/diff/diff_0_srt.txt) from stream 1 as of above (above is from ccextractor compiled off master, below is from ccextractor compiled off this pr):
    ```srt
    5691c5691
    < !
    ---
    > ! que o dia comecebem... !
    ```
    similar [diff](https://github.com/vatsalkeshav/ccx-regression-testing/blob/master/tests/isdb/1/diff/diff_1.txt) from stream 2 from above is also there :
    ```srt
    183c183
    <  e
    ---
    >  em c�psulas menores, mais f�ceis de ingerir.
    189c189
    <  n
    ---
    >  nas costas e musculares.
    207c207
    <  a
    ---
    >  a partir de 10 minutos
    ```
    
    In `/src/lib_ccx/ccx_decoders_isdb.c` ,~ lines `530-535` in `get_text` (func)
    
    ```c
    if (ccx_strstr_ignorespace(text->buf, sb_text->buf))
				{
					found = CCX_TRUE;
					// See if complete string is there if not update that string
					if (!ccx_strstr_ignorespace(sb_text->buf, text->buf))
					{
						reserve_buf(sb_text, text->used);
						memcpy(sb_text->buf, text->buf, text->used);
					}
					break;
				}
    ```
    
    Here, text is copied into the buffer but `sb_text->used` is not updated. Hence, when flushing, only the used bytes are written, so maybe, just `!` is output instead of the full cc.
    
    In `src/rust/src/isdb/mid.rs`, ~ lines `135-143` in `get_text` (func)
    
    ```rs
    if strstr_ignorespace(&text.buf, &sb.buf) {
                    found = true;
                    // See if complete string is there if not update that string
                    if !strstr_ignorespace(&sb.buf, &text.buf) {
                        sb.buf = text.buf.clone();
                    }
                    break;
                }
    ```
    Here, we replace the entire buffer via `clone()` - the reason why the complete text is output

 - The rust code owns the isdb context the entire time, so there's minimal need of c<->rust data transfer library

cfsmp3 (changes requested):
## Review: ISDB Rust Port

Nice work on this port. The code is well-organized across the 4 modules (types/leaf/mid/high), the C-to-Rust mapping tables in each module are very helpful, and the test suite is comprehensive.

I tested against the first sample stream you linked (the ~60min Brazilian broadcast) and the Rust output is 99.95% identical to the C version across 6300+ lines of subtitles — the few minor differences are actually slightly better output from the Rust side (less fragmented text in the rollup deduplication path).

### Bug: Operator precedence in `parse_caption_management_data` (leaf.rs:205)

```rust
if ctx.dmf == 0x0C || ctx.dmf == 0x0D || ctx.dmf == 0x0E && pos < buf.len() {
    ctx.dc = buf[pos];
```

In Rust, `&&` binds tighter than `||`, so this parses as:

```rust
ctx.dmf == 0x0C || ctx.dmf == 0x0D || (ctx.dmf == 0x0E && pos < buf.len())
```

The `pos < buf.len()` bounds check only protects the `0x0E` case. If `dmf` is `0x0C` or `0x0D` and the buffer happens to end right after the dmf byte, `buf[pos]` will **panic** (index out of bounds). Fix:

```rust
if (ctx.dmf == 0x0C || ctx.dmf == 0x0D || ctx.dmf == 0x0E) && pos < buf.len() {
```

### Minor: Wrong bytes in debug log (leaf.rs:210)

```rust
debug!("CC MGMT DATA: languages: {:?}", &buf[pos - 3..pos]);
```

At this point `pos` has been incremented by 1 past the dmf byte, so `buf[pos - 3..pos]` reads 3 bytes *before* the current position (unrelated data). The C code reads `buf[0], buf[1], buf[2]` at the current pointer position. The Rust equivalent should be `&buf[pos..pos + 3]`. This is just a log issue, not a functional bug.

### Note on Rust port merging timeline

Independently of this diff — we're currently working on improving our CI to include proper regression testing for Rust ports (comparing C vs Rust output against reference streams). Until that infrastructure is in place, we won't be merging Rust ports, because we lack confidence that we can catch regressions going forward. This is not a reflection of the quality of your work (which is solid), just a process/tooling gap on our side. We'll revisit once CI is ready.

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

vatsalkeshav:
Thanks for the review!
Please continue whenever convenient - this work was in process for a long time and it's good to see it now in final steps.
Could you please also review #2088

cfsmp3:
Seem abandoned and don't even pass format

PR #2128, opened 2026-02-20
Fix potential out-of-bounds access in write_stringz_as_srt_to_output

rhythmcache:
This PR adds a bound check before accessing string[pos_r + 1] to prevent a potential out-of-bounds read when pos_r == len - 1.
and corrects the iteration boundary when processing the unescaped buffer by using the actual written length (pos_w) instead of the original string length (len)


cfsmp3 (changes requested):
The fixes look correct — good catches on the OOB read and loop bound. However the CI format check is failing: spaces instead of tabs, trailing whitespace. Please fix the formatting and push again.

rhythmcache:
I've fixed the formatting

PR #2131, opened 2026-02-20
Fix PTS rollover detection using per-stream state instead of shared s…

Aman-Cool:
[FIX] Fix PTS rollover detection using per-stream state instead of shared static

- [x] I have read and understood the [[contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md)](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] **I have mentioned this change in the [[changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT)](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

---

`read_video_pes_header()` used a `static` local variable (`current_pts_33`) to track the previous PTS for rollover detection. Since it's static, it was shared across all PIDs and across all input files in a batch run.

When file 1 ends with a high PTS (top bits = 7) and file 2 starts fresh with a low PTS, the rollover detector incorrectly treats this as a legitimate wrap and increments `rollover_bits`. Every subtitle in file 2 then gets its timestamp shifted forward by ~26.5 hours. CCExtractor exits 0, the output file looks valid, but all timestamps are completely wrong.

Fixed by moving `current_pts_33` into `demuxer_data` as `prev_pts_33` so each stream independently tracks its own previous PTS. 3 files, 5 lines changed. No logic changes.

### DIFF
--- a/src/lib_ccx/ccx_demuxer.c
+++ b/src/lib_ccx/ccx_demuxer.c
@@ -452,6 +452,7 @@ struct demuxer_data *alloc_demuxer_data(void)
 	data->len = 0;
 	data->pts = CCX_NOPTS;
 	data->rollover_bits = 0;
+	data->prev_pts_33 = 0;
 	data->tb.num = 1;
 	data->tb.den = 90000;
 	data->next_stream = 0;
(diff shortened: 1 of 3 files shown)


Aman-Cool:
Hey @cfsmp3 ,This fixes a silent bug where batch-processing multiple TS files could shift all subtitle timestamps in file 2 by ~26.5 hours due to a `static` variable being shared across streams. The fix is small; 3 files, 5 lines, just moving that state into the per-stream struct where it belongs.

Verified the bug in the actual source, confirmed it compiles clean, and updated the changelog before opening this. Happy to make changes if anything needs adjusting!

PR #2133, opened 2026-02-22
fix: use absolute paths in bframe_pts regression test script

DhanushVarma-2:

---

{pull request content here}
The test script used a relative path ../../mac/ccextractor which only worked when run from a specific directory. Fixed by using $SCRIPT_DIR to construct absolute paths so the script works correctly from any location.

### DIFF
--- a/tests/regression/bframe_pts/run_test.sh
+++ b/tests/regression/bframe_pts/run_test.sh
@@ -1,0 +1,19 @@
+#!/bin/sh
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+CCX="$SCRIPT_DIR/../../../mac/ccextractor"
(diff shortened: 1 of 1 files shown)


cfsmp3:
Closing — this test script isn't usable in its current form:

- **Hardcodes macOS path** (`mac/ccextractor`) — won't work on Linux or Windows
- **Missing test files** — references `bframe_pts_regression_small.ts` and `bframe_pts_expected.srt` that aren't included in the PR
- **No integration with existing test infrastructure** — the project uses a test suite system that this doesn't plug into
- **Shell scripting issues** — variables like `$SAMPLE` and `$OUTPUT` aren't quoted, which would break on paths with spaces

If you'd like to contribute a regression test, take a look at how the existing test suite works under `tests/` and integrate with that framework. Test sample files should go on the sample platform, not in the repo.

PR #1850, opened 2025-12-19
[FIX] -out=spupng with EIA608/teletext: offset values in XML may be not correct #893

gaurav02081:
**[FIX]** 

**In raising this pull request, I confirm the following (please check boxes):**

- [ ✔] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [ ✔] I have checked that another pull request for this purpose does not exist.
- [ ✔] I have considered, and confirmed that this submission will be valuable to others.
- [✔ ] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [✔ ] I give this submission freely, and claim no ownership to its content.
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [x] I absolutely love CCExtractor, but have not contributed previously.
- [ ✔] I am an active contributor to CCExtractor.

---

[FIX] -out=spupng with EIA608/teletext: offset values in XML may be not correct #893

The issue occurs because the offsets were hardcoded for a 544×390 subtitle image centered on a 720×480 NTSC screen, while the text-to-PNG renderer generates PNGs with different (dynamic) dimensions (e.g. 600px wide and variable height). This results in visibly misaligned subtitles, especially for Teletext (PAL) sources.

This PR removes the hardcoded NTSC assumptions and calculates SPUPNG offsets dynamically based on the actual rendered PNG geometry and the target screen resolution.

Key changes =>

1. Store the real PNG dimensions (img_w, img_h) in spupng_t

2 .Introduce an is_pal flag in encoder_ctx to distinguish PAL (576 lines) vs NTSC (480 lines)

3. Capture final canvas dimensions during text-to-PNG rendering

4.Add a helper to compute centered xOffset / yOffset dynamically

5. Enforce even yOffset to prevent interlacing flicker on DVD/SPU

6. Apply the new logic only to the text-to-SPUPNG pipeline, leaving the bitmap path unchanged

7. This aligns the text-based SPUPNG output with the already-correct bitmap SPUPNG behavior.

**So ,Subtitles are correctly positioned regardless of PNG width/height (PNG size decides the offset — not hardcoded values.).**

in the given image i have drawn the flow of this fix .
IMG_1181

PR #2134, opened 2026-02-22
[Fix]: SPUPNG indexing, EOD counter advance, and header path determinism

x15sr71:

---

### SPUPNG: Fix file indexing, EOD counter advance, and header path determinism

Three correctness fixes in the SPUPNG encoder. These are independent of each other but all live in the same file.

#### 1. File numbering corrected: first file is now sub0000 instead of sub0001

`fileIndex` was initialized to `0`, but `inc_spupng_fileindex()` increments before writing, so the first file was always `sub0001.png`. The [CCExtractor SPUPNG docs](https://ccextractor.org/public/general/using_spupng/) specify numbering starts at `sub0000.png`.

Fix: initialize `fileIndex` to `-1` so the first increment lands at `0`.

#### 2. Index advances on EOD frames with no output

`inc_spupng_fileindex()` was called before the `nb_data == 0 && SUB_EOD_MARKER` early-return in `write_cc_bitmap_as_spupng()`. The counter advanced even though no PNG was written, creating gaps. Moved the increment past the early-return.

#### 3. XML header comment embeds full runtime path

`spupng_write_header()` wrote the full input path into the XML comment. Windows CI produces `.\TestFiles\...` while the golden uses `/repository/TestFiles/...` — they can never match. Changed to basename only.

---

**Note on golden:** The golden for [sample121/c83f765c...](https://sampleplatform.ccextractor.org/sample/c83f765c661595e1bfa4750756a54c006c6f2c697a436bc0726986f71f0706cd) predates PR #1850 (offset correction) and subsequent timing fixes. As a result, its offsets and timestamps do not match current master output and will require regeneration.

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 0.96.6 (2026-02-19)
 -------------------
+- FIX: spupng start numbering at sub0000, avoid index advance on empty EOD, normalize header filename
 - New: 32-bit (x86) Windows build and installer (#2116)
 - New: Add optional machine-readable JSON output for -out=report via --report-format json
 - New: Add Snap packaging support with Snapcraft configuration and GitHub Actions CI workflow


x15sr71:
@cfsmp3 Following up on the office hours discussion, I manually validated [sample121/c83f765c...](https://sampleplatform.ccextractor.org/sample/c83f765c661595e1bfa4750756a54c006c6f2c697a436bc0726986f71f0706cd) timing as suggested.

What I did:
- Extracted SPUPNG and SRT from current master.
- Played the sample in VLC with SRT overlay and compared subtitle arrival against dialogue.
- Inspected generated PNGs and XML directly.

Findings:
- All subtitles align correctly with dialogue during playback, the golden's 1.034 baseline is noticeably early
- Offsets (60,180) match post-#1850 behavior, the golden shows the older values (88,46).
- File numbering in the result is sequential (sub0000→sub0003), matching documented behavior.
- Header path normalization is working, CI now shows basename only, removing the previous platform-dependent mismatch.

The file numbering and path normalization fixes are clearly reflected in the Linux CI output. The Windows run is still pending.. Fix 3 (basename normalization) was introduced specifically to eliminate the previous `.\\TestFiles\\...` vs `/repository/TestFiles/...` path mismatch that caused nondeterministic SPUPNG failures across PRs.

Based on the current CI output and manual validation, the remaining difference for `--out=spupng` reflects a stale golden, generated against pre-#1850 behavior. Current master timing aligns correctly with playback, the golden needs regeneration to match.

Happy to regenerate and submit the updated golden as a follow-up PR if that’s the preferred workflow.

PR #2137, opened 2026-02-22
Add teletext pages and PID type tagging to JSON report output (#1399)

### LINKED ISSUE #1399: [PROPOSAL] - Structured data JSON output of ccextractor -out=report
Would you be kind enough to consider extending `ccextractor -out=report` to export a JSON output so that the output of the ccextractor analyser can be machine readable without a custom parser?  This would mean that an external structured data parser such as `jq` can be used to parse / filter the output.

Similar to:
```
$ ffprobe -show_streams -print_format json -i INFILE | jq .
$ mediainfo -full --Output=JSON INFILE | jq .
```

eg something like...
```
$ ccextractor --output-format json_templatev1 -out=report INFILE | jq .
```

Obviously, it would require the definition of a schema, and the data structure of all the potential outputs of ccextractor may need to be extended in the future (hence proposal of versioning the schema), but it would seem like useful functionality for anyone who uses ccextractor in an automated environment.  I suspect there is already a defined data structure internal to ccextractor anyway.

Thanks!

ananyaaa66:
Closes #1399

Summary:
- Added "type" field to PID entries
- Added "teletext_pages" array
- Added helper in telxcc.c

This builds on the existing --report-format json infrastructure.

### DIFF
--- a/src/lib_ccx/params_dump.c
+++ b/src/lib_ccx/params_dump.c
@@ -429,8 +429,16 @@ void print_file_report_json(struct lib_ccx_ctx *ctx)
 		int is_dvb = 0, is_tlt = 0;
 		for (int k = 0; k < SUB_STREAMS_CNT; k++)
 		{
-			if (demux->freport.dvb_sub_pid[k] == (unsigned)pid) { is_dvb = 1; break; }
-			if (demux->freport.tlt_sub_pid[k] == (unsigned)pid) { is_tlt = 1; break; }
+			if (demux->freport.dvb_sub_pid[k] == (unsigned)pid)
+			{
+				is_dvb = 1;
+				break;
+			}
+			if (demux->freport.tlt_sub_pid[k] == (unsigned)pid)
+			{
+				is_tlt = 1;
+				break;
+			}
 		}
 		if (is_dvb)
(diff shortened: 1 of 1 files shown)


ananyaaa66:
Re: Regression test failures

I've investigated the 8 failing regression tests reported by the bot and confirmed they are not caused by this PR's changes. Here's the analysis:

All code changes are strictly inside the JSON report code path:

The additions in 

params_dump.c
 are inside 

print_file_report_json()
, which is only invoked when --report-format json is passed

tlt_print_seen_pages_json()
 in 

telxcc.c
 is a new standalone function only called from the JSON path

teletext.h
 only adds the declaration for the above function
The failing tests don't use JSON output:

--autoprogram --out=srt --latin1 --quant 0
--autoprogram --out=ttxt --latin1 --ucla
--out=srt --latin1 --autoprogram
--startcreditstext variants
None of these commands pass --report-format json, so the new code is never executed during these tests.

All 22 GitHub Actions CI checks pass (Linux, Windows, Mac builds + formatting).

These appear to be pre-existing flaky tests on the sample platform. Could a maintainer confirm or trigger a re-run?

.

PR #2139, opened 2026-02-24
Hex to int clean

Anayo-Anyafulu:
## [IMPROVEMENT] Port hex_to_int from C to Rust

In raising this pull request, I confirm the following:

- [x] I have read and understood the contributors guide.
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [ ] I have mentioned this change in the changelog. *(not needed — cfsmp3 said this is too minor for changelog)*

### My familiarity with the project

- [x] I have used CCExtractor just a couple of times.

### Description

Clean implementation of the `hex_to_int` port as suggested in #1774.  
Fresh branch from `master` with only the necessary 5 files changed — no unrelated `Cargo.lock` or whitespace changes.

### Changes

- `src/rust/lib_ccxr/src/util/hex.rs` — Rust implementation  
- `src/rust/lib_ccxr/src/util/mod.rs` — added `pub mod hex;`  
- `src/rust/src/libccxr_exports/util.rs` — FFI export  
- `src/rust/src/libccxr_exports/mod.rs` — added `pub mod util;`  
- `src/lib_ccx/utility.c` — added `#ifndef DISABLE_RUST` guard  

### Testing

- All tests passing locally.

PR #2141, opened 2026-02-25
Port hex_string_to_int from C to Rust

Anayo-Anyafulu:
[IMPROVEMENT] Port `hex_string_to_int` from C to Rust

- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).** (not needed — too minor)

---

Clean port of `hex_string_to_int` from C to Rust, following the same pattern as #2139.

## Changes
- `src/rust/lib_ccxr/src/util/hex.rs` — Rust implementation reusing `hex_char_to_val`
- `src/rust/src/libccxr_exports/util.rs` — FFI export with added null check
- `src/lib_ccx/utility.c` — `#ifndef DISABLE_RUST` guard added

Note: The Rust version now accepts uppercase hex (A-F), which is an intentional improvement over the original C version that only accepted lowercase (a-f).

## Testing
All 48 Rust unit tests passing locally.

### DIFF
--- a/src/lib_ccx/utility.c
+++ b/src/lib_ccx/utility.c
@@ -269,6 +269,13 @@ void sleep_secs(int secs)
 #endif
 }
 
+#ifndef DISABLE_RUST
+extern int ccxr_hex_to_int(char high, char low);
+int hex_to_int(char high, char low)
+{
+	return ccxr_hex_to_int(high, low);
+}
+#else
 int hex_to_int(char high, char low)
 {
 	unsigned char h, l;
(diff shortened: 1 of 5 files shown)


Anayo-Anyafulu:
The DVB, General, and Options test failures are pre-existing and unrelated to this change. These same categories fail on master. All build checks and format checks are passing

cfsmp3 (changes requested):
Thanks for the port — the implementation is correct and tests pass. A few things to fix before merging:

1. **Null check needed**: `ccxr_hex_string_to_int` calls `CStr::from_ptr()` without checking if the pointer is null first. Please add a null guard.

2. **Squash commits**: 7 commits is too many for this change. Please squash into 1-2 meaningful commits.

3. **Unrelated change**: There's a blank line removal in `utility.c` that's not related to this PR. Please revert it.

4. **Uppercase hex**: The new Rust version accepts uppercase (A-F) while the C version only accepts lowercase (a-f). This is fine as an improvement, but please mention it in the PR description so it's documented as an intentional behavior change.

Once these are addressed this is ready to merge. Also note: #2139 will be closed in favor of this PR.

Anayo-Anyafulu:
@cfsmp3 

Thanks for the review! I'll address all four points — adding the null guard, squashing the commits, reverting the unrelated blank line in utility.c, and updating the PR description to document the uppercase hex behavior change. Will push the fixes shortly.

PR #2143, opened 2026-02-25
Fix Windows build: MSVC CRT and GPAC linking

cheron2000:

---

{pull request content here}

[FIX] Fix Windows build: CRT linkage and GPAC integration

### TL;DR
Fixes Windows build failures by enforcing dynamic MSVC CRT (/MD) across all targets (C, C++, Rust) and correcting GPAC linking. No impact on Linux/macOS.

### Changes included:

**1. CRT Configuration**
- Added a global MSVC runtime policy (/MD) in `src/CMakeLists.txt` to enforce dynamic CRT across all targets.
- Explicitly set the `ccx` library target to use dynamic CRT in `src/lib_ccx/CMakeLists.txt`.
- Overrode Rust library's default static CRT in `src/rust/CMakeLists.txt`.

**2. GPAC Integration**
- Added GPAC library directories for linking in `src/CMakeLists.txt` and `src/lib_ccx/CMakeLists.txt`.
- Updated include and library paths to ensure GPAC is properly found on Windows.

**3. Platform-specific Fixes**
- Wrapped Unix-style linker flags (-lm -lpthread -ldl) in `if(NOT WIN32)` to prevent Windows warnings.

**4. Cleanup**
- Removed redundant/incorrect CRT settings from `src/CMakeLists.txt` and `src/lib_ccx/CMakeLists.txt`.

### Verification
- ✅ Windows build completed successfully (Visual Studio 2022, CMake 3.31+)
- ✅ Executable runs and reports version correctly
- ✅ No CRT linker errors observed
- ✅ Linux/macOS builds remain unaffected

### Notes
- GPAC path is currently hardcoded to `C:/Program Files/GPAC/sdk/lib`; users with different installations may need to update it.
- Rust CRT override uses `/NODEFAULTLIB:libcmt` to force dynamic CRT.

This PR improves cross-platform build reliability and cleans up CMake configurations without affecting functionality.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -42,6 +42,12 @@ configure_file (
 
 add_definitions(-DVERSION_FILE_PRESENT -DFT2_BUILD_LIBRARY -DGPAC_DISABLE_VTT -DGPAC_DISABLE_OD_DUMP -DGPAC_DISABLE_REMOTERY -DNO_GZIP)
 
+# Set global CRT policy to dynamic (/MD) before any targets are defined
+# This ensures all targets use the same CRT to avoid linker errors
+if(MSVC)
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreadedDLL")
+endif()
+
 if(CMAKE_SIZEOF_VOID_P EQUAL 8)
   add_definitions(-DGPAC_64_BITS)
 endif()
@@ -156,8 +162,10 @@ if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
     set (EXTRA_LIBS ${EXTRA_LIBS} iconv)
 endif (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
 
-
-set (EXTRA_LIBS ${EXTRA_LIBS} -lm -lpthread -ldl)
+# Add platform-specific system libraries
+if(NOT WIN32)
+    set (EXTRA_LIBS ${EXTRA_LIBS} -lm -lpthread -ldl)
+endif()
 
 find_package (PkgConfig)
 
@@ -240,6 +248,11 @@ if (PKG_CONFIG_FOUND AND WITH_HARDSUBX)
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${LEPTONICA_INCLUDE_DIRS})
 endif (PKG_CONFIG_FOUND AND WITH_HARDSUBX)
 
+# Add GPAC library directory to linker search path on Windows
+if(WIN32)
+    link_directories("C:/Program Files/GPAC/sdk/lib")
+endif()
+
 add_executable (ccextractor ${SOURCEFILE} ${FREETYPE_SOURCE} ${UTF8PROC_SOURCE})
 
 ########################################################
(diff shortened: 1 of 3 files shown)


cfsmp3 (changes requested):
Thanks for working on this, but there are several issues that need to be addressed:

1. **Not validated by CI**: Our Windows CI uses msbuild (via the .sln), not CMake. The CMake path you're modifying isn't exercised by CI, so we can't verify these changes automatically.

2. **Hardcoded GPAC path**: `C:/Program Files/GPAC/sdk/lib` is hardcoded — this will break on any system where GPAC is installed elsewhere (or not installed at all). Our CI builds use vcpkg for dependencies.

3. **CRT strategy conflict**: Our CI sets `RUSTFLAGS: -Ctarget-feature=+crt-static` for static CRT linking. This PR switches to dynamic CRT (`/MD`) and adds `/NODEFAULTLIB:libcmt` to suppress conflicts — that directly contradicts our build strategy and will cause issues when the Rust static libs (built with static CRT) are linked against C code built with dynamic CRT.

4. **Fragile `/NODEFAULTLIB:libcmt`**: This is a workaround that masks the real problem rather than fixing it. Mixed CRT linkage leads to subtle runtime bugs (heap corruption when memory is allocated in one CRT and freed in another).

5. **The `-lm -lpthread -ldl` Unix guard**: This part is the only unambiguously correct change — these libs don't exist on Windows and should be guarded. But it should be submitted separately.

Please consider splitting out the Unix library guard into its own PR. The CRT and GPAC linking changes need a fundamentally different approach that aligns with how our CI actually builds.

cheron2000:
> Thanks for working on this, but there are several issues that need to be addressed:
> 
> 1. **Not validated by CI**: Our Windows CI uses msbuild (via the .sln), not CMake. The CMake path you're modifying isn't exercised by CI, so we can't verify these changes automatically.
> 
> 2. **Hardcoded GPAC path**: `C:/Program Files/GPAC/sdk/lib` is hardcoded — this will break on any system where GPAC is installed elsewhere (or not installed at all). Our CI builds use vcpkg for dependencies.
> 
> 3. **CRT strategy conflict**: Our CI sets `RUSTFLAGS: -Ctarget-feature=+crt-static` for static CRT linking. This PR switches to dynamic CRT (`/MD`) and adds `/NODEFAULTLIB:libcmt` to suppress conflicts — that directly contradicts our build strategy and will cause issues when the Rust static libs (built with static CRT) are linked against C code built with dynamic CRT.
> 
> 4. **Fragile `/NODEFAULTLIB:libcmt`**: This is a workaround that masks the real problem rather than fixing it. Mixed CRT linkage leads to subtle runtime bugs (heap corruption when memory is allocated in one CRT and freed in another).
> 
> 5. **The `-lm -lpthread -ldl` Unix guard**: This part is the only unambiguously correct change — these libs don't exist on Windows and should be guarded. But it should be submitted separately.
> 
> Please consider splitting out the Unix library guard into its own PR. The CRT and GPAC linking changes need a fundamentally different approach that aligns with how our CI actually builds.

Thanks for the guidence I'll file my first pr seperately addressing the last issue

cfsmp3:
Closing this PR. The one unambiguously correct change (the Unix library guard for `-lm -lpthread -ldl`) has been merged via #2156 as requested.

The remaining changes in this PR — dynamic CRT switching (`/MD`), hardcoded GPAC path (`C:/Program Files/GPAC/sdk/lib`), and `/NODEFAULTLIB:libcmt` — conflict with our build strategy (static CRT via `RUSTFLAGS: -Ctarget-feature=+crt-static`, vcpkg for dependencies) and would cause mixed CRT linkage issues. These need a fundamentally different approach and shouldn't be pursued in their current form.

Thanks for splitting out the guard into a separate PR!

PR #2147, opened 2026-02-26
[FIX] Prevent MSVC cross-CRT invalid free on output_filename

### LINKED ISSUE #2126: [Bug] CCExtractor 0.96 x64 Debug build displays error message at end of runs on Windows 11 PC
To reproduce this bug, 

1.  Unzip [CCExtractor Windows x64 Debug build](https://github.com/CCExtractor/ccextractor/actions/runs/22192745021/artifacts/5577190458) into a new folder
2.  Copy a video file containing cc into that folder also
3.  Open a Terminal window in the new folder and execute:
      ` .\ccextractorwinfull.exe .\<name of video file>` 
4.  Note that the expected .srt files are created, but the below error message is displayed also.

Image

The same bug occurs also in the x86 Debug package and may be demonstrated in the same way as above.

Atul-Chahar:

---

## Summary

Resolves the Windows MSVC Debug build crash reported in #2126. The root cause was a cross-CRT boundary bug: `ccx_options.enc_cfg.output_filename` is allocated by Rust (`CString::into_raw`), but C code was calling `freep()` on it — triggering the MSVC Debug allocator's invalid-free assertion.

## Fix

`update_encoder_list_cinfo()` and `segment_output_file()` now work with a local copy (`struct encoder_cfg local_cfg`) for dynamic multi-program file allocations, keeping the global Rust-allocated pointer untouched. The crash-inducing `freep()` call in `dinit_libraries()` was also removed.

## Changes

- `src/lib_ccx/general_loop.c` — use local `encoder_cfg` copy for multi-program output
- `src/lib_ccx/lib_ccx.c` — remove invalid `freep()` on Rust-allocated string, fix cleanup path

## CI Notes

The 9 Linux sample platform failures (`--autoprogram`, `--out=spupng`, `--startcredits` tests) are pre-existing on upstream master and unrelated to this PR. The Windows sample platform confirms all 237/237 tests pass and this PR actually fixes those same 9 tests on Windows.

Fixes #2126

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,6 +1,7 @@
 0.96.7 (unreleased)
 -------------------
 - Fix: Prevent crash in Rust timing module when logging out-of-range PTS/FTS timestamps from malformed streams.
+- Fix: Resolve Windows MSVC debug build crash caused by cross-CRT invalid free on Rust-allocated output_filename (#2126)
 
 0.96.6 (2026-02-19)
 -------------------
--- a/src/lib_ccx/general_loop.c
+++ b/src/lib_ccx/general_loop.c
@@ -1102,19 +1102,20 @@ void segment_output_file(struct lib_ccx_ctx *ctx, struct lib_cc_decode *dec_ctx)
 			ctx->segment_counter++;
 			if (enc_ctx)
 			{
+				struct encoder_cfg local_cfg = ccx_options.enc_cfg;
 				// list_del(&enc_ctx->list);
 				// dinit_encoder(&enc_ctx, t);
-				const char *extension = get_file_extension(ccx_options.enc_cfg.write_format);
+				const char *extension = get_file_extension(local_cfg.write_format);
 				// Format: "%s_%06d%s" needs: basefilename + '_' + up to 10 digits + extension + null
 				size_t needed_len = strlen(ctx->basefilename) + 1 + 10 + strlen(extension) + 1;
-				freep(&ccx_options.enc_cfg.output_filename);
-				ccx_options.enc_cfg.output_filename = malloc(needed_len);
-				if (!ccx_options.enc_cfg.output_filename)
+				local_cfg.output_filename = malloc(needed_len);
+				if (!local_cfg.output_filename)
 				{
 					fatal(EXIT_NOT_ENOUGH_MEMORY, "In segment handling: Out of memory allocating output filename.");
 				}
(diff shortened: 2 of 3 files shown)


cfsmp3 (changes requested):
Two issues:

1. **Unrelated SCC fps change**: The SCC framerate fix is a duplicate of #2146. Please remove it from this PR.

2. **Double free claim not substantiated**: I ran valgrind on master with `--multiprogram` and there is no "Invalid free" reported. Code analysis shows `init_encoder` always calls `strdup` on the filename, so `freep` in the cleanup path is correct — it frees the strdup'd copy, not the original. Removing the `freep` actually introduces a small memory leak (the strdup'd 2 bytes are never freed).

Please provide concrete evidence of the double free (e.g. MSVC Debug output, valgrind trace, or a specific code path that skips the strdup). Without that, this change is incorrect.

Atul-Chahar:
@cfsmp3  Thanks for the thorough review! I've updated the PR with an atomic fix addressing both of your points:

1. **Unrelated SCC fps change**: I have completely removed the SCC framerate changes as requested. They are now managed exclusively in #2146.
2. **Double free claim not substantiated**: You are entirely correct that `init_encoder` uses `strdup`, effectively isolating the memory and `wb->filename`. The true underlying mechanism for the "Invalid Free" assertions in the Windows Debug CRT actually stems from cross-boundary allocations between Rust and C. 

Specifically, `ccx_options.enc_cfg.output_filename` is initialized inside Rust via `CString::into_raw`. When `update_encoder_list_cinfo()` or `dinit_libraries()` (at the end of the run) blindly called `freep()` on this exact Rust-allocated pointer, the Windows MSVC allocator caught the mismatch between headers and crashed. The previous attempt to remove the `freep()` inadvertently allowed a `malloc()`'d string to mask the Rust pointer, stopping the crash but inherently introducing the memory leak you identified.

**The Correct Atomic Fix:** 
Instead of modifying and mutating the global `ccx_options.enc_cfg` strings, `update_encoder_list_cinfo` and `segment_output_file` now allocate dynamic multi-program strings via a strictly local copy (`struct encoder_cfg local_cfg = ccx_options.enc_cfg;`). This completely avoids clobbering the Rust string or calling C's `free()` on Rust strings, and allows `malloc()`'d strings to be cleanly `freep()`'d locally without leaks. I've also removed the crash-inducing `freep()` originally placed inside `dinit_libraries`.

Let me know what you think of this approach!

PR #2157, opened 2026-03-01
[BUG FIX] Add NULL check for fopen return value in process_hex()

Anayo-Anyafulu:
## Summary
In process_hex() in src/lib_ccx/general_loop.c, fopen() was called without checking if the returned FILE pointer is NULL. If the file cannot be opened, the subsequent fgets() call would crash with a NULL pointer dereference.

## Changes
- src/lib_ccx/general_loop.c — added NULL check after fopen(), prints an error message and returns early if the file cannot be opened.

## Testing
Manually verified the fix compiles cleanly.


cfsmp3:
Closing — #2202 covers the same fopen() NULL check plus also checks alloc_demuxer_data(). Thanks for identifying the issue though!

PR #2162, opened 2026-03-01
Fix/hardsubx error handling

### LINKED ISSUE #2158: [BUG] Unchecked strdup return values in hardsubx.c
## Summary
In init_hardsubx() in src/lib_ccx/hardsubx.c, strdup() is called twice on lines 235-236 without checking for NULL. If memory allocation fails, subsequent use of these pointers could lead to undefined behavior.

## Code
Line 235: char *pars_vec = strdup("debug_file");
Line 236: char *pars_values = strdup("/dev/null");

## Suggested Fix
Add NULL checks for both strdup calls and free already allocated resources before returning NULL or calling fatal().

### LINKED ISSUE #2160: [BUG] Missing TessBaseAPIDelete() and NULL checks in _init_hardsubx() error paths
I was going through the `_init_hardsubx()` function in `src/lib_ccx/hardsubx.c` and  noticed a few problems with how errors are handled:

1. **`TessBaseAPICreate()` return value is never checked.** If it fails and returns NULL, every subsequent Tesseract call on that handle is going to blow up.

2. **When `tessdata_path` lookup fails** (lines around 244 and 255), the code frees `pars_vec`, `pars_values`, and `ctx`, but never calls `TessBaseAPIDelete()` on the handle that was already created causing a resource leak.

3. **When `TessBaseAPIInit4()` fails** (around line 278), the handle gets leaked because only `free(ctx)` is called without deleting the Tesseract handle first.

Basically any error path between `TessBaseAPICreate()` and the successful `TessBaseAPIInit4()` leaks the Tesseract handle, and before `TessBaseAPICreate()` we don't even know if the handle is valid.

**Steps to trigger:**
- Point CCExtractor at a system with no tessdata installed (or corrupt traineddata files)
- The tessdata probe paths will fail and hit these error returns
- Tesseract handle is leaked every time

**Suggested fix:**
- Add a NULL check right after `TessBaseAPICreate()`
- Add `TessBaseAPIDelete(ctx->tess_handle)` in each error path where the handle was created but Init hasn't succeeded yet
- After Init succeeds, use `TessBaseAPIEnd()` + `TessBaseAPIDelete()` (which the `dec_sub` malloc failure path already does correctly)

### LINKED ISSUE #2161: [BUG] hardsubx_process_data() has no return statement, missing NULL checks, and leaks resources on error
There are several issues in `hardsubx_process_data()` in `src/lib_ccx/hardsubx.c`:

1. **No return statement.** The function is declared as `int` but never returns anything. This will lead to undefined behavior if the caller ever tries to use the return value.

2. **`av_malloc()` result is not checked** . If allocation fails, the NULL pointer gets passed into `av_image_fill_arrays()` and later into frame processing. This might cause segfault.

3. **`sws_getContext()` result is not checked.** The code checks for NULL before calling `sws_freeContext()` at the end, but never checks after creation. The code doesn't guard against it even after possibility of NULL

4. **`init_encoder()` result is not checked.** The pointer goes into the processing functions and `dinit_encoder()` without any validation.

5. **Resource leaks in error paths.** Every `fatal()` call in the function just exits without cleaning up whatever was allocated before that point. For example, if `avcodec_open2()` fails, `codec_ctx` and `format_ctx` are leaked. If `av_frame_alloc()` fails, add frames to that list, etc.

**Suggested fix:**
- Use of a `goto cleanup` pattern with a centralized cleanup section at the end of the function. This avoids repeating cleanup code in every error path and makes it easy to add new allocations later without forgetting to clean them up. 
- add the missing NULL checks and a proper return value.

SuvidhJ:
**[FIX] Fix resource leaks, missing NULL checks, and missing return in hardsubx.c**

## What does this PR do?

Fixes several error-handling issues in `src/lib_ccx/hardsubx.c` across both `_init_hardsubx()` and `hardsubx_process_data()`.

## Changes in `_init_hardsubx()`

- **Added NULL check for `TessBaseAPICreate()`**: The return value was never validated. If it returned NULL, every subsequent Tesseract call would hit undefined behavior.

- **Added NULL checks for both `strdup()` calls**: `pars_vec` and `pars_values` were used without checking if `strdup()` succeeded. On memory allocation failure these would be NULL pointers passed into `TessBaseAPIInit4()`.

- **Added `TessBaseAPIDelete()` in error paths**: When `tessdata_path` lookup failed or `TessBaseAPIInit4()` returned non-zero, the code called `free(ctx)` without first deleting the Tesseract handle that was already created. This leaked the handle every time one of these error paths was hit.

## Changes in `hardsubx_process_data()`

- **Added missing return statement**: The function is declared as returning `int` but never actually returned a value, which is undefined behavior per the C standard.

- **Added NULL checks for `av_malloc()`, `sws_getContext()`, and `init_encoder()`**: All three can return NULL on failure, but the results were being used directly without any validation.

- **Replaced per-error cleanup with `goto cleanup` pattern**: Previously, each `fatal()` call would exit without freeing any of the FFmpeg resources allocated up to that point (codec context, format context, frames, buffers, etc.). Now all error paths jump to a centralized cleanup section that safely frees everything that was allocated.

## Issues Fixed

Fixes #2158
Fixes #2160 
Fixes #2161

## Testing

- Changes are limited to error-handling paths only . The normal execution flow is completely unchanged
- All cleanup uses NULL-safe functions (`free(NULL)` is a no-op per the C standard, `av_frame_free()` and `avcodec_free_context()` handle NULL pointers )
- No new dependencies or behavioral changes introduced

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,6 +1,7 @@
 0.96.7 (unreleased)
 -------------------
 - Fix: Prevent crash in Rust timing module when logging out-of-range PTS/FTS timestamps from malformed streams.
+- Fix: Fixed resource leaks, missing NULL checks, and missing return in hardsubx.c
 
 0.96.6 (2026-02-19)
 -------------------


cfsmp3:
Closing — the review feedback from March 7 (unreachable code after fatal(), remove dead goto/cleanup paths, squash commits) was never addressed. 3 weeks with no response. Feel free to reopen with the fixes.

PR #2163, opened 2026-03-02
Implement error handling to default language allocation in matroska.c

### LINKED ISSUE #2159: [BUG] Unchecked strdup return value in matroska.c
## Summary
In parse_segment_track_entry() in src/lib_ccx/matroska.c, strdup("eng") on line 866 is called without checking for NULL. If allocation fails, lang will be NULL and could cause issues if dereferenced before reassignment.

## Code
Line 866: char *lang = strdup("eng");

## Suggested Fix
Add a NULL check and call fatal(EXIT_NOT_ENOUGH_MEMORY, ...) if strdup returns NULL.

SuvidhJ:

---
**Summary**

In parse_segment_track_entry() (src/lib_ccx/matroska.c), the default language is set via strdup("eng"), but the return value is never checked for NULL. If allocation fails, lang would be NULL and could cause a segfault when it's later passed to strcmp(), memcpy(), or assigned to the track struct.

This PR adds the missing NULL checks so the program exits gracefully with a clear out-of-memory message instead of crashing unpredictably.

**What changed**

- initial default assignment: Added a NULL check with fatal(EXIT_NOT_ENOUGH_MEMORY, ...) immediately after char *lang = strdup("eng");.
- Inside the MATROSKA_SEGMENT_TRACK_LANGUAGE_IETF case (second strdup("eng")): Added the same NULL check after the fallback re-assignment of lang.

**Why this matters**
- Every allocation in this file : malloc, realloc, calloc, already follows the pattern of checking for NULL and calling fatal(). These two strdup() calls were the only ones missing that safety net. While an OOM on a 4-byte allocation is admittedly rare, it's an edge case that static analyzers flag and that can cause hard-to-diagnose crashes in memory-constrained environments.

Related Issue : Fixes #2159

### DIFF
--- a/src/lib_ccx/matroska.c
+++ b/src/lib_ccx/matroska.c
@@ -879,11 +879,9 @@ void parse_segment_track_entry(struct matroska_ctx *mkv_ctx)
 
 	ULLONG track_number = 0;
 	enum matroska_track_entry_type track_type = MATROSKA_TRACK_TYPE_VIDEO;
-	
 	char *lang = strdup("eng");
 	if (lang == NULL)
-    fatal(EXIT_NOT_ENOUGH_MEMORY, "In parse_segment_track_entry: Out of memory allocating default language.");
-	
+    (EXIT_NOT_ENOUGH_MEMORY, "In parse_segment_track_entry: Out of memory allocating default language.");
 	char *header = NULL;
 	char *lang_ietf = NULL;
 	char *codec_id_string = NULL;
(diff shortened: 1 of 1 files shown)


cfsmp3:
Closing — the first hunk is missing the `fatal` call. This:

```c
if (lang == NULL)
    (EXIT_NOT_ENOUGH_MEMORY, "In parse_segment_track_entry: Out of memory allocating default language.");
```

is just a comma operator expression that does nothing. It wouldn't even catch the error. Also uses spaces instead of tabs.

PR #2168, opened 2026-03-02
[FIX] Remove unconditional VBI_DEBUG define — add CMake option for opt-in debug builds

### LINKED ISSUE #2167: BUG: VBI_DEBUG unconditionally defined — debug code always compiled into production builds
## Summary

`#define VBI_DEBUG` is hardcoded unconditionally in 
`src/lib_ccx/ccx_decoders_vbi.h` line 5, meaning debug-only struct 
fields and code are always compiled into every production build.

## Affected Code
```c
// ccx_decoders_vbi.h:4
#define VBI_DEBUG   // ← always defined, should be a build option

// Results in these always being compiled in:
struct ccx_decoder_vbi_cfg {
#ifdef VBI_DEBUG
    char *debug_file_name;   // ← always present in every build
#endif
};

struct ccx_decoder_vbi_ctx {
#ifdef VBI_DEBUG
    FILE *vbi_debug_dump;    // ← always present in every build
#endif
};
```

## Comparison With Other Debug Flags

Other debug flags in the codebase are correctly left undefined unless 
explicitly set at compile time:
```c
// ccx_common_constants.h:205
#ifdef WTV_DEBUG   // ← never unconditionally defined, opt-in only
```

`VBI_DEBUG` is the only debug flag that is unconditionally always-on.

## Impact

- Extra struct fields (`debug_file_name`, `vbi_debug_dump`) are present 
  in every production build, wasting memory
- Debug I/O paths are always compiled in, increasing binary size
- Inconsistent with how all other debug flags work in the codebase

## Suggested Fix

Remove `#define VBI_DEBUG` from the header. Add it as a CMake option 
so developers can opt in:
```cmake
# CMakeLists.txt
option(VBI_DEBUG "Enable VBI debug output" OFF)
if(VBI_DEBUG)
    add_compile_definitions(VBI_DEBUG)
endif()
```

This matches how other optional debug flags work in the project.

Varadraj75:

---

## Summary

Fixes #2167 — `VBI_DEBUG` was unconditionally defined in `ccx_decoders_vbi.h`,
causing debug-only struct fields and code paths to always be compiled into
production builds.

## Root Cause
```c
// ccx_decoders_vbi.h:4 — before this fix
#define VBI_DEBUG   // ← always defined
```

This caused `debug_file_name` and `vbi_debug_dump` fields to be present in
every build of `ccx_decoder_vbi_cfg` and `ccx_decoder_vbi_ctx`, regardless
of whether debug output was needed.

## Fix

Removed the unconditional `#define VBI_DEBUG` from the header and added a
proper CMake option consistent with how other debug flags work in the project:
```cmake
option (VBI_DEBUG "Enable VBI decoder debug output" OFF)

if (VBI_DEBUG)
  add_definitions(-DVBI_DEBUG)
  message(STATUS "VBI debug output enabled")
endif (VBI_DEBUG)
```

Developers who need VBI debug output can now opt in explicitly:
```bash
cmake .. -DVBI_DEBUG=ON
```

## Comparison

Other debug flags in the codebase (`WTV_DEBUG`, `DEBUG_SAVE_TS_PACKETS`) are
never unconditionally defined — they require explicit opt-in. This fix brings
`VBI_DEBUG` in line with that pattern.

## Testing

Built and verified locally on macOS with both `VBI_DEBUG=OFF` (default)
and `VBI_DEBUG=ON`. All existing CI tests pass.

Fixes #2167

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,5 +1,6 @@
 0.96.7 (unreleased)
 -------------------
+- Fix: Remove unconditional VBI_DEBUG define — add CMake option -DVBI_DEBUG=ON for opt-in debug builds (#2167)
 - Fix: Remove strdup() memory leaks in WebVTT styling encoder, fix invalid CSS rgba(0,256,0) green value, fix missing free(unescaped) on write-error path (#2154)
 - Fix: Prevent crash in Rust timing module when logging out-of-range PTS/FTS timestamps from malformed streams.
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,11 +1,11 @@
 cmake_minimum_required (VERSION 3.24.0)
 project (CCExtractor)
-
 include (CTest)
 
 option (WITH_FFMPEG "Build using FFmpeg demuxer and decoder" OFF)
 option (WITH_OCR "Build with OCR (Optical Character Recognition) feature" OFF)
 option (WITH_HARDSUBX "Build with support for burned-in subtitles" OFF)
+option (VBI_DEBUG "Enable VBI decoder debug output" OFF)
 
 # Version number
 set (CCEXTRACTOR_VERSION_MAJOR 0)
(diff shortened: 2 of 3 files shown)


Varadraj75:
The Windows CI test failures are pre-existing and unrelated to this PR.

All failing tests show `Empty path name is not legal` — a known Windows 
test runner path issue that affects every PR. All these tests show 
"Last passed: Never" in the CI summary, confirming they were already 
broken on master before this change.

The previous Windows build failure (C2016 empty struct) has been fixed 
by adding a `reserved` member under `#ifndef VBI_DEBUG`. Windows now 
builds and runs cleanly — only the pre-existing path failures remain.

Linux CI passes all tests cleanly.

Varadraj75:
@cfsmp3 All three review points addressed:
- Indentation corrected to tabs
- Combined into single #ifdef/#else block as suggested
- Squashed into a single commit and rebased onto latest master

Ready for re-review.

PR #2173, opened 2026-03-04
[FIX] Use dynamic current_fps in CEA-708 SCC frame delay calculations

Atul-Chahar:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [x] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor.

---

## Summary

`dtvcc_write_scc()` in `ccx_decoders_708_output.c` uses hardcoded `1000 / 29.97` in 6 places to calculate frame delay offsets for CEA-708 SCC output. This causes incorrect subtitle timing on non-NTSC streams (25fps PAL, 23.976 film, 24fps, etc.).

This is the same class of bug fixed for CEA-608 in #2146, but in the CEA-708 (DTVCC) output path.

## Fix

Replaced all 6 instances of `1000 / 29.97` with `1000 / current_fps`. The `current_fps` global is already accessible via the existing include chain — no new headers needed.

## Changes

- `src/lib_ccx/ccx_decoders_708_output.c` — 6 lines changed in `dtvcc_write_scc()`
- `docs/CHANGES.TXT` — added changelog entry

Fixes #2172

PR #2184, opened 2026-03-07
[IMPROVEMENT] --out report now probes both EIA-608 fields by default

### LINKED ISSUE #2177: [PROPOSAL] Display both EIA-608 fields in --out report by default
#### Summary

By default, `--out report` will only report the presence of captions in the EIA-608 field 1.  Captions in field 2 (CC3/4) are not detected by default.  Currently, report will always explicitly state that there are no captions in CC3/4 unless the user passes `--output-field 'both'`.

The `--out report` is misleading to an end-user because it explicitly states there are no captions in CC3, whereas this is only the result of the defaults.

##### Proposal:
Either:
- a. report mode should not explicitly list CC channels which do not belong to the field being probed, or state `unknown`
- b. report mode should probe both EIA-608 fields by default `--output-field 'both'`. This is the more elegant solution suggested by @Varadraj75 at the end of this ticket.

#### Impact / Priority
- Low, because a known workaround exists & the functionality is all there today if the user passes the appropriate options.  From an end user perspective, it is a minor usability inconvenience.

#### Sample File & Test Case

Download sample (copying to TS, just because of an unrelated issue of MPEG2 in MKV, which is already resolved in master)

Run ccextractor on TS input...
```
$ ffmpeg -i "https://weather.horse/samples/cc3.mkv" -c 'copy' "./cc3.ts" -y

$ ccextractor --out report "./cc3.ts" 
File: ./cc3.ts
Stream Mode: Transport Stream
Program Count: 1
Program Numbers: 1 
PID: 256, Program: 1, MPEG-2 video
PID: 257, Program: 1, AC3 audio
//////// Program #1: ////////
DVB Subtitles: No
Teletext: No
ATSC Closed Caption: Yes
EIA-608: Yes
XDS: No
CC1: No
CC2: No
CC3: No <<<<<<<<<<<<<<<<<<<<<< CC3 is not detected
CC4: No
CEA-708: No
Width: 720
Height: 480
Aspect Ratio: 02 - 4:3
Frame Rate: 04 - 29.97

MPEG-4 Timed Text: No
```

#### Workaround

Specify both fields during execution...
```
$ ccextractor --output-field 'both' --out 'report' -o - "./cc3.ts"
File: ./cc3.ts
Stream Mode: Transport Stream
Program Count: 1
Program Numbers: 1 
PID: 256, Program: 1, MPEG-2 video
PID: 257, Program: 1, AC3 audio
//////// Program #1: ////////
DVB Subtitles: No
Teletext: No
ATSC Closed Caption: Yes
EIA-608: Yes
XDS: No
CC1: No
CC2: No
CC3: Yes <<<<<<<<<<<<<<<<<<<<<<< CC3 is now detected if --output-field 'both' is used
CC4: No
CEA-708: No
Width: 720
Height: 480
Aspect Ratio: 02 - 4:3
Frame Rate: 04 - 29.97

MPEG-4 Timed Text: No
```

#### Also, errors with --stdout

Workaround is to use `-o -` rather than `--stdout`
```
$ ccextractor --output-field 'both' --out 'report' --stdout "./cc3.ts"
Error: You can't extract both fields to stdout at the same time in broadcast mode.
```

#### Discussion from other ticket

https://github.com/CCExtractor/ccextractor/issues/2149#issuecomment-3977498803

@Varadraj75 wrote...
> the report mode behavior is genuinely misleading.
> `CC3: No` when the captions are actually there is worse than not showing CC3
> at all, since it could make users conclude the file has no CC3 content.
> 
> I agree this deserves a separate ticket. The cleanest fix is probably to have
> `--out report` internally use `--output-field both` for its analysis pass,
> with the broadcast mode conflict handled gracefully (just suppress the error
> when in report-only mode since there's no actual `--stdout` conflict).

@Varadraj75 I hope you are not offended in me resurfacing your previous reply and creating a separate, independent issue.

Atul-Chahar:

---

## Summary

`--out report` only probes EIA-608 **field 1** by default, so CC3/CC4 always show `No` even when captions exist in field 2. This is misleading — a user sees `CC3: No` and concludes the file has no CC3 content, when in reality the report just never looked at field 2.

## Root Cause

In the `OutFormat::Report` handler in `parser.rs`, `extract` is never set — it defaults to `1` (field 1 only). CC3/CC4 belong to field 2, so they are never decoded and the report always says `No` for them.

## Fix

1. **Set `extract = 12`** (both fields) and `is_608_enabled = true` in the report handler so both fields are probed.
2. **Exempt `OutputFormat::Null`** from the broadcast-mode stdout conflict check at the validation stage — report mode sets `write_format = Null`, so there is no actual stdout output and the "can't extract both fields to stdout" error does not apply.

## Changes

- `src/rust/src/parser.rs` — 3 locations:
  - Report handler: added `extract = 12` + `is_608_enabled = true`
  - Stdout conflict check: added `write_format != Null` guard
  - Tests: updated `test_out_report_enables_file_reports`, added `test_out_report_with_stdout_no_conflict`

## Precedence

If a user explicitly passes `--output-field 1` alongside `--out report`, their choice wins — `--output-field` is parsed after `--out` and overwrites `extract`.

**AI disclosure:** 🟡 AI-assisted — I reviewed, tested, and understand every line.

Fixes #2177

### DIFF
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -2,6 +2,7 @@
 -------------------
 - Fix: Prevent crash in Rust timing module when logging out-of-range PTS/FTS timestamps from malformed streams.
 - Fix: Use dynamic current_fps instead of hardcoded 29.97 in CEA-708 SCC frame delay calculations (#2172)
+- Improvement: --out report now probes both EIA-608 fields by default, so CC3/CC4 are correctly detected without requiring --output-field both (#2177)
 
 0.96.6 (2026-02-19)
 -------------------
--- a/docs/CHANGES.TXT
+++ b/docs/CHANGES.TXT
@@ -1,6 +1,7 @@
 0.96.7 (unreleased)
 -------------------
 - Fix: Prevent crash in Rust timing module when logging out-of-range PTS/FTS timestamps from malformed streams.
+- Fix: Use dynamic current_fps instead of hardcoded 29.97 in CEA-708 SCC frame delay calculations (#2172)
 
 0.96.6 (2026-02-19)
 -------------------
--- a/src/lib_ccx/ccx_decoders_708_output.c
+++ b/src/lib_ccx/ccx_decoders_708_output.c
@@ -550,11 +550,11 @@ void dtvcc_write_scc(dtvcc_writer_ctx *writer, dtvcc_service_decoder *decoder, s
 	if (tv->old_cc_time_end > time_show.time_in_ms)
 	{
 		// Correct the frame delay
-		time_show.time_in_ms -= 1000 / 29.97;
+		time_show.time_in_ms -= 1000 / current_fps;
 		print_scc_time(time_show, buf);
 		buf_len = strlen(buf);
 		SCC_SNPRINTF("\t942c 942c");
-		time_show.time_in_ms += 1000 / 29.97;
+		time_show.time_in_ms += 1000 / current_fps;
 		// Clear the buffer and start pop on caption
 		SCC_SNPRINTF("94ae 94ae 9420 9420");
 	}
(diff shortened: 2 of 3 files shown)


Atul-Chahar:
Tested locally on Linux:

**Build:**
```
$ cd linux && ./build
Compilation successful, no compiler messages.
```

**Version:**
```
$ ./ccextractor --version
CCExtractor 0.96.5, Git commit: 1c8bac37, Compilation date: 2026-03-06
```

**Rust tests:**
```
$ cd src/rust && cargo test
test result: ok. 399 passed; 0 failed; 0 ignored
```

**Format check:**
```
$ cd src/rust && cargo fmt --check
(no output — all formatted correctly)
```

**New tests added:**
- `test_out_report_enables_file_reports` — updated to verify `extract == 12` and `is_608_enabled == true`
- `test_out_report_with_stdout_no_conflict` — verifies report mode + `--stdout` does not trigger broadcast mode error

Atul-Chahar:
### Manual verification — CC3 detection now works with the fix: 

screenshot-2026-03-07_07-12-54

Atul-Chahar:
Yes! I tested it against standard NTSC files that only contain CC1. Probing CC3/CC4 does not cause any harm or crashes; it gracefully returns empty streams as expected, while ensuring we capture all data on streams that do utilize both fields.

cfsmp3:
Closing due to inactivity — #2173 was merged over a month ago and this just needed a rebase, but there's been no response. Feel free to reopen with a rebased branch if you'd like to continue.

PR #2185, opened 2026-03-07
Fix: Add NULL check after strdup in parse_segment_track_entry

sadiapeerzada:
strdup("eng") may return NULL if memory allocation fails.

This patch adds a NULL check after strdup and calls
fatal(EXIT_NOT_ENOUGH_MEMORY) to prevent a potential
NULL pointer dereference.


cfsmp3:
Closing — this is superseded by #2193 which covers the same strdup NULL check along with additional cleanup of the IETF language handling in the same function. Thanks for the contribution.

PR #2170, opened 2026-03-04
feat: replace GPAC with FFmpeg for MP4 demuxing 

DhanushVarma-2:
**In raising this pull request, I confirm the following (please check boxes):**

- [ ] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [ ] I have checked that another pull request for this purpose does not exist.
- [ ] I have considered, and confirmed that this submission will be valuable to others.
- [ ] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [ ] I give this submission freely, and claim no ownership to its content.
- [ ] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor.

---
Adds a Rust FFmpeg-based MP4 demuxer (rsmpeg) as a compile-time alternative to GPAC. Enabled with `cmake -DWITH_FFMPEG=ON`.

**Supported:** AVC/H.264, HEVC/H.265, CEA-608, CEA-708, tx3g, chapter extraction
**Known gap:** VobSub detected but not decoded
**GPAC path:** Unchanged, still the default

Note: `mp4_ffmpeg_exports.rs` shows as binary in the diff due to null-terminated C string literals for FFI — the file is clean UTF-8.
```

PR #2186, opened 2026-03-07
fix: add link_directories for tesseract/leptonica on macOS

DhanushVarma-2:
Fixes linker failure when building with -DWITH_OCR=ON -DWITH_HARDSUBX=ON on macOS arm64 (Homebrew).

pkg_check_modules populates TESSERACT_LIBRARY_DIRS and LEPTONICA_LIBRARY_DIRS but these were never passed to the linker. Added link_directories() calls in both the OCR and HARDSUBX sections of CMakeLists.txt.

Tested on macOS arm64 with Homebrew tesseract 5.5.1 and leptonica 1.86.0.


DhanushVarma-2:
> The CMake link_directories fix is correct and we'd like to merge it — but the PR also includes unrelated mp4.c changes (ENABLE_FFMPEG_MP4 track counting) that belong in #2170.
> 
> Please remove the mp4.c changes from this PR so we can merge the tesseract/leptonica fix cleanly.

Done 👍

PR #2191, opened 2026-03-08
[IMPROVEMENT] feat(mp4): add FFmpeg/libavformat backend for MP4 demuxing

gaurav02081:
[IMPROVEMENT] **feat(mp4): add FFmpeg/libavformat backend for MP4 demuxing**

---

### Add optional FFmpeg-based MP4 parser as an alternative to GPAC

This PR introduces an alternative MP4 parsing backend using **FFmpeg's libavformat**, while keeping the existing **GPAC-based implementation** unchanged and as the default.

#### Motivation

In a previous discussion (Gsoc meeting 2 MARCH) we talked about updating the GPAC dependency used for MP4 processing in CCExtractor. One suggestion was to explore whether there is a Debian-friendly alternative rather than only focusing on upgrading GPAC.

FFmpeg is already used in other parts of the codebase (for example in the demuxing/decoding integration and in the HardsubX module), so extending its use for MP4 parsing seemed like a reasonable option to explore.

#### Implementation

A new implementation (`mp4_ffmpeg.c`) was added which uses **FFmpeg's libavformat** to open and parse MP4 containers.

The general workflow is:

* Open the MP4 container with `avformat_open_input()`
* Discover streams using `avformat_find_stream_info()`
* Read packets sequentially using `av_read_frame()`
* Dispatch packets based on stream type

Video packets (H.264 / HEVC) are passed to the existing `do_NAL()` processing logic, while caption tracks (CEA-608 / CEA-708) and subtitle tracks (tx3g) continue to use the existing CCExtractor parsing functions.

One difference from the GPAC implementation is that FFmpeg reads **packets sequentially across all streams**, whereas the GPAC implementation reads **samples per track**. The downstream caption extraction pipeline remains unchanged.

For H.264 / HEVC streams, codec configuration data is obtained from the stream extradata (`avcC` / `hvcC`) in order to determine the NAL unit length prefix size and extract SPS/PPS before processing packets.

#### Build configuration

This backend is **optional** and controlled through a compile-time flag:

```
-DUSE_FFMPEG_MP4=ON
```

* Default build → uses GPAC (`mp4.c`)
* FFmpeg build → uses the new implementation (`mp4_ffmpeg.c`)

The runtime behavior of CCExtractor remains unchanged — the difference only affects how the MP4 container is parsed internally.

#### Summary

This PR:

* Adds an FFmpeg-based MP4 parser
* Keeps GPAC as the default implementation
* Introduces a compile-time option to switch between the two backends
* Leaves the caption extraction pipeline unchanged

This provides a potential alternative MP4 backend using a widely available multimedia framework while preserving the existing behavior.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -103,6 +103,26 @@ jobs:
       working-directory: build
     - name: Display version information
       run: ./build/ccextractor --version
+  cmake_ffmpeg_mp4:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install dependencies
+      run: sudo apt update && sudo apt-get install libgpac-dev libtesseract-dev libleptonica-dev libavcodec-dev libavformat-dev libavutil-dev libavfilter-dev libavdevice-dev libswresample-dev libswscale-dev
+    - uses: actions/checkout@v6
+    - name: cmake (default GPAC build)
+      run: mkdir build && cd build && cmake ../src
+    - name: build (default GPAC build)
+      run: make -j$(nproc)
+      working-directory: build
+    - name: Display version information (GPAC build)
+      run: ./build/ccextractor --version
+    - name: cmake (FFmpeg MP4 build)
+      run: mkdir build_ffmpeg && cd build_ffmpeg && cmake -DWITH_FFMPEG=ON -DWITH_OCR=ON -DWITH_HARDSUBX=ON ../src
+    - name: build (FFmpeg MP4 build)
+      run: make -j$(nproc)
+      working-directory: build_ffmpeg
+    - name: Display version information (FFmpeg MP4 build)
+      run: ./build_ffmpeg/ccextractor --version
   cmake_ocr_hardsubx:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/build_mac.yml
+++ b/.github/workflows/build_mac.yml
@@ -107,6 +107,21 @@ jobs:
       working-directory: build
     - name: Display version information
       run: ./build/ccextractor --version
+  cmake_ffmpeg_mp4:
+    runs-on: macos-latest
+    steps:
+    - uses: actions/checkout@v6
+    - name: Install dependencies
+      run: brew install pkg-config gpac ffmpeg tesseract leptonica
+    - name: cmake (FFmpeg MP4 build)
+      run: |
+        mkdir build && cd build
+        cmake -DWITH_FFMPEG=ON -DWITH_OCR=ON -DWITH_HARDSUBX=ON ../src
+    - name: build
+      run: make -j$(nproc)
+      working-directory: build
+    - name: Display version information
+      run: ./build/ccextractor --version
   build_shell_hardsubx:
     # Test build.command with -hardsubx flag (burned-in subtitle extraction)
     runs-on: macos-latest
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -194,20 +194,25 @@ if (PKG_CONFIG_FOUND AND WITH_FFMPEG)
   pkg_check_modules (AVCODEC REQUIRED libavcodec)
   pkg_check_modules (AVFILTER REQUIRED libavfilter)
   pkg_check_modules (SWSCALE REQUIRED libswscale)
-
-  set (EXTRA_LIBS ${EXTRA_LIBS} ${AVFORMAT_LIBRARIES})
-  set (EXTRA_LIBS ${EXTRA_LIBS} ${AVUTIL_LIBRARIES})
-  set (EXTRA_LIBS ${EXTRA_LIBS} ${AVCODEC_LIBRARIES})
-  set (EXTRA_LIBS ${EXTRA_LIBS} ${AVFILTER_LIBRARIES})
-  set (EXTRA_LIBS ${EXTRA_LIBS} ${SWSCALE_LIBRARIES})
+  pkg_check_modules (SWRESAMPLE REQUIRED libswresample)
 
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${AVFORMAT_INCLUDE_DIRS})
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${AVUTIL_INCLUDE_DIRS})
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${AVCODEC_INCLUDE_DIRS})
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${AVFILTER_INCLUDE_DIRS})
   set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${SWSCALE_INCLUDE_DIRS})
+  set (EXTRA_INCLUDES ${EXTRA_INCLUDES} ${SWRESAMPLE_INCLUDE_DIRS})
 
   set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DENABLE_FFMPEG")
+  set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DENABLE_FFMPEG_MP4")
+
+  # NOTE: FFmpeg libs are added to EXTRA_FFMPEG_LIBS (not EXTRA_LIBS) so we
+  # can place them AFTER ccx_rust in the link order. GNU ld processes
+  # left-to-right; ccx_rust (containing rsmpeg) references FFmpeg symbols
+  # like swr_get_out_samples, so FFmpeg shared libs must come after it.
+  set (EXTRA_FFMPEG_LIBS ${AVFORMAT_LIBRARIES} ${AVUTIL_LIBRARIES}
+       ${AVCODEC_LIBRARIES} ${AVFILTER_LIBRARIES} ${SWSCALE_LIBRARIES}
+       ${SWRESAMPLE_LIBRARIES})
 endif (PKG_CONFIG_FOUND AND WITH_FFMPEG)
 
 ########################################################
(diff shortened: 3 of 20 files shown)


gaurav02081:
Update: Added a new CI job to build CCExtractor with the optional FFmpeg MP4 backend.

The workflow now performs two builds:

Default build using the GPAC implementation

FFmpeg build using -DUSE_FFMPEG_MP4=ON

Both builds run --version to verify the binaries execute correctly, and separate build directories are used to avoid CMake cache conflicts.

cfsmp3:
Thanks for the PR — the implementation is clean and well-structured, and the dedicated CI job is a nice touch.

However, we're actively moving the codebase from C to Rust, so we can't accept new C modules. We'd need the FFmpeg MP4 demuxer to be implemented in Rust. See #2170 for an example of how this can be done using rsmpeg (Rust FFmpeg bindings) with a thin C bridge for the callbacks into the existing C code.

If you'd like to rework this in Rust we'd be happy to review it. The overall approach (avformat_open_input → av_read_frame → dispatch by stream type) is sound, it's just the implementation language that needs to change.

gaurav02081:
FFmpeg-based implementation written in Rust using rsmpeg, activated via -DWITH_FFMPEG=ON. The GPAC path
  remains the default and is completely untouched when FFmpeg is not enabled.

  Architecture

  The implementation follows the same 3-layer pattern as PR #2170:

=  Layer 1 — Rust Core (src/rust/src/demuxer/mp4.rs)
  - Opens MP4 via rsmpeg::AVFormatContextInput
  - Classifies streams: AVC/H.264, HEVC/H.265, CEA-608, CEA-708, tx3g
  - Parses SPS/PPS/VPS from codec extradata
  - Main av_read_frame loop dispatches packets to C bridge functions

=  Layer 2 — FFI Exports (src/rust/src/mp4_ffmpeg_exports.rs)
  - ccxr_processmp4() — replaces processmp4()
  - ccxr_dumpchapters() — replaces dumpchapters()

=  Layer 3 — C Bridge (src/lib_ccx/mp4_rust_bridge.c)
  - Flat FFI-safe wrappers around existing C processing functions
  - ccx_mp4_process_avc_sample() → NAL parsing via do_NAL()
  - ccx_mp4_process_hevc_sample() → NAL parsing + store_hdcc() flush
  - ccx_mp4_process_cc_packet() → CEA-608 via process608(), CEA-708 via ccdp_find_data() + ccxr_dtvcc_process_data()
  - ccx_mp4_process_tx3g_packet() → 3GPP timed text
  - No GPAC structs cross the FFI boundary — only primitive types

  Build System

`  cmake -DWITH_FFMPEG=ON -DWITH_OCR=ON -DWITH_HARDSUBX=ON ../src`

  - `ENABLE_FFMPEG_MP4` compile flag gates all new code
  - Cargo feature enable_mp4_ffmpeg pulls in rsmpeg
  - Corrosion passes the feature to Cargo when WITH_FFMPEG is on
  - FFmpeg libs are added as INTERFACE_LINK_LIBRARIES of ccx_rust to ensure correct GNU ld link order
  - Bridge functions use --undefined linker flags (same pattern as existing decode_vbi/do_cb/store_hdcc)

  Files

  New (5):
 
  
  ```
 src/rust/src/demuxer/mp4.rs         Rust FFmpeg demuxer core      
   src/rust/src/mp4_ffmpeg_exports.rs  C-callable FFI exports        
   src/lib_ccx/mp4_rust_bridge.c       Thin C bridge                 
   src/lib_ccx/mp4_rust_bridge.h       Bridge header                 
   src/lib_ccx/ccx_gpac_types.h        Minimal GPAC-compatible types 
```

  Modified (13):```
 Cargo.toml, build.rs, lib.rs, demuxer/mod.rs, wrapper.h, rust/CMakeLists.txt, src/CMakeLists.txt, lib_ccx/CMakeLists.txt, ccx_mp4.h,
  ccextractor.c, build_linux.yml, build_mac.yml
```

  Testing

  - GPAC-only build (cmake ../src) — compiles and links, no regression
  - FFmpeg build (cmake -DWITH_FFMPEG=ON -DWITH_OCR=ON -DWITH_HARDSUBX=ON ../src) — compiles and links
  - Runtime: ./ccextractor tests/samples/BBC1.mp4 produces identical output to GPAC path
  - CI: Linux and macOS cmake_ffmpeg_mp4 jobs added

gaurav02081:
@cfsmp3 Hi, 

this PR is failing in the Linux environment (in SP) so i tested it and found The CI Linux (sample platform) failures are pre-existing on master and not introduced by this PR. 

All MP4 tests (3/3) pass.

cfsmp3 (changes requested):
## Comprehensive Comparison Review (Apr 18)

We tested this PR alongside #2170 (DhanushVarma-2), which implements the same FFmpeg MP4 demuxer feature. Both were built with `-DWITH_FFMPEG=ON -DWITH_OCR=ON -DWITH_HARDSUBX=ON` and tested against all 36 MP4/MOV samples in our collection.

### Key finding: both PRs produce byte-identical output

#2170 and #2191 use the same rsmpeg backend and produce exactly the same output on all 36 samples. Neither is functionally superior to the other.

### Critical: caption extraction gaps vs GPAC

The FFmpeg path is NOT a drop-in replacement for GPAC. 7 samples lose captions:

| Sample | GPAC | FFmpeg (#2170 and #2191) | Issue |
|--------|------|-------------------------|-------|
| `132d7df7e993.mov` | 108KB | 104B | CEA-608 in H.264 SEI — no separate subtitle track |
| `1974a299f050.mov` | 127KB | 104B | Same — captions embedded in video NALs |
| `99e5eaafdc55.mov` | 164KB | 104B | Same |
| `8849331ddae9.mp4` | 48KB (2352 lines, clean) | 6KB (227 lines, garbled `÷` chars) | c608 track exists but decoded as raw bytes, not through CEA-608 decoder |
| `b2771c84c2a3.mp4` | 2.6KB (130 lines) | 284B (11 lines) | Same c608 garbling |
| `5df914ce773d.mp4` | 1KB | 0B | c608 track present but entirely missed |
| `1f3e951d516b.mp4` | 5KB | 0B | dvdsub (bitmap) in MP4 — not supported |

One sample works BETTER with FFmpeg:
| `ad9f9e03240e.m4v` | 0B | 51KB | FFmpeg extracts what GPAC can't |

The remaining 28 samples produce identical output (including 0B on both).

### Three bugs to fix

**Bug 1 — CEA-608 in H.264 SEI (3 MOV samples):** These files have no separate subtitle track — captions are embedded as SEI user data in the H.264 video stream NAL units. GPAC handles this via `process_avc_sample()` which parses NALs. The FFmpeg demuxer needs to do the same — scan AVC samples for SEI-embedded CEA-608 data.

**Bug 2 — c608 track garbled (2 samples):** The c608 subtitle track exists and is found, but the data comes out as raw CC bytes (`÷÷ ÷ ÷HI÷, ÷TH÷ER÷E.÷`) instead of decoded text. The data needs to go through the CEA-608 character decoder, not be dumped as raw bytes.

**Bug 3 — c608 track missed (1 sample):** `5df914ce773d.mp4` has a c608 track visible to ffprobe but FFmpeg extracts nothing. Investigate why.

(dvdsub/bitmap in MP4 can be documented as a known limitation for now.)

### PR-specific issues for #2191

Compared to #2170:

- **More verbose** (1019 lines vs 437) — more code to maintain for identical functionality
- **12 commits** with many linker fix iterations (vs 5 clean commits in #2170) — suggests the build integration was iterated rather than designed
- **`Cargo.lock` in diff** — should not be committed, remove it
- **Still has `ENABLE_HARDSUBX` coupling** — `WITH_FFMPEG` still defines `ENABLE_HARDSUBX` in `lib_ccx/CMakeLists.txt`. Our #2259 (merged) fixed this on master. You need to rebase and drop the coupling.
- **CI job is a plus** — the `cmake_ffmpeg_mp4` job in both Linux and Mac workflows is genuinely useful. #2170 doesn't have this.
- **CMake approach is cleaner** — `INTERFACE_LINK_LIBRARIES` on the ccx_rust target is the right CMake idiom vs #2170's `--no-as-needed` hack
- **Bridge is much larger** (306 lines vs 54) — worth investigating if the extra code adds value or is redundant

### Samples tested

All 36 MP4/MOV/M4V files from `~/media_samples/completed/` and `~/media_samples/failed/`, including files up to 4.5GB. No files were skipped.

gaurav02081:
@cfsmp3 Thanks for the detailed review. 
 
 I have addressed all three reported issues and the PR-specific points.

----------------------------------------------------
###  BUG fix 

**== Bug 1** — CEA-608 in H.264 SEI (MOV samples)**
Fixed by draining `sub->got_output`at EOF in the FFmpeg path.
All three MOV samples are now **byte-identical** to GPAC.

**== Bug 2** — c608 garbled output (MP4 samples)**
Two issues were handled:

  1. Strip ISO BMFF wrapper `(`cdat/cdt2/ccdp`)` before decoding
  2. Handle `cc_data` triplet format (`[cc_info][b1][b2]`) directly via `process608` the correct field routing

Results:

  * `b2771c84c2a3.mp4` → **byte-identical**
  * `8849331ddae9.mp4` → identical content and caption count (uniform ~2 ms timing offset noted below)

**== Bug 3** — c608 track missed**
  Covered by triplet handling → now **byte-identical**

* **dvdsub (bitmap subtitles)**
Not supported (same as GPAC). Documented in the `mp4.rs`module doc comment.

-----------------------------------------------------

### PR cleanup (used git squash to bifurcate the commits into 3 parts)

* Commit history reduced **12 → 3 clean commits**
* `Cargo.lock` cleaned (`libc` dependency removed via `std::fs`)
* `ENABLE_HARDSUBX` decoupled (rebased onto latest master)
* CI job for FFmpeg MP4 build kept and passing (Linux + macOS)
* `INTERFACE_LINK_LIBRARIES` approach retained

------------------------------------------------------

### Bridge design note

I kept the FFmpeg bridge **self-contained**, calling only public CCExtractor APIs (`do_NAL`, `process608`, `process_cc_data`, etc.), without exposing GPAC internals.

An alternative approach (as in #2170) reduces bridge size by reusing `mp4.c 'static helpers', but that couples the FFmpeg path back to GPAC. I preferred keeping the **two backends independent**. 

------------------------------------------------------

### Known difference vs GPAC

* One sample (`8849331ddae9.mp4`) has a **uniform ~2 ms timestamp offset**
* Caption content, count, and ordering are identical
* Rooted in shared `set_fts "timing logic" (not MP4-specific)

Fixing this would require changes in shared timing infrastructure, so I have left it as-is for now.

-------------------------------------------------------

### `sampleplatform.ccextractor.org` CI status

The two red `test/9266` (Linux) and `test/9267` (Windows) checks show the same failing categories as master (same ratios as recent merged PRs like #2266 / #2267). The bot comment on this PR also notes these failures are present on master.
**MP4**, the category this PR implements, passes on both sides. All GitHub Actions CI checks are green.

--------------------------------------------------------

### Verification

* Tested against all 6 samples from the review in my local
* 5/6 byte-identical, 1 with minor timing offset
* No caption loss in any sample

---

Looking forward to your feedback.

cfsmp3:
### Timestamp note

We investigated the ~2ms timing offset on `8849331ddae9.mp4`. **GPAC is correct, FFmpeg is wrong.**

The MP4 container's `stts` box has DTS=113113 for the first subtitle sample (timescale 30000). GPAC reads this directly → 3771ms. FFmpeg reports PTS=113050 (63 ticks short) → 3769ms. The loss happens inside `av_rescale_q()` when FFmpeg rescales timestamps internally.

This affects all 492 subtitle timestamps uniformly (~2ms early). Content is identical. Full analysis documented in our repo at `plans/FFMPEG_TIMESTAMP_ROUNDING.md`. May be worth reporting upstream to FFmpeg.

Not a blocker for this PR — just documenting.

gaurav02081:
Thanks for the review, will raise an issue for this in ffmpeg.

PR #2192, opened 2026-03-09
feat(rust): add --codec isdb CLI flag for ISDB/ARIB subtitle selection

pranavshar223:
Extends CCXCodec enum with Isdb variant and wires it to Codec::IsdbCc in the demuxer config, enabling users to explicitly select ISDB/ARIB subtitle streams via:
  --codec isdb
  --no-codec isdb

The underlying C decoder (isdbsub_decode) and Rust Codec::IsdbCc were already present but had no CLI
entry point. This change completes the wiring.

Relevant for Japanese broadcast (ARIB B-24) support.

---
## Summary
Extends `CCXCodec` enum with `Isdb` variant and wires it to
`Codec::IsdbCc` in the demuxer config, enabling users to explicitly
select ISDB/ARIB subtitle streams via:
  --codec isdb
  --no-codec isdb

## Motivation
The underlying C decoder (`isdbsub_decode`) and Rust `Codec::IsdbCc`
were already present but had no CLI entry point. This change completes
the wiring for Japanese broadcast (ARIB B-24) support.

## Testing
- `--codec isdb` is now accepted without error
- `--no-codec isdb` correctly excludes the ISDB stream
- All existing codec tests pass

### DIFF
--- a/src/rust/src/args.rs
+++ b/src/rust/src/args.rs
@@ -754,6 +754,7 @@ pub struct Args {
     ///     nothing is selected and no subtitle is generated
     /// --codec teletext
     ///     select the teletext subtitle from elementary stream
+    /// --codec isdb      select ISDB/ARIB subtitle stream (Japanese broadcast)
     #[arg(long, verbatim_doc_comment, value_name="value", help_heading=OUTPUT_AFFECTING_CODEC)]
     pub codec: Option<CCXCodec>,
     /// --no-codec dvbsub
@@ -977,6 +978,7 @@ pub struct Args {
 pub enum CCXCodec {
     Dvbsub,
     Teletext,
+    Isdb,
 }
 
 #[derive(Display, Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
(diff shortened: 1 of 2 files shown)


cfsmp3:
The code change looks correct — small and clean wiring.

A few questions before we can move forward:

1. **How did you test this?** The testing section says `--codec isdb` is accepted without error, but that just means the CLI parser doesn't reject it. Did you actually test with an ISDB/ARIB sample? If so, which specific sample file did you use?

2. **What's the end-to-end use case?** Once a user passes `--codec isdb`, what happens? Does ccextractor actually produce subtitle output from an ISDB stream? What output format? Can you show us the actual output from a real file?

3. **Without this flag, what happens with an ISDB stream?** Is the codec auto-detected, and this flag is just for forcing selection when there are multiple subtitle streams? Or does ISDB processing not work at all without this flag?

We'd like to understand the practical value before merging — the wiring is correct but we need to know it actually does something useful end-to-end.

cfsmp3:
Closing — no response in 35 days to the request for testing proof. Feel free to reopen with a concrete test scenario if you'd like to continue.

Issue #1492, opened 2023-03-10
[BUG] Incorrect path for loading tesseract traineddata


ibrahim-akrab:
CCExtractor version: 0.94
# Necessary information

- Is this a regression (i.e. did it work before)? {NO}
- What platform did you use? {Linux}
- What were the used arguments? `{}`

# Video links

[channel5-2018-02-12.ts](https://drive.google.com/file/d/1Etq-pv5G3jGqVhhRl7cNrfuw4gaKkLoV/view?usp=sharing) from the TV Samples page

# Additional information

ccextractor tries to load tesseract traineddata from a wrong location then blames it on the TESSDATA_PREFIX. Here's the output it produces:
```
Opening file: /home/ibrahim/Downloads/channel5-2018-02-12.ts
File seems to be a transport stream, enabling TS mode
Analyzing data in general mode
Error opening data file /usr/share/eng.traineddata
Please make sure the TESSDATA_PREFIX environment variable is set to your "tessdata" directory.
Failed loading language 'eng'
Tesseract couldn't load any languages!
Failed TessBaseAPIInit4 -1
```
I checked the logic in `ocr.c` and found that `probe_tessdata_location` works fine by tracing the syscalls it makes to each possible tessdata location by running `strace -e trace=openat ./ccextractor ~/Downloads/channel5-2018-02-12.ts` and the result is as follows:
```
Opening file: /home/ibrahim/Downloads/channel5-2018-02-12.ts
openat(AT_FDCWD, "/home/ibrahim/Downloads/channel5-2018-02-12.ts", O_RDONLY) = 3
File seems to be a transport stream, enabling TS mode
Analyzing data in general mode
openat(AT_FDCWD, "./tessdata/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/share/tessdata/", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 4
openat(AT_FDCWD, "/usr/share/eng.traineddata", O_RDONLY|O_CLOEXEC) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/usr/share/eng.traineddata", O_RDONLY) = -1 ENOENT (No such file or directory)
Error opening data file /usr/share/eng.traineddata
Please make sure the TESSDATA_PREFIX environment variable is set to your "tessdata" directory.
[...]

PR #2257, opened 2026-04-04
fix: resolve TESSDATA_PREFIX path correctly for all Tesseract versions

DhanushVarma-2:
There were two bugs in init_ocr():

The Tesseract 4/5 branch always appended /tessdata to the probed path — if TESSDATA_PREFIX was already set to point at the tessdata dir itself, this doubled it: /usr/share/tessdata/tessdata.
The legacy Tesseract <4 branch passed the raw probed path to TessBaseAPIInit4 with no /tessdata appended at all — so Tesseract looked for /usr/share/eng.traineddata instead of /usr/share/tessdata/eng.traineddata. This is the exact error in #1492.

Fix: build the tessdata path once before both branches — check if the path already ends with tessdata, otherwise append it. Windows backslash separators handled too. Both branches now use the same resolved path. Added an mprint line showing the resolved path to make future debugging easier.

### DIFF
--- a/src/lib_ccx/ocr.c
+++ b/src/lib_ccx/ocr.c
@@ -261,20 +261,46 @@ void *init_ocr(int lang_index)
 	}
 
 	ctx->api = TessBaseAPICreate();
+
+	/* Build the correct tessdata path for TessBaseAPIInit4.
+	 * probe_tessdata_location() returns a base dir e.g. "/opt/homebrew/share/".
+	 * TessBaseAPIInit4 expects "<base>/tessdata" as the data path.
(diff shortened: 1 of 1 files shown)


DhanushVarma-2:
The format_rust CI failures are pre-existing on master .
 src/rust/lib_ccxr/src/teletext.rs has a formatting issue unrelated to this PR. No files in src/rust/ were modified in this change.

DhanushVarma-2:
@cfsmp3 can you review it.

cfsmp3:
Seem abandoned and don't even pass format

PR #2261, opened 2026-04-11
Fix memory safety bug in copy_demuxer_data_from_rust

THE-Amrit-mahto-05:

Reason for this PR:

Sanity check:

---

### Problem

The function `copy_demuxer_data_from_rust` is documented to copy buffer contents, but the implementation was incorrectly reassigning the buffer pointer:

```rust
(*c_data).buffer = rust_data.buffer as *mut c_uchar;
```

This leads to:
- Use-after-free if the Rust buffer is deallocated
- Dangling pointers in the C layer
- Corrupted demuxed data (affecting subtitle extraction)

### Reproduction 

1. Build the project:
```bash
mkdir build && cd build
cmake ..
make -j
```

2. Run with any TS/MKV input that triggers Rust → C demuxer data transfer.
3. Before this fix:
- Buffer pointer is reassigned instead of copied
- Leads to undefined behavior depending on memory lifecycle
- Tests do not guarantee actual memory copy into C buffer

### Fix
- Replaced pointer reassignment with safe memory copy using:
 ```rust 
std::ptr::copy_nonoverlapping
```

- Ensured safe copy using:
```rust
min(source_len, dest_len)
```

- Added null pointer checks
- Set length to 0 when buffers are invalid

### Test Improvements

Previously, the tests for copy_demuxer_data_from_rust only validated high-level properties such as buffer length. They did not verify whether the buffer contents were actually copied into the destination C memory.

As a result, the incorrect implementation:

```rust
(*c_data).buffer = rust_data.buffer as *mut c_uchar; 
```

could still pass the tests despite not performing a real memory copy.

### Changes Made
The tests have been improved to explicitly validate the contents of the destination buffer. For example:
```rust
assert_eq!(small_c_buffer, vec![0x42; 100]);
```

### Impact
- Ensures that data is properly copied into C-managed memory
- Detects incorrect pointer reassignment behavior
- Prevents similar issues in future changes
- Aligns test coverage with the documented behavior of the function

### DIFF
--- a/src/rust/src/libccxr_exports/demuxerdata.rs
+++ b/src/rust/src/libccxr_exports/demuxerdata.rs
@@ -4,7 +4,6 @@ use crate::ctorust::FromCType;
 use crate::demuxer::common_types::CcxRational;
 use crate::demuxer::demuxer_data::DemuxerData;
 use lib_ccxr::common::{BufferdataType, Codec};
-use std::os::raw::c_uchar;
 use std::os::raw::{c_int, c_uint};
 
 /// Convert from C demuxer_data to Rust DemuxerData
@@ -254,7 +253,11 @@ mod tests {
         };
 
         unsafe {
+            let original_ptr = c_data.buffer;
             copy_demuxer_data_from_rust(&mut c_data, &rust_data);
+
+            // Verify the pointer was NOT reassigned (the bug we're fixing)
+            assert_eq!(c_data.buffer, original_ptr);
 
             // Verify all fields were copied correctly
             assert_eq!(c_data.program_number, rust_data.program_number);
--- a/src/rust/src/libccxr_exports/demuxerdata.rs
+++ b/src/rust/src/libccxr_exports/demuxerdata.rs
@@ -256,7 +256,6 @@ mod tests {
             let original_ptr = c_data.buffer;
             copy_demuxer_data_from_rust(&mut c_data, &rust_data);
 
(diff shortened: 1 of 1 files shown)


src/rust/src/libccxr_exports/demuxerdata.rs
```diff
@@ -245,8 +253,12 @@ mod tests {
         };
 
         unsafe {
+            let original_ptr = c_data.buffer;
             copy_demuxer_data_from_rust(&mut c_data, &rust_data);
 
+            // Verify the pointer was NOT reassigned (the bug we're fixing)
```
cfsmp3: Please remove AI comments.

cfsmp3 (changes requested):

THE-Amrit-mahto-05:
@cfsmp3  done can you review it

cfsmp3:
Closing because this is not reachable.

 copy_demuxer_data_from_rust() has no callers outside its own unit tests. Searching the tree, every reference is either the definition itself or one of the seven test call sites in the
  same file. Nothing in the demuxer path invokes it.

  That is why the request for a specific sample never got anywhere: there isn't an input that reaches this code, so the pointer reassignment can't produce a use-after-free or corrupt
  anyone's output today. We've closed a few PRs on this basis before (#2263, #2238, #2239) — hardening code that no caller can reach adds risk without removing any.

PR #2268, opened 2026-04-19
[Fix](608): restore current_visible_start_ms set-on-first-char for rollup

gaurav02081:

Reason for this PR:

Sanity check:

Repro instructions:

---
### ## Summary

Fixes an issue where CEA-608 roll-up captions were emitted with a start time of `00:00:00,000 " instead of the actual first-character FTS.

This occurs in roll-up mode streams where fewer than the required number of lines are written before EOF (or more generally, when no CR event results in `changes == true`).

**Before:** `00:00:00,000 --> ... "
**After:**  `00:00:13,913 --> ...` (matches first-character FTS)

---

### ## Root cause

Commits `bc5d6055` / `54df50f4` introduced a `rollup_from_popon` flag in `ccx_decoder_608_context` to delay setting `current_visible_start_ms` until a CR event triggered scrolling.

However, in cases where:
- no CR produces `changes == true`, or  
- the stream ends before enough lines are written (e.g., roll-up-2),

the flag is never cleared. As a result, `current_visible_start_ms` remains `0`, and captions are emitted starting at `00:00:00,000`.

In contrast, commit `042716ad` (pre-change) set `current_visible_start_ms` immediately when the first character was written to an empty buffer (in non-POPON mode), which correctly handles these cases.

---

## Fix

Remove the `rollup_from_popon` mechanism and restore the previous behaviour of setting the start time on the first character.

### Changes

- `src/lib_ccx/ccx_decoders_608.c`
  - Remove `rollup_from_popon` guard from character handler  
  - Restore original CR handling logic  
  - Remove related initialization

- `src/lib_ccx/ccx_decoders_608.h`
  - Remove `rollup_from_popon` field

- Cleanup:
  - Remove debug traces (`dbg_print`, `printf`) from:
    - `ccx_decoders_common.c`
    - `sequencing.c`
    - `avc_functions.c`
  - No functional changes in these files

---

## Test plan

-  Default build passes (`cmake --build .`) with no new warnings  
-  Reproducer sample:
  - Start time corrected (`00:00:13,913` vs `00:00:00,000`)
  - Caption content, count, and ordering unchanged vs pre-`bc5d6055` behaviour.

### DIFF
--- a/src/lib_ccx/ccx_decoders_608.c
+++ b/src/lib_ccx/ccx_decoders_608.c
@@ -150,6 +150,7 @@ ccx_decoder_608_context *ccx_decoder_608_init_library(struct ccx_decoder_608_set
 	data->my_channel = channel;
 	data->have_cursor_position = 0;
 	data->rollup_from_popon = 0;
+	data->ts_first_char_rollup_transition = -1;
 	data->output_format = output_format;
 	data->cc_to_stdout = cc_to_stdout;
 	data->textprinted = 0;
@@ -246,6 +247,12 @@ void write_char(const unsigned char c, ccx_decoder_608_context *context)
 			context->cursor_column++;
 		if (context->ts_start_of_current_line == -1)
 			context->ts_start_of_current_line = get_fts(context->timing, context->my_field);
+		// First char after a pop-on -> roll-up transition: remember its FTS so
+		// write_cc_buffer can use it if EOF fires before a scrolling CR.
+		// Unlike ts_start_of_current_line, this field is NOT touched by
+		// intermediate CR commands (changes=0 CRs overwrite ts_start_of_current_line).
+		if (context->rollup_from_popon && context->ts_first_char_rollup_transition == -1)
+			context->ts_first_char_rollup_transition = get_fts(context->timing, context->my_field);
 		context->ts_last_char_received = get_fts(context->timing, context->my_field);
 	}
 }
@@ -310,6 +317,17 @@ int write_cc_buffer(ccx_decoder_608_context *context, struct cc_subtitle *sub)
 	if (context->mode == MODE_FAKE_ROLLUP_1 && // Use the actual start of data instead of last buffer change
 	    context->ts_start_of_current_line != -1)
 		context->current_visible_start_ms = context->ts_start_of_current_line;
+
+	// Pop-on -> roll-up transition that never saw a scrolling CR (e.g. EOF
+	// with fewer lines than the roll-up window). The CR handler never ran,
+	// so back-fill current_visible_start_ms from the first-char FTS instead
+	// of emitting a caption starting at 0.
+	if (context->rollup_from_popon && context->ts_first_char_rollup_transition > 0)
+	{
+		context->current_visible_start_ms = context->ts_first_char_rollup_transition;
+		context->rollup_from_popon = 0;
+		context->ts_first_char_rollup_transition = -1;
+	}
 
 	start_time = context->current_visible_start_ms;
 	end_time = get_visible_end(context->timing, context->my_field);
@@ -758,6 +776,7 @@ void handle_command(unsigned char c1, const unsigned char c2, ccx_decoder_608_co
 				// Start time will be set when CR causes scrolling (matching FFmpeg behavior)
 				context->rollup_from_popon = 1;
 				context->ts_start_of_current_line = -1;
+				context->ts_first_char_rollup_transition = -1;
 			}
 			erase_memory(context, false);
 
@@ -817,6 +836,7 @@ void handle_command(unsigned char c1, const unsigned char c2, ccx_decoder_608_co
 				{
 					context->current_visible_start_ms = context->ts_start_of_current_line;
 					context->rollup_from_popon = 0;
+					context->ts_first_char_rollup_transition = -1;
 				}
 
 				// Only if the roll up would actually cause a line to disappear we write the buffer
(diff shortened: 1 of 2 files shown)


cfsmp3:
What sample did you use to test it? I haven't been able to repro with any of the ones I have locally

gaurav02081:
here is the sample i used.

### * File: [c4dd893cb9d67be50f88bdbd2368111e16b9d1887741d66932ff2732969d9478.ts](https://github.com/user-attachments/files/26906356/c4dd893cb9d67be50f88bdbd2368111e16b9d1887741d66932ff2732969d9478.ts)

* Duration: ~15s (roll-up-2 never reaches required lines before EOF)

**Command:**
`ccextractor sample.ts -o out.srt`

On master:
1
00:00:00,000 --> 00:00:14,980
>> WHICH OF THESE STORIES WILL
YOU BE TALKING ABOUT TRO

This PR:

1
00:00:13,913 --> 00:00:14,980
>> WHICH OF THESE STORIES WILL
YOU BE TALKING ABOUT TRO

Diff (only the start timestamp changes; caption text, end time, and cue count remain identical):

- 00:00:00,000 --> 00:00:14,980
+ 00:00:13,913 --> 00:00:14,980

cfsmp3 (changes requested):
## Deep Review — Regressions Found

The rollup timing fix works on your sample (`c4dd893cb9d6`): start time corrects from `00:00:00,000` to `00:00:13,913`. That bug is real.

However, the broad revert of commits `bc5d6055`/`54df50f4` (slice_header, do_cb, sequencing changes) introduces timing regressions on multiple other samples. We tested against reference files we have on disk:

| Sample | Format | Master (matches reference) | This PR | Delta |
|--------|--------|---------------------------|---------|-------|
| `0069dffd2180.mpg` | MPG CEA-608 | `00:00:03,603` | `00:00:00,766` | **-2.8s** |
| `5cbb21adb6e0.dvr-ms` | DVR-MS CEA-608 | `00:00:00,534` | `00:00:00,734` | +200ms |
| `6395b281adf0.asf` | ASF CEA-608 | `00:00:02,436` | `00:00:01,068` | **-1.4s** |
| `8849331ddae9.mp4` | MP4 c608 | `00:00:03,771` | `00:00:04,738` | +1s |
| `132d7df7e993.mov` | MOV CEA-608 | `00:00:12,812` | `00:00:13,980` | +1.2s |
| `b22260d065ab.ts` | TS CEA-608 | `00:00:04,838` | `00:00:02,635` | **-2.2s** |

These are seconds-level shifts, not rounding. Master matches the reference files; this PR does not.

### Root cause

The reverted commits were fixing real timing issues across container formats:
- **slice_header IDR-only flush** — flushing on every reference frame (as this PR restores) causes reordering problems in I/P-only streams
- **do_cb field counting guards** — the H.264/PES guards on `cb_field1++` prevent double-counting in container formats where PTS already represents the correct timestamp
- **sequencing.c PES reset** — removing PES from `reset_cb` breaks timing for MPEG-2 PES streams

### What needs to happen

The rollup `00:00:00,000` bug is real and should be fixed, but with a **narrower fix** that only addresses the `rollup_from_popon` edge case. Specifically:

1. Keep the slice_header changes (IDR-only flush, large PTS gap handling) — do NOT revert
2. Keep the do_cb field counting guards — do NOT revert  
3. Keep the sequencing.c PES reset — do NOT revert
4. Keep the `get_visible_start`/`get_visible_end` Rust calls — do NOT revert
5. Fix ONLY the `rollup_from_popon` logic so it handles the case where no CR fires before EOF

The bug is that `rollup_from_popon` never clears when the stream ends with fewer lines than the roll-up window requires. A targeted fix would be to clear `rollup_from_popon` and set `current_visible_start_ms` during the end-of-stream flush path, not by reverting the entire mechanism.

### Test commands

```bash
# The bug (should NOT be 00:00:00,000):
ccextractor c4dd893cb9d6...ts -o out.srt

# Regressions (should match reference):
ccextractor 0069dffd2180...mpg -o out.srt   # expect 00:00:03,603
ccextractor 5cbb21adb6e0...dvr-ms -o out.srt # expect 00:00:00,534
ccextractor 6395b281adf0...asf -o out.srt    # expect 00:00:02,436
```

gaurav02081:
Thanks for your time :-)
 
I restored all three commits I had reverted. I added a precise fix just for the EOF edge case.                                                                                                                         
                                                                                                                                                      
Reshaped the branch to a single commit (1d225172).                                                                                                  
                                                                                                                                                      
### Fix
                                                                                                                                                      
Added a new field, ts_first_char_rollup_transition, that captures the first-char FTS right after the popon→rollup switch. Unlike ts_start_of_current_line, it doesn't get clobbered by changes=0 CRs in between, so it survives to the end-of-stream flush. write_cc_buffer uses it to backfill current_visible_start_ms when rollup_from_popon is still set at emit time. The normal CR path is untouched.
                                                                                                                                                      Numbers
                                                                                                                                                      
c4dd893c still fixed: first caption at 00:00:13.913 (was 00:00:00.000).

### Ran all six regression samples you listed = every one matches the master exactly:
                                          
 ```
0069dffd.mpg    Expected: 00:00:03,603 | Got: 00:00:03,603
5cbb21ad.dvr-ms Expected: 00:00:00,534 | Got: 00:00:00,534
6395b281.asf    Expected: 00:00:02,436 | Got: 00:00:02,436
8849331d.mp4    Expected: 00:00:03,771 | Got: 00:00:03,771
132d7df7.mov    Expected: 00:00:12,812 | Got: 00:00:12,812
b22260d0.ts     Expected: 00:00:04,838 | Got: 00:00:04,838
```

Issue #2273, opened 2026-05-21
[Bug] Use-After-Free in switch_output_file due to dangling pointer when get_basename fails


RigelYoung:
A Use-After-Free (UAF) vulnerability exists in the `switch_output_file function` within `src/lib_ccx/ccx_encoders_common.c`.

When switching output files, the existing `enc_ctx->out->filename` is freed. However, if the subsequent call to `get_basename()` returns `NULL` (e.g., due to memory allocation failure or a NULL `original_filename`), the pointer `enc_ctx->out->filename` is never reassigned. This leaves a dangling pointer that is later passed to `write_subtitle_file_header()` and ultimately dereferenced in `spunpg_init()`, leading to a Use-After-Free.

**Execution Flow / Root Cause**:

1. Free: In `switch_output_file`, the memory pointed to by `enc_ctx->out->filename` is freed.

```c
if (enc_ctx->out->filename != NULL)
{
    free(enc_ctx->out->filename); // [1] Memory is freed here
    close(enc_ctx->out->fh);
}
```

2. Missing Reassignment: The function attempts to generate a new basename. If `get_basename()` returns `NULL`, the `if (basename != NULL)` block is skipped. `enc_ctx->out->filename` is not updated and remains a dangling pointer.

```c
char *basename = get_basename(enc_ctx->out->original_filename);
   if (basename != NULL) // [2] If basename is NULL, reassignment is skipped
   {
       enc_ctx->out->filename = create_outfilename(basename, suffix, ext);
       ...
   }
```

3. Use: The dangling pointer is immediately passed down the call stack and used.

```c
// [3] write_subtitle_file_header is called with the context containing the dangling pointer
   write_subtitle_file_header(enc_ctx, enc_ctx->out); 
   // Trace: write_subtitle_file_header -> write_spumux_header -> spunpg_init (UAF triggered here)
```

Invinsible-Coder:
I would like to work on this issue

Invinsible-Coder:
I have submitted a PR to fix this issue

cfsmp3:
Closing after analysis. The dangling pointer is real as a code pattern, but it is unreachable in practice:

- `init_write()` calls `fatal()` if the `strdup()` for `original_filename` fails, so `original_filename` is never NULL by the time `switch_output_file()` runs.
[...]

PR #2280, opened 2026-06-12
Fix output filename handling in encoder context

Invinsible-Coder:

Reason for this PR:

Sanity check:

Repro instructions:

---

{pull request content here}


cfsmp3:
Thanks for the PR. Closing along with #2286 and issue #2273: the underlying bug is only reachable on malloc failure (`init_write()` fatals if the `original_filename` strdup fails, so `get_basename()` can only return NULL under out-of-memory). No sample can trigger it, and project policy requires a reproducible bug with a sample.

One technical note from the review: on the failure path this patch leaves `out->fh = -1` while processing continues, so subsequent captions for the track would be silently dropped — keeping the previous file open (as #2286 did) avoids the data loss. Worth keeping in mind for future fixes. Real, reproducible bugs are very welcome.

Invinsible-Coder:
Thanks for the reply I will keep it in mind
I am new to contributing Hope I can help with something useful

PR #2286, opened 2026-06-26
fix: resolve dangling pointer in switch_output_file

### LINKED ISSUE #2273: [Bug] Use-After-Free in switch_output_file due to dangling pointer when get_basename fails

TacoJr01:
Free old filename/fh only after new file is successfully opened. Uses prepare-then-swap pattern to avoid partial state where encoder has no valid output. Fixes #2273

Reason for this PR:

Sanity check:

Repro instructions:

---

Fixes #2273

Problem:

In switch_output_file, enc_ctx->out->filename was freed before confirming a new file could be successfully opened. If get_basename() or open() failed, the function would proceed with a dangling pointer being passed to write_subtitle_file_header(), causing undefined behavior.

Fix:

Uses a prepare-then-swap approach — the new filename and file descriptor are fully prepared first. The old values are only freed and replaced after the new file is confirmed open. If any step fails, the function returns early leaving the encoder context in its original valid state.
Testing:

Tested on multiprogram_spain.ts (multiprogram transport stream with 7 programs) — processed all programs without crashing and produced expected output.

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.c
+++ b/src/lib_ccx/ccx_encoders_common.c
@@ -1352,35 +1352,47 @@ unsigned int get_font_encoded(struct encoder_ctx *ctx, unsigned char *buffer, in
 
 void switch_output_file(struct lib_ccx_ctx *ctx, struct encoder_ctx *enc_ctx, int track_id)
 {
-	// Do nothing when in "report" mode
 	if (enc_ctx == NULL || enc_ctx->out == NULL)
+		return;
+
+	const char *ext = get_file_extension(ctx->write_format);
+
+	char suffix[32];
+
+	snprintf(suffix, sizeof(suffix), "_%d", track_id);
+	// Prepare new values BEFORE freeing old ones
+	char *basename = get_basename(enc_ctx->out->original_filename);
+
+	if (basename == NULL)
+		return;
+	char *new_filename = create_outfilename(basename, suffix, ext);
+
+	free(basename);
+
+	if (new_filename == NULL)
+		return;
+	int new_fh = open(new_filename, O_RDWR | O_CREAT | O_TRUNC | O_BINARY, S_IREAD | S_IWRITE);
+
+	if (new_fh == -1)
 	{
+		free(new_filename);
 		return;
 	}
-
+	// Only now is it safe to free and replace the old values
 	if (enc_ctx->out->filename != NULL)
-	{ // Close and release the previous handle
-		free(enc_ctx->out->filename);
-		close(enc_ctx->out->fh);
-	}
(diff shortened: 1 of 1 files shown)


cfsmp3:
Thanks for the PR — this was reviewed in depth, including building it and testing against a dual-AVC-track MP4 (outputs byte-identical to master on the reachable path, valgrind clean). Of the three PRs filed for #2273, yours had the best approach: prepare-then-swap is the right structure, it never leaves a transient bad state, and it degrades without losing data.

However, we're closing all PRs for #2273: the bug is only reachable on malloc failure. `init_write()` fatals if the `original_filename` strdup fails, so `get_basename()` here can only return NULL under out-of-memory. No sample can trigger it, and project policy is that fixes require a reproducible bug with a sample (the PR template checkbox about a sample existing applies).

For future reference, two style points from the review: the project requires braces on all `if` bodies (this diff removed some existing ones), and please don't drop existing comments. Real, reproducible bugs are very welcome — hope to see you around.

PR #2291, opened 2026-07-05
fix(608): only apply pop-on→roll-up EOF fallback when start time is unset

x15sr71:

Reason for this PR:

Sanity check:

Repro instructions:

This PR fixes [Regression Test 11](https://sampleplatform.ccextractor.org/test/9351), which fails on master on both Windows and Linux. On Linux, it currently shows a separate “No output generated but there should be” issue that is still under investigation; Reference test run - [Windows](https://sampleplatform.ccextractor.org/test/9351), [Linux](https://sampleplatform.ccextractor.org/test/9350).

---

## Problem

A 608 regression caused incorrect cue start times on a WTV sample, while subtitle text and end times remained correct.

## Root cause

[`689b27ce`](https://github.com/CCExtractor/ccextractor/commit/689b27ce2f0bf9c5ffac54ca7f14ef0ef2c642f5) added an EOF fallback for pop-on -> roll-up transitions by back-filling `current_visible_start_ms` from `ts_first_char_rollup_transition`.

That fallback was too broad: if `rollup_from_popon` was still set when `write_cc_buffer()` ran, it could overwrite an already valid `current_visible_start_ms` with the first-character FTS.

## Fix

Only apply the EOF fallback when `current_visible_start_ms` is still unset:

```c
if (context->rollup_from_popon &&
    context->ts_first_char_rollup_transition > 0 &&
    context->current_visible_start_ms <= 0)
```

This keeps the EOF fix from [`689b27ce`](https://github.com/CCExtractor/ccextractor/commit/689b27ce2f0bf9c5ffac54ca7f14ef0ef2c642f5) while avoiding clobbering valid cue start times on normal paths.

## Testing

- Reproduced the regression on the affected WTV sample - `611b4a9235c08....`
- Applied the guard above in `write_cc_buffer()`.
- Rebuilt CCExtractor and compared the generated SRT against the reference file.
- Verified that the diff between current output and the reference file (`2af8168380d0...`) is clean.

### DIFF
--- a/src/lib_ccx/ccx_decoders_608.c
+++ b/src/lib_ccx/ccx_decoders_608.c
@@ -317,18 +317,18 @@ int write_cc_buffer(ccx_decoder_608_context *context, struct cc_subtitle *sub)
 	if (context->mode == MODE_FAKE_ROLLUP_1 && // Use the actual start of data instead of last buffer change
 	    context->ts_start_of_current_line != -1)
 		context->current_visible_start_ms = context->ts_start_of_current_line;
-
 	// Pop-on -> roll-up transition that never saw a scrolling CR (e.g. EOF
 	// with fewer lines than the roll-up window). The CR handler never ran,
 	// so back-fill current_visible_start_ms from the first-char FTS instead
 	// of emitting a caption starting at 0.
-	if (context->rollup_from_popon && context->ts_first_char_rollup_transition > 0)
+	if (context->rollup_from_popon &&
+	    context->ts_first_char_rollup_transition > 0 &&
+	    context->current_visible_start_ms <= 0)
 	{
 		context->current_visible_start_ms = context->ts_first_char_rollup_transition;
 		context->rollup_from_popon = 0;
 		context->ts_first_char_rollup_transition = -1;
 	}
-
 	start_time = context->current_visible_start_ms;
 	end_time = get_visible_end(context->timing, context->my_field);
 	sub->type = CC_608;


cfsmp3:
I applied this change onto current master (the branch base is from July, so building it directly would confound the comparison) and compared RT11's output against its approved baseline. It moves away from the baseline rather than toward it:

| | mismatching cue times vs approved |
|---|---|
| master | 3 |
| master + this PR | 4 |

The regression is cue 25:

```
approved:  00:01:52,446 --> 00:01:57,583
master:    00:01:52,446 --> 00:01:57,583   (matches)
this PR:   00:01:25,219 --> 00:01:57,583   (32 seconds)
```

Master already gets that cue exactly right, and the guard makes it start 2 ms after cue 24 ends and run for 32 seconds — for `[ BIRDS CHIRPING ] / [ DOG BARKS ]`. Cue 26 also moves further away, from 68 ms off to 3.2 s off.

The underlying reasoning is right — a fallback that overwrites an already-valid `current_visible_start_ms` is wrong — but `current_visible_start_ms <= 0` is not the condition that distinguishes the two cases. On this sample the value is already set at the point the fallback runs, and the fallback was what made it correct.

One process note: the platform runs cited (9350/9351) predate the baseline fix in sample-platform#1175, so at the time a PR's red/green on the suite was partly determined by which run preceded it. Worth re-running against a current baseline before drawing conclusions from those numbers.

Happy to look again if you have a sample where the current guard demonstrably helps — the diff itself is small and the intent is clearly right.

@x15sr71

x15sr71:
Agreed, closing — I'd already left this in draft with the same doubt, and your numbers settle it.

Thanks for re-running against a current baseline — that's conclusive. You're right that `current_visible_start_ms <= 0` isn't the discriminator; the value is already set by the time the fallback runs, so the guard skips the assignment that was making cue 25 correct.

Appreciate the offer to look again, but I don't have a sample where the guard demonstrably helps.

I'm still working the underlying timing issue, specifically RT84 and RT11, which both fail on current master. I have a candidate and I'm validating it against the ~165 regression tests I have samples for on my Linux box; if it holds up across the suite I'll open a PR right away.

PR #2295, opened 2026-07-14
[FIX] Correct typo: hat was -> that was

bbayles:

Reason for this PR:

Sanity check:

Repro instructions:

1. Execute `ccextractor -h | grep -i " hat was"`
2. Observe that this string is printed to stdout: `are processing video hat was split with a editing`

---

This PR fixes three instances of the same typo.


bbayles:
I don't think the test failures are related to the changes here.

PR #1138, opened 2019-12-07
[FEATURE] Simple MacOS GUI

sampoder:
- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.

**My familiarity with the project is as follows (check one):**

- [x] I have used CCExtractor just a couple of times.

---

I have created a simple GUI for MacOS. It was generated using the open-source Platypus tool created by Sveinbjörn Þórðarson after I saw a blog post about it and felt it would help CCExtractor. 

The reason I feel that this feature should be added is because the command line can be complex for some people. A simple GUI like this is perfect for these people as all they have to do it drag in a file. However, for more advanced users, they can still use the command line. I have tried to make the install process as simple as possible

Maybe if there is interest in a more advanced GUI I can try and custom develop one.

PR #2297, opened 2026-07-23
[FIX] macOS GUI silently reports "Process Complete" even when CCExtractor is missing or extraction fails

### LINKED ISSUE #2296: [BUG] macOS GUI always reports "Process Complete", even when CCExtractor is missing or extraction fails
**CCExtractor version:** 0.96.5
**Regression:** NO
**Platform:** Mac (Platypus-based GUI droplet)
**Arguments used:** None default GUI settings, drag-and-drop
**Video/sample links:** N/A bug is in the GUI wrapper script itself, reproduces with any input file

### Description

The macOS GUI (`mac/gui/src/CCExtractor.app`, added in #1138) is a Platypus droplet that calls `/usr/local/bin/ccextractor $1` and then unconditionally displays "Process Complete. Check the source file's folder for the subtitles." regardless of what actually happened. The script never checks whether `ccextractor` exists / is on `PATH`, or the exit code returned by the run.

As a result, all of the following silently present as "success":

- CCExtractor binary missing or not installed at the expected path
- Extraction failing outright
- A captionless input file (CCExtractor exits with code 10 / `EXIT_NO_CAPTIONS`)

Users are left hunting for output files that were never created, with no indication anything went wrong.

**Additional bug:** `$1` is unquoted in the script, so file paths containing spaces break the invocation entirely.

### Steps to reproduce

1. Open the CCExtractor macOS GUI app
2. Drag in any video file (or trigger the run with the CLI missing/misconfigured)
3. Observe the app reports "Process Complete" even when no `.srt`/output file is produced

### Expected behavior

The GUI should check the exit code of the `ccextractor` call and surface a meaningful message success, failure, or "no captions found" instead of a hardcoded success string.

### Screen recording / video

No build script present -> NO ERROR

https://github.com/user-attachments/assets/ac2de0af-0501-43c8-a9af-2ee0cb5aa535

### Notes

I'm happy to submit a PR fixing this checking the binary's exit status, quoting `$1`, and surfacing the actual result to the user.

kaihere14:
## Summary

The macOS GUI (Platypus droplet, `mac/gui/src/CCExtractor.app`, added in #1138) called `/usr/local/bin/ccextractor $1` and then unconditionally displayed "Process Complete," regardless of whether the binary existed, the run succeeded, or the file had no captions at all. `$1` was also unquoted, breaking on file paths containing spaces.

## Changes

- Added shebang and defined a `CCEXTRACTOR` variable for clarity.
- Added a check for the existence/executability of the CCExtractor CLI at the expected path, with a user-friendly error dialog if it's missing.
- Added a check for missing input (no file dropped onto the app), with a clear error dialog instead of silently proceeding.
- Quoted `$1` throughout so file paths with spaces no longer break the script.
- Replaced the unconditional success dialog with exit-status handling based on `src/lib_ccx/ccx_common_common.h`:
  - `0` → "Process Complete. Check the source file's folder for the subtitles."
  - `10` (`EXIT_NO_CAPTIONS`) → "No captions were found in this file, so no subtitle file was created."
  - any other code → "CCExtractor failed (exit code $status). Run it from the Terminal to see the detailed error output."
- Script now exits with the same status code it received from CCExtractor, so failures propagate correctly instead of being masked.

Reason for this PR:

Sanity check:

## Repro instructions

Before the fix, the GUI showed "Process Complete" in all three of the following cases, with no indication anything had gone wrong. After the fix, each case shows the correct, distinct message and the script exits with the matching status code:

1. **Success case** — a video file with captions is dropped onto the app; CCExtractor runs, exits `0`, and the app correctly reports "Process Complete." (video attached)

https://github.com/user-attachments/assets/7f923196-0f53-4ad5-ab26-e0ce60fea83c

2. **No captions case** — a video file with no captions is dropped onto the app; CCExtractor exits `10` (`EXIT_NO_CAPTIONS`); the app now reports "No captions were found in this file, so no subtitle file was created" instead of falsely reporting success. (video attached)

https://github.com/user-attachments/assets/18e0a55a-0cd7-4ecf-be1d-c4b1b74be1b7

3. **Failure case** — reproduced by pointing `CCEXTRACTOR` at a bad/missing path (or by triggering a genuine extraction failure); the app now reports "CCExtractor failed (exit code $status)..." instead of "Process Complete." (video attached)

https://github.com/user-attachments/assets/d7ea1b33-40c8-44b6-9086-c50c5f0f926b

Closes #2296


cfsmp3 (changes requested):
Three things before this can go in.

  1. Handle all dropped files, not just `$1` (it's possible to drop more than one).
  2. Fall back to PATH for the binary location.  /usr/local/bin/ccextractor is Intel-Homebrew's prefix. On Apple Silicon, Homebrew installs to /opt/homebrew/bin
  3. The shipped artifacts still contain the old script. End users don't run mac/gui/src/scripts. They install through InstallCCExtractorMacGUI.app, which unpacks a bundled DMG. Both DMGs in the repo still embed CCExtractor.app with the old droplet script.

kaihere14:
@cfsmp3  Thanks for the review, all three points are addressed:

1. **Handle all dropped files**: the script now loops over `"$@"` instead of only reading `$1`, runs `ccextractor` on each file, and shows one consolidated dialog summarizing the whole batch (successes, no-captions-found, and failures separately).

2. **PATH fallback**: it now resolves the binary with `command -v ccextractor` first (respects the user's actual PATH), and only falls back to explicitly checking `/usr/local/bin` and `/opt/homebrew/bin` if that fails, so both Intel and Apple Silicon Homebrew prefixes are covered.

3. **Shipped artifacts**: both `mac/gui/src/CCExtractor.dmg` and `mac/gui/InstallCCExtractorMacGUI.app/Contents/Resources/CCExtractor.dmg` embedded a `CCExtractor.app` with the old droplet script. Regenerated both from the fixed app so the install flow (not just building from source) picks up the fix.

**Incidental finding** while regenerating those: the old installer DMG (`InstallCCExtractorMacGUI.app`'s bundled `CCExtractor.dmg`) was actually unmountable on current macOS. `hdiutil attach` failed with `"no mountable file systems"`. It's a legacy raw, unpartitioned HFS+ image, a format that no longer reliably mounts. The regenerated DMGs use a standard GUID-partitioned HFS+ image and mount/verify cleanly, so this fixes a second latent bug along the way.

Screenshots covering all five states (binary not found, no input file, success, extraction failure, and multiple dropped files with a mixed result):

Screenshot 2026-07-26 at 11 46 48 PM
Screenshot 2026-07-26 at 11 47 01 PM
Screenshot 2026-07-26 at 11 49 28 PM
Screenshot 2026-07-26 at 11 50 22 PM
Screenshot 2026-07-26 at 11 55 51 PM

cfsmp3:
@kaihere14 

this revision introduces a new bug.

  **Filenames go into the AppleScript dialog unescaped.** Double quotes are legal  in macOS filenames, and one breaks out of the string literal. With a file named   `my"movie bad.ts`, the generated script is:

  ```applescript
  display dialog "Finished processing 1 file(s).
  Failed:
  my"movie bad.ts (exit code 4)
  Run from the Terminal to see detailed error output." buttons {"OK"} ...
  ```
  That's a syntax error, so osascript fails and the user gets no dialog at all after processing — which is the same class of silent failure this PR set   out to fix. It's also an execution vector, since AppleScript has  do shell script; low severity for a local droplet, but worth closing. Only the  failure and no-captions paths are affected — the all-success path uses a fixed  string.

  Something like this is enough:

 ```
  escape_as() { printf '%s' "$1" | sed 's/\\/\\\\/g; s/"/\\"/g'; }
  display "$(escape_as "$message")"
...

I think that's all needed

kaihere14:
@cfsmp3 

Good catch, thanks. Fixed. Root cause was exactly as you described: the message was interpolated into the AppleScript string literal unescaped, so a stray " in a filename terminated the string early and made osascript choke, producing the same silent-failure symptom this PR was meant to fix.

Fix: added an escape_as() helper that backslash-escapes \ and " before the message goes into the dialog string, applied inside display() so every call site is covered, not just the failure/no-captions paths:

```
function escape_as() {
  printf '%s' "$1" | sed 's/\\/\\\\/g; s/"/\\"/g'
}

function display() {
  local escaped
  escaped="$(escape_as "$1")"
  osascript <<EOT
    ...
      display dialog "$escaped" buttons {"OK"} ...
    ...
EOT
}
```

Since none of this is build-generated, I also had to sync the fix into the checked-in binary copies that actually ship: CCExtractor.app/Contents/Resources/script, and repackaged both CCExtractor.dmg copies (mac/gui/src/ and the one embedded in InstallCCExtractorMacGUI.app) from the updated app bundle. Otherwise the source fix alone wouldn't reach users.

Tested by dropping a file literally named te"st.png onto the app. Dialog now renders correctly instead of erroring out silently. Screenshot below:

Screenshot 2026-07-27 at 12 56 15 AM

PR #2316, opened 2026-08-11
Fix SSA compliance issues #2310

### LINKED ISSUE #2310: Substation Alpha (SSA/ASS) output is not fully compliant with format specification
Substation Alpha output produced by CCExtractor is correctly interpreted by most playback software, but is not _technically_ compliant with the format specification. Here is example output produced from [my previously uploaded DresdenTest.ts](https://1drv.ms/v/c/8b5ee05414946e03/IQDLLzwLvNY3RJi8rYvDN7PyAVTV7dC2YotrGHTGfnRkFUM?e=4YsJPQ):

```
Dialogue: 0,00:00:08.90,00:00:12.07,Default,,0000,0000,0000,,{\an7\pos(67,228)}   - WHAT IS IT?                \N   - THIS WAS YOUR MOTHER'S.    
Dialogue: 0,00:00:12.7,00:00:14.29,Default,,0000,0000,0000,,{\an7\pos(154,244)}           <i> GO AHEAD, OPEN IT.</i>  
```

The first issue is with the timing format. Substation Alpha timing format uses a fixed number of digits for all timings: two for all units **except the hours**, which are limited to one digit. All of the example lines use two digits for the hours rather than one. Additionally, the second line does not add the leading zero to the decimal for the start time, causing it to incorrectly be presented as starting at 12.70 seconds rather than 12.07 seconds. The correct format for the timings would be as follows:

```
Dialogue: 0,0:00:08.90,0:00:12.07,Default,,0000,0000,0000,,{\an7\pos(67,228)}   - WHAT IS IT?                \N   - THIS WAS YOUR MOTHER'S.    
Dialogue: 0,0:00:12.07,0:00:14.29,Default,,0000,0000,0000,,{\an7\pos(154,244)}           <i> GO AHEAD, OPEN IT.</i>  
```

Second, Substation Alpha does not use html tags for format changes, but rather its own formatting tags which can be applied mid-text like html, but unlike html **do not need to be followed by a corresponding close tag if the format persists to the end of the subtitle** since line formatting is reset to default with each new subtitle. Again, here is the current output produced from DresdenTest.ts:

```
Dialogue: 0,00:00:01.45,00:00:03.40,Default,,0000,0000,0000,,{\an7\pos(125,244)}        <i> ( opens chest )</i>        
Dialogue: 0,00:00:08.90,00:00:12.07,Default,,0000,0000,0000,,{\an7\pos(67,228)}   - WHAT IS IT?                \N   - THIS WAS YOUR MOTHER'S.    
Dialogue: 0,00:00:12.7,00:00:14.29,Default,,0000,0000,0000,,{\an7\pos(154,244)}           <i> GO AHEAD, OPEN IT.</i>  
Dialogue: 0,00:00:23.92,00:00:25.42,Default,,0000,0000,0000,,{\an7\pos(106,244)}      <i> THAT'S A SHIELD BRACELET.</i>
Dialogue: 0,00:00:25.42,00:00:26.40,Default,,0000,0000,0000,,{\an7\pos(163,228)}            <i> I NEVER REALLY KNEW</i>\N            <i> HOW IT WORKED,</i>     
```

To match Substation Alpha specification (including the timing issue mentioned above) the output should look as follows:

```
Dialogue: 0,0:00:01.45,0:00:03.40,Default,,0000,0000,0000,,{\an7\pos(125,244)}        {\i1} ( opens chest )        
Dialogue: 0,0:00:08.90,0:00:12.07,Default,,0000,0000,0000,,{\an7\pos(67,228)}   - WHAT IS IT?                \N   - THIS WAS YOUR MOTHER'S.    
Dialogue: 0,0:00:12.07,0:00:14.29,Default,,0000,0000,0000,,{\an7\pos(154,244)}           {\i1} GO AHEAD, OPEN IT.  
Dialogue: 0,0:00:23.92,0:00:25.42,Default,,0000,0000,0000,,{\an7\pos(106,244)}      {\i1} THAT'S A SHIELD BRACELET.
Dialogue: 0,0:00:25.42,0:00:26.40,Default,,0000,0000,0000,,{\an7\pos(163,228)}            {\i1} I NEVER REALLY KNEW{\i0}\N            {\i1} HOW IT WORKED,     
```

alaotach:

Reason for this PR:

Sanity check:

Repro instructions:

1. Download the `DresdenTest.ts` sample provided in Issue #2310: [DresdenTest.ts](https://1drv.ms/v/c/8b5ee05414946e03/IQDLLzwLvNY3RJi8rYvDN7PyAVTV7dC2YotrGHTGfnRkFUM?e=4YsJPQ)
2. Run CCExtractor to generate Substation Alpha output:
   `ccextractor DresdenTest.ts -out=ass`
3. Inspect the resulting `.ass` file.
   - **Before this PR:** The file contains two-digit hours (`00:`), misaligned milliseconds (`.7` instead of `.07`), and raw HTML tags (`<i>`).
   - **After this PR:** The file contains compliant single-digit hours (`0:`), correctly padded milliseconds (`.07`), and ASS override tags (`{\i1}`).

---

### Description
This PR resolves several compliance issues with the Substation Alpha (SSA/ASS) output generated by CCExtractor, addressing the bugs reported in #2310.

**Fixes included:**
1. **Timestamp Formatting**
   - Changed the hour format from zero-padded two digits (`%02u`) to a single unpadded digit (`%u`) in all `Dialogue:` lines to comply with strict SSA parser expectations.
   - Fixed a critical padding bug where centiseconds/milliseconds were incorrectly formatted (missing leading zeros), which caused timestamp values to shift silently (e.g., `12.07s` rendering as `12.7s`).

2. **CEA-608 Override Tag Generation**
   - Modified `get_decoder_line_encoded()` so that when outputting SSA, it emits valid SSA override tags directly (e.g., `{\i1}`, `{\u1}`, `{\c&H00FF00&}`) instead of generating HTML tags (`<i>`, `<u>`, `<font...>`) that are technically invalid in SSA. 
   - Hoisted the SSA color mapping table (`ssa_cols`) to file-scope in `ccx_encoders_helpers.c` to ensure the emitted font colors stay safely synchronized with the tag-stack tracking logic.

3. **HTML-to-SSA Tag Translation (DVB/Teletext/708)**
   - Implemented a table-driven tag replacement inside `write_stringz_as_ssa()` for subtitle formats that generate generic HTML-tagged C-strings upstream. This gracefully translates `<i>`, `<u>`, `<b>`, and `<font color="#RRGGBB">` into their respective SSA equivalents (`{\i1}`, `{\u1}`, `{\b1}`, `{\c&HBBGGRR&}`) without breaking other format outputs that rely on the HTML representation.

Fixes #2310.

### DIFF
--- a/src/lib_ccx/ccx_encoders_helpers.c
+++ b/src/lib_ccx/ccx_encoders_helpers.c
@@ -263,15 +263,24 @@ unsigned char *close_tag(struct encoder_ctx *ctx, unsigned char *buffer, char *t
 		switch (cur)
 		{
 			case 'F':
-				buffer += encode_line(ctx, buffer, (unsigned char *)"</font>");
+				if (ctx->write_format == CCX_OF_SSA)
+					buffer += encode_line(ctx, buffer, (unsigned char *)"{\\c}");
+				else
+					buffer += encode_line(ctx, buffer, (unsigned char *)"</font>");
 				(*pchanged_font)--;
 				break;
 			case 'U':
-				buffer += encode_line(ctx, buffer, (unsigned char *)"</u>");
+				if (ctx->write_format == CCX_OF_SSA)
+					buffer += encode_line(ctx, buffer, (unsigned char *)"{\\u0}");
+				else
+					buffer += encode_line(ctx, buffer, (unsigned char *)"</u>");
 				(*punderlined)--;
 				break;
 			case 'I':
-				buffer += encode_line(ctx, buffer, (unsigned char *)"</i>");
+				if (ctx->write_format == CCX_OF_SSA)
+					buffer += encode_line(ctx, buffer, (unsigned char *)"{\\i0}");
+				else
+					buffer += encode_line(ctx, buffer, (unsigned char *)"</i>");
 				(*pitalics)--;
 				break;
 		}
@@ -310,7 +319,21 @@ unsigned get_decoder_line_encoded(struct encoder_ctx *ctx, unsigned char *buffer
 
 			// Add new font tag
 			if (MAX_COLOR > its_color)
-				buffer += encode_line(ctx, buffer, (unsigned char *)color_text[its_color][1]);
+			{
+				if (ctx->write_format == CCX_OF_SSA)
+				{
+					const char *ssa_cols[] = {
+						"", "{\\c&H00FF00&}", "{\\c&HFF0000&}", "{\\c&HFFFF00&}", 
+						"{\\c&H0000FF&}", "{\\c&H00FFFF&}", "{\\c&HFF00FF&}", 
+						"{\\c&H", "", ""
+					};
+					buffer += encode_line(ctx, buffer, (unsigned char *)ssa_cols[its_color]);
+				}
+				else
+				{
+					buffer += encode_line(ctx, buffer, (unsigned char *)color_text[its_color][1]);
+				}
+			}
 			else
 			{
 				ccx_common_logging.log_ftn("WARNING:get_decoder_line_encoded:Invalid Color index Selected %d\n", its_color);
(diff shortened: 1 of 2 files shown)


cfsmp3 (changes requested):
The timestamp fixes are right and worth having: `.%01u` was printing `12.07s` as `12.7`, and single-digit hours match the spec.

Blocking issue: the output can contain uninitialized heap memory. Run with `--out=ass --startcreditstext 'Alpha\nBravo\nCharlie'`:

```
master : Alpha\NBravo\NCharlie\NrmatÁ\N
this PR: Alpha\NBravo\NCharlie\Ned end of bitstream.
```

`write_stringz_as_ssa` reads the segments back with `while (begin < unescaped + len)`, bounded by the *input* length even though the loop wrote `pos_w` bytes. `\n` consumes two characters and writes one, so `pos_w < len` and the read-back runs past the last NUL into freshly `malloc`'d memory. It is pre-existing on master, but this PR changes when it triggers rather than fixing it — tag expansion can now push `pos_w` above `len` as well, which truncates instead. Bounding by `pos_w` fixes both directions:

```c
unsigned char *end = unescaped + pos_w;
while (begin < end)
```

Two smaller points:

`write_stringz_as_ssa` is only reachable from `--startcreditstext` / `--endcreditstext` — it has no other callers. So the HTML→ASS table added there does not affect normal caption output; that comes from `get_decoder_line_encoded`, which the other half of this PR handles. Worth stating in the description so the next reader does not go looking.

The `unescaped[pos_w] = 0; ... pos_w++` → `unescaped[pos_w++] = 0; continue;` change is equivalent to master (both branches previously fell through to a shared `pos_w++`), so it is a readability change rather than a fix.

I swept 30 samples with `--out=ass`: Dialogue and `\N` counts identical to master, so the normal caption path looks unaffected.

PR #2321, opened 2026-08-12
fix: correctly initialize frame type for tx3g and mov_text subtitles

shubhrai23:
This PR fixes MP4 subtitle timestamp initialization by ensuring subtitle-only tracks (including \mov_text\/\	x3g\) get an I-frame picture type so \set_fts()\ can set \min_pts\ from the first sample.


cfsmp3:
How can I reproduce the problem? Do you have a sample video? @shubhrai23

PR #2324, opened 2026-08-13
[FIX] transcript: suppress spurious blank lines from write_cc_line_as_transcript2

x15sr71:

Reason for this PR:

Sanity check:

Repro instructions:

Run with `--autoprogram --out=ttxt --latin1` on any sample with a pop-on → roll-up mode transition. The output contains one or more spurious blank lines not present in the reference file. Confirmed on the sample platform via [RT41](https://sampleplatform.ccextractor.org/regression/test/41/view), RT84, RT11, RT29, RT30.

---
## Issue:
`write_cc_buffer_as_transcript2` emits spurious blank lines in `--out=ttxt output` on samples with pop-on → roll-up transitions.

## Root cause:
Introduced in PR #2105 - it refactored `write_cc_buffer_as_transcript2` to emit `encoded_end_frame` after each block, and changed the call to:
```
write_cc_line_as_transcript2(data, context, i);
wrote_something = 1;   // unconditional
```
Because `write_cc_line_as_transcript2` returned void, the caller couldn't know it wrote nothing (blank cursor row). `wrote_something` got set to 1 anyway, so a bare newline was emitted for every empty row.

## Fix:
Change `write_cc_line_as_transcript2` from void to int (returns 1 if content was written, 0 if not) and use `|=` in the caller:
`wrote_something |= write_cc_line_as_transcript2(data, context, i)`; `encoded_end_frame` is now only written when at least one line produced output.

## Testing: 
- [RT41](https://sampleplatform.ccextractor.org/regression/test/41/view) (`01509...ts`, `--out=ttxt --latin1`): two spurious blank lines removed. Output is byte-identical to accepted variant ref `9f139b8c...`

### DIFF
--- a/src/lib_ccx/ccx_encoders_common.h
+++ b/src/lib_ccx/ccx_encoders_common.h
@@ -238,7 +238,7 @@ void write_cc_buffer_to_gui(struct eia608_screen *data, struct encoder_ctx *cont
 int write_cc_buffer_as_g608(struct eia608_screen *data, struct encoder_ctx *context);
 int write_cc_buffer_as_transcript2(struct eia608_screen *data, struct encoder_ctx *context);
 
-void write_cc_line_as_transcript2(struct eia608_screen *data, struct encoder_ctx *context, int line_number);
+int write_cc_line_as_transcript2(struct eia608_screen *data, struct encoder_ctx *context, int line_number);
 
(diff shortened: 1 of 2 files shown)


x15sr71:
The `Docker build (hardsubx)` failure is unrelated to this PR — it's a transient Docker Hub auth outage (`500 Internal Server Error` on `auth.docker.io/token`). The other two Docker jobs passed in the same run. A re-run should clear it.

PR #1736, opened 2025-08-23
[Rust]Ported ES Module to Rust

steel-bucket:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [x] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor.

---
This PR ports the entire `es_functions.c` and `es_userdata.c` files to Rust.
Tested on [This sample](https://sampleplatform.ccextractor.org/sample/b22260d065ab537899baaf34e78a5184671f4bcb2df0414d05e6345adfd7812f)

PR #1662, opened 2025-02-11
[FEAT] added `demuxer` and `file_functions` module

steel-bucket:
**In raising this pull request, I confirm the following (please check boxes):**

- [x] I have read and understood the [contributors guide](https://github.com/CCExtractor/ccextractor/blob/master/.github/CONTRIBUTING.md).
- [x] I have checked that another pull request for this purpose does not exist.
- [x] I have considered, and confirmed that this submission will be valuable to others.
- [x] I accept that this submission may not be used, and the pull request closed at the will of the maintainer.
- [x] I give this submission freely, and claim no ownership to its content.
- [x] **I have mentioned this change in the [changelog](https://github.com/CCExtractor/ccextractor/blob/master/docs/CHANGES.TXT).**

**My familiarity with the project is as follows (check one):**

- [ ] I have never used CCExtractor.
- [ ] I have used CCExtractor just a couple of times.
- [ ] I absolutely love CCExtractor, but have not contributed previously.
- [x] I am an active contributor to CCExtractor(2 previous contributions).

---

In this PR, I have attempted to port the large `demuxer` module to Rust, the primary logic of the heavily interconnected C libraries `file_functions.c` and, `ccx_demuxer.c` and their corresponding header files has aleady been implemented here. 
This PR was inspired by the ones done for the 708 Decoder in CCextractor. 
The part of the codebase that the demuxer part of this PR migrates to Rust is the part that Opens a File(ccx_demuxer_open), points the codebase towards that file, detects the stream type and some other parameters like myth, and then closes the file or gets the file size.
The file_functions part of this PR is tested locally, and in unit tests, but integrating it into C made the codebase really slow, due to the constant copying back and forth C and Rust, so it was left to be used in future Rust Libraries like MythTV, MXF, GXF, etc.
Any criticism or suggestion is wholeheartedly welcome.
- It is built and ready to be reviewed now.

PR #2327, opened 2026-08-25
[CLEANUP]: Remove redundant es_functions.c C code superseded by the Rust ES port

x15sr71:

Reason for this PR:

- [ ] This PR fixes a bug that I have personally experienced or that a real user has reported and for which a sample exists.

Sanity check:
---
**Note — this is a verification, not a bug repro** (this PR removes dead code, there's no bug to reproduce): it verifies the deletion is behavior-preserving by running the sample through the exact functions this PR touches (`process_m2v → es_video_sequence → delegates`) and confirming baseline vs branch output is byte-identical.

### Verification

**Sample:** (the filename [725a49f871dc5a2](https://sampleplatform.ccextractor.org/sample/725a49f871dc5a2ebe9094cf9f838095aae86126e9629f96ca6f31eb0f4ba968) ~19 MB MPEG-2 program stream, NTSC line-21 EIA-608). Such streams route `CCX_PES → process_m2v → es_video_sequence → delegates → Rust` (`general_loop.c:892`), exercising exactly the functions this PR touches.

Deadness itself is proven statically (every deleted statement sits after an unconditional `return ccxr_…()`, unreachable by C semantics). The steps below are corroboration: they confirm the sample actually drives the touched path, and that deleting the unreachable code is behavior-preserving.

```sh
# Build master (baseline) and this branch
git worktree add /tmp/ccx-base master && (cd /tmp/ccx-base/linux && cmake ../src && make -j4)
(cd linux && cmake ../src && make -j4)

# 1. Confirm the sample drives the touched path (expect thousands of hits)
./linux/ccextractor --debug sample.mpg -o /dev/null 2>&1 | grep -c 'es_video_sequence'

# 2. Diff output — expect byte-identical across every format
for f in srt ttxt sami; do
  /tmp/ccx-base/linux/ccextractor --out=$f sample.mpg -o base.$f
  ./linux/ccextractor             --out=$f sample.mpg -o pr.$f
  diff base.$f pr.$f && echo "$f: identical"
done
```

**Expected:** step 1 prints a large count (8,722 in my run — the ES path is heavily used); step 2 prints `srt: identical`, `ttxt: identical`, `sami: identical`.

---

### What this removes

1. **The dead bodies of the 7 ES entry points.** Each begins with an unconditional `return ccxr_…()` that forwards to the Rust implementation; every statement after it is unreachable. This PR reduces each to that single delegating line.
2. **The 6 now-orphaned C parser helpers**, whose *only* callers were those dead bodies:
   `sequence_header`, `sequence_ext`, `gop_header`, `pic_header`, `pic_coding_ext`, `extension_and_user_data` — plus their forward declarations.

So the deleted code is unreachable in two forms: each of the 7 delegate bodies sits after an unconditional `return ccxr_…()`, and the 6 helpers are called **only** from those dead bodies.

### What is kept (all live, untouched)

- `process_m2v` — the entry point (real C body: bitstream init + dispatch).
- `es_video_sequence` — the C state-machine dispatcher reached from `process_m2v`. It calls **only** the delegates, never the helpers.
- The 7 one-line delegates, which continue to forward to Rust via FFI.

---

### What made this code redundant

The MPEG-2 elementary-stream parser was ported to Rust across two merged PRs:

- **#1662** — Added the `extern` declarations for the ES `ccxr_*` functions and converted the 7 C entry points into thin delegates by prepending `return ccxr_…()`. **This is what rendered the original C parser bodies unreachable.**
- **#1736** — Added the actual Rust ES parsing implementation (`src/rust/src/es/`) that those delegates now call.

The old C parser bodies and helpers were left in place as unreachable reference code. The compiler can't flag them (`-Wunused-function`) because the helpers are still statically referenced *from* the dead bodies — which is why this survived the earlier cleanup pass (#1738).

### What runs in its place today

The live runtime path is unchanged by this PR:

```
process_m2v (C)  →  es_video_sequence (C dispatcher)  →  7 thin C delegates  →  ccxr_* (Rust)
```

The delegates forward via FFI to the Rust ES module at `src/rust/src/es/mod.rs`. The deleted C helpers correspond directly to Rust functions that already do the work:

| Deleted C helper(s) | Lived inside (dead body of) | Now handled in Rust by |
|---|---|---|
| `sequence_header`, `sequence_ext` | `read_seq_info` | `ccxr_read_seq_info` (`es/mod.rs:37`, `es/seq.rs`) |
| `gop_header` | `read_gop_info` | `ccxr_read_gop_info` (`es/mod.rs:53`, `es/gop.rs`) |
| `pic_header`, `pic_coding_ext` | `read_pic_info` | `ccxr_read_pic_info` (`es/mod.rs:67`, `es/pic.rs`) |
| `extension_and_user_data` | `read_eau_info` | `ccxr_read_eau_info` (`es/mod.rs:81`) |
| (start-code scanning) | `next_start_code` / `search_start_code` | `ccxr_next_start_code` (`es/mod.rs:19`) / `ccxr_search_start_code` (`es/mod.rs:28`) |
| (slice scanning) | `read_pic_data` | `ccxr_read_pic_data` (`es/mod.rs:98`) |

The C→Rust *behavioral* equivalence for these functions was established and tested by #1662/#1736 when the port landed; this PR neither relies on nor changes it, since the deleted C wasn't executing in either case.

---

### Why the removed code is never run, in any condition

**1. Static / language-level (input-independent certainty).**
- Every helper call site sits strictly *after* an unconditional `return ccxr_…()` — unreachable by C semantics. No input, CLI flag, mode, or build configuration can reach it.
- No conditional compilation gates those returns. The only preprocessor directive in the file is a permanently-disabled `#if 0` block *inside* a removed helper body.
- No indirect/function-pointer use: grep for `&sequence_header` etc. across the tree — zero address-of any helper. All calls were direct, and all were in the dead bodies.
- No references anywhere else: whole-tree grep across **C (`.c`/`.h`) and Rust (`src/rust/`)** finds zero live references to the 6 helpers. The only mention outside the deleted code is one *comment* in `general_loop.c`.
- Post-deletion file has **zero** dangling references and compiles with **zero** warnings (`es_functions.c.o`).

**2. Empirical (executed the exact path we touched).**
Built a baseline binary (master) and a branch binary (this PR), and ran both on an MPEG-2 program-stream sample carrying NTSC line-21 (EIA-608) captions — which route `CCX_PES → process_m2v → es_video_sequence → delegates → Rust` (`general_loop.c:892`). Confirmed via `--debug` that this sample drives `es_video_sequence` **8,722 times**.

| Check | Result |
|---|---|
| Full `--debug` execution trace | **byte-identical** (matching md5) |
| `--out=srt` | **byte-identical** |
| `--out=ttxt` | **byte-identical** |
| `--out=sami` | **byte-identical** |

Identical execution traces prove the runtime path through the delegates→Rust is unchanged; identical output confirms no functional regression.


x15sr71:
These test failures aren't from this PR (a pure deletion in `es_functions.c`).

The one test the **windows** comment singles out under "pass there, fail here" (`--autoprogram --out=srt --latin1`) actually fails on **linux too** — linux just doesn't list it. Each platform's comment diffs against its own tip-of-master run: **windows → test 9495**, **linux → test 9499** (both commit `3af3fc2`). Those two runs recorded that test differently — **PASS in 9495, FAIL in 9499** — under the old self-approving baseline, fixed since in https://github.com/CCExtractor/sample-platform/pull/1175 and https://github.com/CCExtractor/sample-platform/pull/1178. So it reads as pass→fail on windows (listed) and fail→fail on linux (not listed): identical failure, different frozen baselines.

This is only a leftover of those two pre-fix baseline runs — it should clear once this PR is re-tested against a newer master run (recorded after the fix), where both platforms record the test as failing consistently.

PR #2329, opened 2026-09-01
[FIX] chore: remove two dead includes flagged by clangd

### LINKED ISSUE #2328: [CLEANUP] Two dead includes in ccx_decoders_common.c and general_loop.c
### Summary

clangd's unused-includes check flagged three headers as not directly used. I investigated each one properly rather than trusting the tool blindly, since one of the three turned out to be a false positive.

### 1. `ccx_common_char_encoding.h` in `src/lib_ccx/ccx_decoders_common.c:6` — dead, safe to remove

This header declares five functions (`get_char_in_latin_1`, `get_char_in_unicode`, `get_char_in_utf_8`, `cctolower`, `cctoupper`). None of them are referenced anywhere in this 736-line file.

History: added in August 2014 (`301c2a71`, "Library - step one and half") to support `get_decoder_line_basic()` and `find_limit_characters()`, which lived in this file at the time. Both functions have since moved elsewhere (`find_limit_characters` is now in `ccx_encoders_common.c:103`; `get_decoder_line_basic` no longer exists in the decoders). The include was left behind after that migration and has been dead weight since.

### 2. `ffmpeg_intgr.h` in `src/lib_ccx/general_loop.c:17` — NOT dead, false positive

`ffmpeg_get_more_data` (declared in this header) is genuinely used at `general_loop.c:1418`, but inside `#ifdef ENABLE_FFMPEG`. clangd's indexer doesn't appear to have that flag set, so it can't see the usage and incorrectly flags the include as unused. In an FFmpeg-enabled build this include is required. Flagging this here for visibility, no change needed.

### 3. `ccx_demuxer_mxf.h` in `src/lib_ccx/general_loop.c:20` — dead, safe to remove

This header declares `ccx_probe_mxf`, `ccx_mxf_init`, and a few MXF-specific types, none of which are used in this file. `general_loop.c` does handle MXF (`case CCX_SM_MXF`), but via `ccx_mxf_getmoredata`, which is declared in `lib_ccx.h` — already included directly in this file. This include appears to have been added by reflex alongside the MXF feature in November 2017 (`42ab1640`, "Add MXF support") without ever actually being needed.

### Proposed fix

Remove the two dead includes (#1 and #3). Leave `ffmpeg_intgr.h` untouched, it's a genuine dependency under `ENABLE_FFMPEG`, just invisible to clangd's default indexing.

I'd like to be assigned and submit a PR for this.

kaihere14:
ccx_common_char_encoding.h in ccx_decoders_common.c is unused — the functions it declares moved to ccx_encoders_common.c after a 2014 refactor and the include was never cleaned up.

ccx_demuxer_mxf.h in general_loop.c is also unused — general_loop.c handles MXF via ccx_mxf_getmoredata, which is already declared in lib_ccx.h (included separately). This header appears to have been added by reflex alongside MXF support in 2017 without ever being needed.

A third flagged include, ffmpeg_intgr.h in general_loop.c, was investigated and left untouched — it's a genuine dependency under `#ifdef ENABLE_FFMPEG` (used at general_loop.c:1418), invisible to clangd's default indexing since that flag isn't set during static analysis.

Closes #2328

Reason for this PR:

Sanity check:

## Repro instructions

This isn't a crash-reproducing bug fix, it's a dead-code cleanup, so "reproduction" here means verifying the includes are genuinely unused and that removing them changes nothing functionally.

**To reproduce the original finding:**
1. Run clangd's unused-includes check (or open `src/lib_ccx/ccx_decoders_common.c` and `src/lib_ccx/general_loop.c` in an IDE with clangd enabled)
2. It flags `ccx_common_char_encoding.h` (ccx_decoders_common.c:6) and `ccx_demuxer_mxf.h` (general_loop.c:20) as included but not directly used

**To verify each finding before this PR:**
- `grep -n "get_char_in_latin_1\|get_char_in_unicode\|get_char_in_utf_8\|cctolower\|cctoupper" src/lib_ccx/ccx_decoders_common.c` returns nothing
- `grep -n "ccx_probe_mxf\|ccx_mxf_init\|MXFTrack\|MXFContext" src/lib_ccx/general_loop.c` returns nothing
- `git log -p --follow src/lib_ccx/ccx_decoders_common.c` around commit `301c2a71` shows the original functions that needed this header, and shows they've since moved to `ccx_encoders_common.c`
- `git log -p --follow src/lib_ccx/general_loop.c` around commit `42ab1640` ("Add MXF support") shows `ccx_demuxer_mxf.h` was added alongside the `CCX_SM_MXF` case, but the actual function used (`ccx_mxf_getmoredata`) is declared in `lib_ccx.h`, already included separately

**To verify this PR doesn't break anything:**
1. Build before this change: `cmake --build build` (or the standard project build)
2. Build after this change with the two includes removed
3. Both builds should succeed identically — no compile errors, since nothing in either file references anything from the removed headers

No functional or behavioral change; only the two confirmed-dead includes are removed.

### DIFF
--- a/src/lib_ccx/ccx_decoders_common.c
+++ b/src/lib_ccx/ccx_decoders_common.c
@@ -3,7 +3,6 @@ made to reuse, not duplicate, as many functions as possible */
 
 #include "ccx_decoders_common.h"
 #include "ccx_common_structs.h"
-#include "ccx_common_char_encoding.h"
 #include "ccx_common_constants.h"
 #include "ccx_common_timing.h"
 #include "ccx_common_common.h"
--- a/src/lib_ccx/general_loop.c
+++ b/src/lib_ccx/general_loop.c
@@ -17,7 +17,6 @@
 #include "ffmpeg_intgr.h"
 #include "ccx_gxf.h"
 #include "dvd_subtitle_decoder.h"
-#include "ccx_demuxer_mxf.h"
 #include "ccx_dtvcc.h"
 
 int end_of_file = 0; // End of file?


kaihere14:
Hey @cfsmp3, small cleanup PR whenever you have a sec. Two dead includes, verified, no functional change

Issue #2305, opened 2026-08-05
sampleplatform CI ("CI - linux" / "CI - windows" checks) is broken across all open PRs — stale baseline + infra failures


CapThunder19:
### Problem
The CI - linux and CI - windows checks (powered by sampleplatform.ccextractor.org) are failing on essentially every open PR right now, regardless of what the PR actually changes.

### Evidence
Checked 6 currently-open PRs (#2294, #2298, #2301, #2302, #2304, and one more) — none touching related code to each other  and all show CI - linux and/or CI - windows as failing:

| PR | CI - linux | CI - windows |
|----|-----------|--------------|
| #2301 (EIA-608 encoding) | fail  "Not all tests completed successfully" | fail  same |
| #2302 (libpng ARM/CMake) | fail  same | fail  same |
| #2298 (EIA-608 CC3/CC4) | fail  *"VM creation failed (ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS). Please contact the administrator."* | fail  "Not all tests completed successfully" |
| #2294 (Rust network framing) | fail  *"Could not complete test, time limit exceeded"* | pending |
| #2304 (DVB subtitle fix, mine) | fail  same as #2301/#2302 | fail/pending |

Three distinct failure modes visible:
1. *Stale comparison baseline: the "compared to test for commit ..." reference used across multiple different PR runs is 2feb09a (dated **2026-04-24), while current master HEAD is ba4a8390 (dated **2026-07-26)  a **3-month gap*. Sample outputs/expected results have likely drifted in that window, producing failures unrelated to any given PR's actual diff (e.g. failures in Broken, CEA-708, General, Options, Teletext, WTV, XDS categories  none of which relate to PR #2304's DVB-subtitle-only change).
2. *Cloud VM provisioning failures*: ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS (PR #2298)  the underlying cloud runner pool appears to be exhausted.
3. *Timeouts*: "time limit exceeded" (PR #2294).

### Suggested fix direction
- Update/automate the baseline comparison commit so it tracks current master instead of a fixed, aging reference.
- Investigate the sampleplatform cloud runner pool exhaustion / timeout issues (may be capacity/quota related).

PR #2331, opened 2026-09-01
[FIX] NUL-terminate mode/info strings in add_cc_sub_text

### LINKED ISSUE #2330: [BUG] add_cc_sub_text does not NUL-terminate 4-character mode strings (ISDB/BURN)
CCExtractor version: master (0.96)

# Necessary information

- Is this a regression (i.e. did it work before)? NO (as far as I can tell this has been present in `add_cc_sub_text` for a while)
- What platform did you use? Windows (the bug is in portable C, not Windows-specific)
- What were the used arguments? Not a single-file extraction failure. This is a C string bug in a helper used by ISDB and HardSubx when those subtitles are later treated as C strings (for example transcript output with `--transcript` / showMode).

# Video links

No sample file: the defect is visible from the source. Callers pass 4-character string literals into a `strncpy` of length 4, which does not write a terminator. I can add a sample later if maintainers want one for ISDB/HardSubx transcript output.

# Additional information

## Problem

`add_cc_sub_text()` in `src/lib_ccx/ccx_common_common.c` copies `info` and `mode` with `strncpy(..., 4)`.

`struct cc_subtitle` stores:

- `char mode[5]` — 4 characters plus a required `'\0'`
- `char info[4]` — 3 characters plus a required `'\0'`

`strncpy(dest, src, n)` does **not** add a terminating `'\0'` when `src` is `n` characters long. It only pads with zeros when `src` is *shorter* than `n`.

## Why this matters

Later code treats `sub->mode` as a C string:

- `strcmp(sub->mode, "TLT")` in `src/lib_ccx/ccx_encoders_transcript.c`
- `fdprintf(..., "%s|", sub->mode)` in the same file

If there is no terminator, `strcmp` / `"%s"` read past the array (undefined behavior: garbage in transcript output, or a crash).

## Call sites that pass a 4-character `mode`

These are exactly 4 characters, so `strncpy(..., 4)` copies all four letters and never writes `'\0'`:

- `"ISDB"` from `src/lib_ccx/ccx_decoders_isdb.c`
- `"BURN"` from `src/lib_ccx/hardsubx_decoder.c`

`"TLT"` is only 3 characters, so that path is terminated by `strncpy` itself.

`mode[4]` can still look correct on the **first** subtitle if the struct was created with `calloc` (zeros). The **second** node is allocated with `malloc` (not zeroed), so `mode[4]` is leftover garbage. That makes the bug easy to miss on a short run.

## Proposed direction (not implemented in this issue)

Copy with `sizeof(field) - 1` and always write the last byte as `'\0'`, so `mode` and `info` are always valid C strings.

I would like to send a small PR for this if the diagnosis looks right.

NaitikVerma6776:
Fixes #2330

Reason for this PR:

- [x] This PR adds new functionality.
- [x] This PR fixes a bug that I have personally experienced or that a real user has reported and for which a sample exists. (Reported in #2330; this is a source-level C string bug, not a video-sample extraction failure.)

Sanity check:

Repro instructions:

This is a C string bug in `add_cc_sub_text()`, not a single-file caption extraction failure. No video sample is attached because the defect is in how `mode` / `info` are copied.

1. `struct cc_subtitle` has `char mode[5]` and `char info[4]`.
2. Callers pass 4-character modes: `"ISDB"` (`ccx_decoders_isdb.c`) and `"BURN"` (`hardsubx_decoder.c`).
3. Old code used `strncpy(..., 4)`. When the source is exactly 4 characters, `strncpy` does not write `'\0'`.
4. Transcript output then uses `strcmp(sub->mode, "TLT")` and `"%s"` on `sub->mode` (`ccx_encoders_transcript.c`), which requires a terminator.
5. `"TLT"` (3 characters) was already terminated by `strncpy`. The first subtitle can also look fine if the struct was `calloc`'d (the extra byte is already 0). Later nodes use `malloc` and can leave `mode[4]` as garbage.

How to confirm from source: search for `add_cc_sub_text(..., "ISDB"` / `"BURN"` and the old `strncpy(..., 4)` on master vs this branch.

---

## Summary

`add_cc_sub_text()` now copies `info` and `mode` with `sizeof(field) - 1` and always writes a terminating `'\0'`.

That keeps 4-character labels such as `ISDB` and `BURN` valid C strings, so later `strcmp` / `"%s"` in transcript output cannot read past the array.

No changelog update: this is a bug fix only. No new feature code.

### DIFF
--- a/src/lib_ccx/ccx_common_common.c
+++ b/src/lib_ccx/ccx_common_common.c
@@ -69,9 +69,15 @@ int add_cc_sub_text(struct cc_subtitle *sub, char *str, LLONG start_time,
 	sub->start_time = start_time;
 	sub->end_time = end_time;
 	if (info)
-		strncpy(sub->info, info, 4);
+	{
+		strncpy(sub->info, info, sizeof(sub->info) - 1);
+		sub->info[sizeof(sub->info) - 1] = '\0';
+	}
 	if (mode)
-		strncpy(sub->mode, mode, 4);
+	{
+		strncpy(sub->mode, mode, sizeof(sub->mode) - 1);
+		sub->mode[sizeof(sub->mode) - 1] = '\0';
+	}
 	sub->got_output = 1;
 	sub->next = NULL;
 
--- a/src/rust/lib_ccxr/src/encoder/txt_helpers.rs
+++ b/src/rust/lib_ccxr/src/encoder/txt_helpers.rs
@@ -42,21 +42,19 @@ fn change_utf8_encoding(dest: &mut [u8], src: &[u8], len: i32, out_enc: Encoding
 
     while src_idx < max {
         let c = src[src_idx];
-        let c_len: usize;
-
-        if c < 0x80 {
-            c_len = 1;
+        let c_len: usize = if c < 0x80 {
+            1
         } else if (c & 0x20) == 0 {
-            c_len = 2;
+            2
         } else if (c & 0x10) == 0 {
-            c_len = 3;
+            3
         } else if (c & 0x08) == 0 {
-            c_len = 4;
+            4
         } else if (c & 0x04) == 0 {
-            c_len = 5;
+            5
         } else {
-            c_len = 1;
-        }
+            1
+        };
 
         match out_enc {
             Encoding::UTF8 => {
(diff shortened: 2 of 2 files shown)


NaitikVerma6776:
Greetings maintainers 
The 70 failing tests are pre-existing and fail identically on master (see #2305) — this PR doesn't introduce any new test failures.

cfsmp3 (changes requested):
@NaitikVerma6776 The C bit looks good, but the Rust part is unrelated. Can just remove it and leave the C issue fix only?

NaitikVerma6776:
hey @cfsmp3 .... removed the Rust/clippy change. The PR is only the C fix in `add_cc_sub_text` now.

PR #2334, opened 2026-09-06
[FIX] Warn when non-DVB caption PIDs are silently ignored

### LINKED ISSUE #2333: [BUG] Non-DVB caption PIDs (Teletext/ISDB/ATSC) are silently dropped when multiple exist in one program
## Current State

When a program contains multiple caption-carrying PIDs of the same non-DVB codec (Teletext, ISDB, or ATSC), `get_best_stream()` picks exactly one, and every other same-type PID gets silently discarded — no error, no warning, no log line. The user has no way to know a second language track existed and was dropped, unless they already know to inspect the PMT themselves and manually force it via `--datapid`.

DVB does not have this problem: `general_loop.c`'s extra-PID loop (added in `ad4886e7`, #447) explicitly re-processes every other DVB PID so multi-language DVB output works correctly. No equivalent exists for Teletext/ISDB/ATSC.

## Root Cause

The actual gate is in the demuxer, not the loop:

- `src/lib_ccx/ts_info.c:40-47` (`ignore_other_stream()`) and the equivalent in `ignore_other_sib_stream()` (`ts_info.c:90-97`) mark every non-selected PID as `ignore = 1` **unless its codec is `CCX_CODEC_DVB`**.
- Once ignored, `src/lib_ccx/ts_functions.c:674` (`copy_payload_to_capbuf()`) returns early for that PID, so its payload is never buffered, and `ts_functions.c:969` tears down its decoder context entirely.
- By the time `general_loop.c:1321-1370`'s DVB-only extra-PID loop runs, non-DVB extra PIDs already have no data left to process regardless of that loop's codec check.

## Evidence

Reproduced on a real broadcast recording (single program, two Teletext PIDs, not synthetic):

**Sample:** `FILE_1_TICKET_3514_Input_stream_SD_mpeg2video_1_video_8_audios_10_subs.ts`
https://samples.ffmpeg.org/ffmpeg-bugs/trac/ticket3514/

Program 1002 carries two Teletext PIDs:

Stream #0:9 [0xd49](https://github.com/CCExtractor/ccextractor/issues/spa): dvb_teletext (Initial page — magazine, not subtitles)
Stream #0:10 [0xd4a](https://github.com/CCExtractor/ccextractor/issues/dut,nor,dan,swe,por): dvb_teletext (5 declared Subtitle page languages)

Default run:

$ ccextractor --tpages-all t3514.ts
VBI/teletext stream ID 3401 (0xd49) for SID 1002 (0x3ea)
VBI/teletext stream ID 3402 (0xd4a) for SID 1002 (0x3ea)
Teletext decoder: 0 packets processed <-- 0xd4a (5-language subtitle PID)
Teletext decoder: 1663 packets processed <-- 0xd49 (magazine PID, gets kept)

`get_best_stream()` selects purely by list order and has no concept of `teletext_type` — it kept the "Initial page" magazine PID and discarded the PID explicitly declaring five "Subtitle page" languages, with zero indication to the user that this happened.

Forcing the dropped PID recovers real, decodable data:

$ ccextractor --tpages-all --datapid 3402 t3514.ts
Teletext decoder: 831 packets processed

> In this particular sample the recovered languages happen to duplicate content already present via the kept PID, so no unique captions are lost *here* — but the silent-drop behavior itself is confirmed on real broadcast data, and is trivially reproducible with any file where the second PID carries content not present elsewhere. Confirmed independently on a synthetic split of `dvbteletext.ts` where the drop does cost real, non-duplicate caption lines.

## Proposed Fix (scoped)

This issue is scoped to the silent-drop / missing-warning problem only. Full multi-PID decoding support for Teletext/ISDB/ATSC is a separate, larger effort — it would require generalizing per-PID decoder allocation in `lib_ccx.c:352-374`, per-language encoder routing in `lib_ccx.c:452-466`, and storing ISO-639 codes into `cinfo->lang` for Teletext at `ts_tables.c:474`, none of which exist today for non-DVB codecs.

**For this issue:** add a warning in `ignore_other_stream()` / `ignore_other_sib_stream()` when a stream being ignored is a non-DVB caption codec (Teletext/ISDB/ATSC), naming the PID and pointing the user to `--datapid` to retrieve it manually. No decoding behavior changes.

## Status

Confirmed, reproducible on real broadcast data

kaihere14:
Closes #2333

When a program contains multiple caption-carrying PIDs of the same non-DVB codec (Teletext, ISDB, or ATSC), only one is ever decoded — every other same-type PID is silently discarded with no warning, no log line, and no indication to the user that a second language track existed.

This adds a warning at the actual root cause: the DVB-only exemption in `ignore_other_stream()` and `ignore_other_sib_stream()` (`src/lib_ccx/ts_info.c`). A new static helper,
`warn_ignored_caption_stream()`, prints the ignored PID and points the user to `--datapid` to recover it manually. DVB is explicitly excluded since it already has its own extra-PID handling and doesn't need this.

**Scope**: this is a warning only. It does NOT add multi-PID decoding support for Teletext/ISDB/ATSC — see #2333 for why that's a separate, larger effort (per-PID decoder allocation, per-language encoder routing, and ISO-639 storage for Teletext don't exist today for non-DVB codecs).

**Real broadcast sample** (two Teletext PIDs, one program — from samples.ffmpeg.org/ffmpeg-bugs/trac/ticket3514):

$ ccextractor --tpages-all t3514.ts
VBI/teletext stream ID 3401 (0xd49) for SID 1002 (0x3ea) VBI/teletext stream ID 3402 (0xd4a) for SID 1002 (0x3ea) Warning: Teletext caption stream ID 3402 (0xd4a) for SID 1002 (0x3ea) will be ignored - only one caption stream per program is extracted. Use --datapid 3402 to extract it in a separate run.

Correct PID, correct SID. The 8 DVB subtitle PIDs in the same file stay silent, as expected.

**Synthetic two-PID split** (dvbteletext.ts split across two PIDs): same correct behavior, warns on the dropped PID (202/0xca).

**ISDB and ATSC branches** — verified against a constructed sample (`mixed_codecs.ts`, built from `dvbteletext.ts` by injecting two ES entries into its PMT: PID 0xca as ISDB via a data_component descriptor, PID 0xcb as ATSC via a caption_service descriptor):

$ ccextractor --tpages-all mixed_codecs.ts
VBI/teletext stream ID 201 (0xc9) for SID 201 (0xc9) *****ISDB subtitles detected
Warning: ISDB caption stream ID 202 (0xca) for SID 201 (0xc9) will be ignored - only one caption stream per program is extracted. Use --datapid 202 to extract it in a separate run. Warning: ATSC caption stream ID 203 (0xcb) for SID 201 (0xc9) will be ignored - only one caption stream per program is extracted. Use --datapid 203 to extract it in a separate run.

Teletext output (`mixed_codecs_p694.srt`, `_p765.srt`) is byte-identical to the unmodified `dvbteletext.ts` baseline — the injected PIDs change nothing except triggering the new warnings.

Note: I could not source a real multi-PID ISDB recording — the one public ARIB sample I found (samples.ffmpeg.org/MPEG2/subcc, single PID) has a multi-TS-packet PMT my synthetic splitter doesn't handle, so the ISDB/ATSC branches above are exercised by a constructed sample rather than a genuine broadcast recording. Flagging this rather than implying otherwise — the Teletext path is verified on real broadcast data (see above), ISDB/ATSC are verified by construction.

**DVB control** — confirms no regression: ran both a single-PID and a split two-PID DVB sample before and after the patch. No new warnings on either, and output is byte-identical (PNG counts and `diff -rq` match exactly pre- and post-patch). All decoded `.srt` files from the Teletext samples are also byte-identical to their pre-patch versions — this change only adds a print statement, no decoding logic is touched.

**Caught during testing, fixed before this PR**: the first version of this warned on every video PID in every file, because 608/708 captions are internally tagged `CCX_CODEC_ATSC_CC` even though they live inside the video stream, not a separate PID (`ts_tables.c:492-498`). Fixed by excluding `CCX_STREAM_TYPE_VIDEO_MPEG2/_H264/_HEVC` from the warning. Re-verified against a European DVB file (now silent, as it should be) and a US ATSC file with real 608/708 captions in video (still silent, correct — genuine caption-only ATSC PIDs still warn correctly).

**Known cosmetic issue, pre-existing, not introduced by this patch**: under `--pmt`, the ATSC warning can visually run together with a preceding debug line (`ts_tables.c:445-449` emits its `dbg_print` without a trailing newline, so whatever prints next collides with it on the same line). This only occurs with `--pmt` and would happen with any output following that debug line, not just this warning. Out of scope for this PR per its stated scope, noting it here so it isn't mistaken for something this patch caused.

**Reason for this PR:**

**Sanity check:**
- [x] I have read and understood the contributors guide.

### DIFF
--- a/src/lib_ccx/ts_info.c
+++ b/src/lib_ccx/ts_info.c
@@ -37,13 +37,62 @@ int need_cap_info(struct ccx_demuxer *ctx, int program_number)
 	return CCX_FALSE;
 }
 
+/**
+	Warn once about a caption stream that is about to be discarded because it is not
+	the single PID that get_best_stream() selected for this program.
+
+	DVB subtitles are deliberately excluded: extra DVB PIDs are picked up later by the
+	multi-language loop in process_non_multiprogram_general_loop(), so they are not lost
+	and must stay silent. Teletext, ISDB and ATSC caption PIDs have no such handling, so
+	without this notice they would disappear with no indication at all (issue #2333).
+
+	The caller only reaches this for a stream whose ignore flag is still clear, which
+	keeps the message to one line per PID instead of one per demuxer iteration.
+*/
+static void warn_ignored_caption_stream(struct cap_info *iter)
+{
+	const char *codec_name;
+
+	/* Every video elementary stream is registered as CCX_CODEC_ATSC_CC because 608/708
+	   can ride inside it, so an ignored video PID is not a caption stream the user lost
+	   and --datapid is not the way to reach it. Those PIDs also keep being fed to the
+	   decoder when --analyze-video-stream is on. Warn only about the caption-only PIDs. */
+	if (iter->stream == CCX_STREAM_TYPE_VIDEO_MPEG2 ||
+	    iter->stream == CCX_STREAM_TYPE_VIDEO_H264 ||
+	    iter->stream == CCX_STREAM_TYPE_VIDEO_HEVC)
+		return;
+
+	switch (iter->codec)
+	{
+		case CCX_CODEC_TELETEXT:
+			codec_name = "Teletext";
+			break;
+		case CCX_CODEC_ISDB_CC:
+			codec_name = "ISDB";
+			break;
+		case CCX_CODEC_ATSC_CC:
+			codec_name = "ATSC";
+			break;
+		default:
+			return;
+	}
+
+	mprint("Warning: %s caption stream ID %u (0x%x) for SID %u (0x%x) will be ignored - only one caption stream per program is extracted.\n",
+	       codec_name, iter->pid, iter->pid, iter->program_number, iter->program_number);
+	mprint("         Use --datapid %u to extract it in a separate run.\n", iter->pid);
+}
+
 void ignore_other_stream(struct ccx_demuxer *ctx, int pid)
 {
 	struct cap_info *iter;
 	list_for_each_entry(iter, &ctx->cinfo_tree.all_stream, all_stream, struct cap_info)
 	{
 		if (iter->pid != pid && iter->codec != CCX_CODEC_DVB)
+		{
+			if (!iter->ignore)
+				warn_ignored_caption_stream(iter);
 			iter->ignore = 1;
+		}
 	}
 }
 
@@ -93,7 +142,11 @@ void ignore_other_sib_stream(struct cap_info *head, int pid)
 	list_for_each_entry(iter, &head->sib_head, sib_stream, struct cap_info)
 	{
 		if (iter->pid != pid && iter->codec != CCX_CODEC_DVB)
+		{
+			if (!iter->ignore)
+				warn_ignored_caption_stream(iter);
 			iter->ignore = 1;
+		}
 	}
 }
 


kaihere14:
Hey @cfsmp3 , PR for #2333 up whenever you get a sec

cfsmp3 (changes requested):
Good write-up — particularly calling out that ISDB/ATSC are exercised by a constructed sample rather than real broadcast data, and that the video-PID false positive was caught and fixed before submitting. That is the part reviewers usually have to find themselves.

Three things from the repo's code-quality rules before this can go in:

**Braces.** All `if`/`else`/`for`/`while` bodies need `{ }`, including single statements. Three places here:

```c
	if (iter->stream == CCX_STREAM_TYPE_VIDEO_MPEG2 || … )
		return;

	if (!iter->ignore)
		warn_ignored_caption_stream(iter);   // twice
```

**Format specifiers.** `pid` and `program_number` are `int` (`ccx_demuxer.h:55-56`) but printed with `%u` and `%x`. Fine for real PID values, wrong for the declared type — please use `%d`, or cast explicitly if unsigned output is intended.

**What I could verify:** no over-warning. Across 59 local samples the warning never fires, every output is byte-identical to master, and exit codes match — so the video-PID exclusion is doing its job and nothing regressed.

**What I could not:** reach the warning at all. The only two local files with multiple `dvb_teletext` entries turn out to expose the same PID (`0x12e`) in a single program, so ccextractor sees one caption PID and nothing is discarded. The positive path therefore rests on your samples. If `t3514.ts` is small enough to add to the regression suite it would be worth doing — this is precisely the kind of behaviour that silently regresses otherwise.

One question on scope: `ignore_other_sib_stream()` gets the same treatment, but its callers iterate sibling streams of one program. Is it possible for both functions to warn about the same PID in one run, and if so is the message emitted twice?

— Claude on behalf of @cfsmp3

kaihere14:
> Good write-up — particularly calling out that ISDB/ATSC are exercised by a constructed sample rather than real broadcast data, and that the video-PID false positive was caught and fixed before submitting. That is the part reviewers usually have to find themselves.
> 
> Three things from the repo's code-quality rules before this can go in:
> 
> **Braces.** All `if`/`else`/`for`/`while` bodies need `{ }`, including single statements. Three places here:
> 
> ```c
> 	if (iter->stream == CCX_STREAM_TYPE_VIDEO_MPEG2 || … )
> 		return;
> 
> 	if (!iter->ignore)
> 		warn_ignored_caption_stream(iter);   // twice
> ```
> 
> **Format specifiers.** `pid` and `program_number` are `int` (`ccx_demuxer.h:55-56`) but printed with `%u` and `%x`. Fine for real PID values, wrong for the declared type — please use `%d`, or cast explicitly if unsigned output is intended.
> 
> **What I could verify:** no over-warning. Across 59 local samples the warning never fires, every output is byte-identical to master, and exit codes match — so the video-PID exclusion is doing its job and nothing regressed.
> 
> **What I could not:** reach the warning at all. The only two local files with multiple `dvb_teletext` entries turn out to expose the same PID (`0x12e`) in a single program, so ccextractor sees one caption PID and nothing is discarded. The positive path therefore rests on your samples. If `t3514.ts` is small enough to add to the regression suite it would be worth doing — this is precisely the kind of behaviour that silently regresses otherwise.
> 
> One question on scope: `ignore_other_sib_stream()` gets the same treatment, but its callers iterate sibling streams of one program. Is it possible for both functions to warn about the same PID in one run, and if so is the message emitted twice?
> 
> — Claude on behalf of @cfsmp3

Thanks for the review. All three addressed in the follow-up commit:

Braces — added to the video-type early return in warn_ignored_caption_stream() and to both if (!iter->ignore) call sites in ignore_other_stream() / ignore_other_sib_stream().

Format specifiers — pid and program_number now print with %d. The hex forms keep %x with an explicit (unsigned int) cast, since hex output is the intent there (matches the existing VBI/teletext stream ID ... (0x...) lines the warning sits next to).

Regression sample — t3514.ts is a real broadcast capture from samples.ffmpeg.org; I'll check its size and, if it's reasonable, open a request to add it to the sample repo so the positive path is covered rather than resting on my local files. Agreed this is exactly the kind of thing that regresses silently.

On the scope question — no, the two functions can't both warn on the same PID in one run, for two reasons:

1. The callers are on opposite branches of if (!ctx->multiprogram) in general_loop.c:1447. The non-multiprogram path calls ignore_other_stream() (general_loop.c:1151); the multiprogram path iterates pg_stream and calls ignore_other_sib_stream() (general_loop.c:1482). ctx->multiprogram comes from the CLI and doesn't change mid-run, so only one of the two functions ever executes per process.
2. Within either function, the if (!iter->ignore) guard means a PID warns once even though the caller runs on every demuxer iteration. The only place ignore is reset to 0 is update_capinfo() (ts_info.c:274), and that branch only fires while the entry is still CCX_STREAM_TYPE_UNKNOWNSTREAM or CCX_CODEC_NONE. Once the PMT has typed the stream the condition is false and the flag stays set, so a repeated PMT can't re-arm it.

The one edge case is a PID first registered as CCX_CODEC_NONE (ts_tables.c:510) and later re-typed by a new PMT version: update_capinfo() resets ignore, and the next iteration warns. But the first pass would have hit default: return in the codec switch (codec was still NONE), so the user still sees exactly one warning.

PR #2337, opened 2026-09-08
[FIX] Use fixed-width format specifiers for 64-bit values (corrupted X-TIMESTAMP-MAP on Windows)

### LINKED ISSUE #2336: [BUG] X-TIMESTAMP-MAP MPEGTS value is truncated on Windows: int64_t printed with %ld (plus 23 further format mismatches hidden by missing printf attributes)
CCExtractor version: master (907be052)

# Necessary information

- Is this a regression (i.e. did it work before)? **NO** — long-standing on all Windows builds.
- What platform did you use? **Windows** (reproduced with MinGW-w64 UCRT gcc 15.2, x86_64; the same LLP64 ABI as the shipped MSVC x64/Win32 binaries)
- What were the used arguments? `--timestamp-map -o out.vtt`

# Video links

**No sample is required to reproduce this** — it is a compile-time-provable type/format mismatch, and I've included a self-contained 40-line C program below that reproduces it deterministically. I'm happy to supply a TS sample if you'd still like one for the regression suite, but the defect is verifiable without one. (Details on methodology at the bottom.)

# Additional information

## Summary

`X-TIMESTAMP-MAP` in WebVTT output carries a **wrong MPEGTS value on every Windows build**, because a 64-bit `LLONG` is printed with `%ld`. On Windows `long` is 32 bits (LLP64 on x64, ILP32 on Win32), so the value is truncated.

This is the field HLS players use to anchor a WebVTT segment to the MPEG-TS presentation clock, so the result is caption desync or a rejected segment.

## The bug

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_webvtt.c#L224-L226

```c
snprintf(header_string, sizeof(header_string), "X-TIMESTAMP-MAP=LOCAL:%02u:%02u:%02u.%03u,MPEGTS:%ld%s",
         h1, m1, s1, ms1, context->timing->sync_pts2fts_pts,
         ccx_options.enc_cfg.line_terminator_lf ? "\n\n" : "\r\n\r\n");
```

`sync_pts2fts_pts` is `LLONG` (`ccx_common_timing.h:55`), and `typedef int64_t LLONG` (`ccx_common_platform.h:118`). It is printed with `%ld`, i.e. `long`.

GCC flags the real source line and names the fix:

```
src/lib_ccx/ccx_encoders_webvtt.c:225:64: warning: format '%ld' expects argument of type
'long int', but argument 8 has type 'LLONG' {aka 'long long int'} [-Wformat=]
  225 | ... "X-TIMESTAMP-MAP=LOCAL:%02u:%02u:%02u.%03u,MPEGTS:%ld%s",
  226 |     h1, m1, s1, ms1, context->timing->sync_pts2fts_pts,
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                    LLONG {aka long long int}
      | note: format string is defined here ... %lld
```

## Reproduction

```c
#include <stdio.h>
#include <stdint.h>
typedef int64_t LLONG;

int main(void) {
    char h[200];
    unsigned h1=0,m1=0,s1=0,ms1=0;
    LLONG pts = 8589934592LL;   /* 2^33 - an ordinary 90kHz broadcast PTS */

    printf("sizeof(long) = %d\n", (int)sizeof(long));
    snprintf(h, sizeof h, "X-TIMESTAMP-MAP=LOCAL:%02u:%02u:%02u.%03u,MPEGTS:%ld",
             h1,m1,s1,ms1, pts);
    printf("current (%%ld)  -> %s\n", h);
    snprintf(h, sizeof h, "X-TIMESTAMP-MAP=LOCAL:%02u:%02u:%02u.%03u,MPEGTS:%lld",
             h1,m1,s1,ms1, pts);
    printf("fixed   (%%lld) -> %s\n", h);
    return 0;
}
```

Built with `gcc -Wall -Wformat` on Windows x86_64:

```
sizeof(long) = 4
current (%ld)  -> X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:0
fixed   (%lld) -> X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:8589934592
```

A PTS of 2^33 emits **`MPEGTS:0`**. A PTS of 2^31 (≈6.6 h of 90 kHz ticks) emits **`MPEGTS:-2147483648`** — negative, which is invalid for this field. Broadcast streams routinely start at an arbitrary offset in the 33-bit PTS space, so this is not an edge case.

On **Win32** it is worse: cdecl varargs are stack-packed with no promotion, so `%ld` consumes only 4 of the 8 bytes and the trailing `%s` reads the PTS's high dword as a `char *`. For a 33-bit PTS that dword is 0 or 1 — `0` prints `(null)` and destroys the header's terminating blank line; `1` dereferences address `0x1`.

On Linux/macOS (LP64) `long` is 64-bit, so `%ld` happens to be correct and `-Wformat` stays silent. That is why this survived.

## Why the whole class is invisible to the build

`mprint()` and `dbg_print()` are the project's logging front ends, and **neither carries `__attribute__((format(printf, ...)))`** — there is no such annotation anywhere in `src/lib_ccx`. GCC/Clang therefore never check any of those call sites. Only direct `printf`/`snprintf` calls get checked.

I added the two attributes locally and rebuilt `src/lib_ccx/*.c` with `-Wformat`. That surfaces **24 warnings across 10 files**, all currently silent:

| File:line | Specifier | Actual argument type |
|---|---|---|
| **ccx_encoders_webvtt.c:225** | `%ld` | `LLONG` — **corrupts output file** |
| **ts_functions.c:345** | `%d` | `unsigned char *` — see below |
| general_loop.c:1676-1680 (×5) | `%ld` | `int64_t` |
| general_loop.c:281 | `%d` | `size_t` |
| ts_functions.c:1024 | `%ld` | `int64_t` |
| ts_functions.c:93 | `%lx` | `uint64_t` |
| avc_functions.c:125, 191, 778 | `%d` | `LLONG` |
| dvb_subtitle_decoder.c:1916, 1925 | `%d` | `LLONG` |
| matroska.c:1348, 1594 | `%d` | `ULLONG` |
| telxcc.c:1485 | `%lx` | `uint64_t` |
| telxcc.c:593 (×2) | `%d` | `size_t` |
| wtv_functions.c:325 | `%x` | `uint64_t` |
| wtv_functions.c:549 | `%ld` | `int64_t` |
| networking.c:825 | `%d` | `size_t` |

Sites that already cast correctly (`params_dump.c:261-272`, `asf_functions.c:187/415/501`, `spupng.c:810/934`, `ccextractor.c:397`) are of course fine, and `networking.c`'s `%ld` with `WSAGetLastError()` is correct.

### Second real defect: `ts_functions.c:345`

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ts_functions.c#L345-L347

```c
dbg_print(CCX_DMT_PARSE, "TS pid: %d  PES start: %d  counter: %u  payload length: %u  adapt length: %d\n",
          payload->pid, payload->start, payload->counter, payload->length,
          (int)(adaptation_field_length));
```

`payload->start` is `unsigned char *` (the payload data pointer). The field the label wants is `payload->pesstart` ("PES or PSI start"). So the `-debug` TS trace prints a truncated pointer as "PES start", and on 64-bit it desynchronizes the three arguments after it — `counter`, `payload length` and `adapt length` are all shifted and wrong. Anyone reading this trace to debug TS parsing is being actively misled.

## Proposed fix

1. `ccx_encoders_webvtt.c:225` — `%ld` → `%lld`. One token; fixes the user-visible output corruption. `%lld` is already the repo convention (50 existing uses).
2. `ts_functions.c:345` — `payload->start` → `payload->pesstart`.
3. The remaining specifier mismatches — `%lld` / `%llu` / `%llx` / `%zu` as appropriate.
4. Add `__attribute__((format(printf, 2, 3)))` to `dbg_print` and `(1, 2)` to `mprint`, guarded for MSVC, so this class cannot regress.

Output on Linux/macOS is **byte-identical** before and after (both specifiers consume 64 bits on LP64). Only Windows output changes, from wrong to right. No signature, struct, option, or format change.

I'd like to work on this if it's not already taken — happy to split it (1+2 as the bug fix, 3+4 as a follow-up) if you'd prefer smaller reviews.

## Methodology

To be transparent: I found this by static analysis of the source with AI assistance, then **verified every claim on real hardware** — compiling the actual file with MinGW-w64 on Windows to get the `-Wformat` diagnostic quoted above, and running the reproduction program to capture the real output. The 24-site table is compiler output, not inference. Nothing here is a theoretical finding; I understand from `PULL_REQUEST_TEMPLATE.md` that unverified AI-generated reports are a burden, and I've tried to make this one carry its own proof.

pranayr710:
Fixes #2336

Reason for this PR:

- [x] This PR fixes a bug that I have personally experienced or that a real user has reported and for which a sample exists. *(see the caveat on "sample" under Repro below — this one is reproducible without a media file, and I've included an exact reproduction)*

Sanity check:
- [x] This is a bug fix, so I have NOT added it to the changelog.
- [x] I am NOT adding new C code — this only corrects format specifiers and one wrong field reference in existing lines.

---

## What was wrong

`X-TIMESTAMP-MAP` carried a corrupted MPEGTS value on **every Windows build**:

```c
// ccx_encoders_webvtt.c:224
snprintf(..., "X-TIMESTAMP-MAP=LOCAL:%02u:%02u:%02u.%03u,MPEGTS:%ld%s",
         h1, m1, s1, ms1, context->timing->sync_pts2fts_pts, ...);
```

`sync_pts2fts_pts` is `LLONG` → `int64_t`, printed with `%ld`. `long` is 32 bits on Windows (LLP64 on x64, ILP32 on Win32), so the value is truncated. HLS players use this field to anchor a WebVTT segment to the transport clock.

Linux/macOS are LP64 where `long` is 64 bits, so `%ld` is accidentally correct there — which is why this survived.

## Repro instructions

**No media sample is needed** — this is a type/format mismatch that the compiler itself diagnoses. I understand the template asks for a sample and why; I'm happy to hunt one down for the regression suite if you want it, but I don't think a file is what makes this one verifiable. Here is the proof instead.

**1. The compiler flags the real line.** Building the actual file with MinGW-w64 (same LLP64 ABI as the shipped MSVC binaries):

```
$ gcc -fsyntax-only -Wformat -DDISABLE_RUST -Isrc -Isrc/lib_ccx ... src/lib_ccx/ccx_encoders_webvtt.c

src/lib_ccx/ccx_encoders_webvtt.c:225:64: warning: format '%ld' expects argument of type
'long int', but argument 8 has type 'LLONG' {aka 'long long int'} [-Wformat=]
  226 |     h1, m1, s1, ms1, context->timing->sync_pts2fts_pts,
      |                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      | note: format string is defined here ... %lld
```

**2. Standalone runnable reproduction** (`gcc -Wall repro.c && ./repro` on Windows x86_64):

```c
#include <stdio.h>
#include <stdint.h>
typedef int64_t LLONG;
int main(void) {
    char h[200]; unsigned h1=0,m1=0,s1=0,ms1=0;
    LLONG pts = 8589934592LL;             /* 2^33, an ordinary 90kHz PTS */
    printf("sizeof(long) = %d\n", (int)sizeof(long));
    snprintf(h,sizeof h,"MPEGTS:%ld",  pts); printf("current -> %s\n", h);
    snprintf(h,sizeof h,"MPEGTS:%lld", pts); printf("fixed   -> %s\n", h);
    return 0;
}
```

Actual output:

```
sizeof(long) = 4
current -> MPEGTS:0
fixed   -> MPEGTS:8589934592
```

A PTS of 2^31 (≈6.6 h of 90 kHz ticks) prints `MPEGTS:-2147483648` — negative, which is invalid for this field. On Win32 it is worse: cdecl varargs are stack-packed, so `%ld` eats 4 of 8 bytes and the trailing `%s` reads the PTS high dword as a `char *`.

## Why `PRId64` and not `%lld`

`%lld` would fix Windows but break Linux, where `int64_t` is `long`, not `long long`. The `<inttypes.h>` macros are correct on every target, are already used in `ccx_demuxer_mxf.c`, `ccx_encoders_curl.c`, `dvb_subtitle_decoder.c` and `telxcc.c`, and `ccx_common_platform.h` already includes `<inttypes.h>`, so no new include is needed.

## Scope: 22 sites, one root cause

`mprint()` and `dbg_print()` carry no `__attribute__((format(printf, ...)))`, so the compiler never checks those call sites — the whole class is invisible to the build. I added the attributes locally, rebuilt `src/lib_ccx/*.c` with `-Wformat`, and fixed everything it found:

- `PRId64` / `PRIu64` / `PRIx64` for `int64_t` / `uint64_t` / `LLONG` / `ULLONG`
- `%zu` for `size_t`
- explicit `(long long)` cast for pointer differences

After the fix that build is clean (the only three diagnostics left are checker artifacts, not defects — details in #2336).

### One genuine second bug

`ts_functions.c:345` printed `payload->start`, an `unsigned char *`, under the label **"PES start"**. The intended field is `payload->pesstart`. The pointer was truncated to `int` and shifted the three arguments after it, so `counter`, `payload length` and `adapt length` were all wrong in every `-debug` TS trace.

## Risk

Output on Linux and macOS is **byte-identical** before and after. Only Windows output changes, from wrong to right. No signature, struct, option or output-format change. `clang-format` clean.

## Deliberately left out

- `matroska.c:1348/1594` use the `LLD` macro, which expands to `%I64d` on Windows and is **correct**. GCC only reported it because its `gnu_printf` checker skips the MSVC `I64` modifier. Untouched.
- I did **not** add the `format(printf)` attributes to `mprint`/`dbg_print` in this PR, even though that is what would stop this class from coming back. I can only test the Windows leg locally, and I'd rather not surface unknown warnings on the Linux build without your say-so. Happy to send it as a follow-up if you want it.

---

*Disclosure: I found this with AI assistance, then verified every claim on real hardware — the compiler diagnostics and program output above are captured from actual runs on Windows, not inferred. I saw the note in the PR template about theoretical AI-generated fixes and tried to make this one carry its own proof.*


pranayr710:
Rebased onto master (now includes #2342) and, more importantly, **actually ran the patched binary** — which I had not done when I opened this.

## Why a plain output diff would not have proved anything

21 of the 22 changed sites are `mprint()` / `dbg_print()` calls. A normal run never reaches them, so comparing only the `.vtt`/`.srt` files would pass trivially while saying nothing about the lines this PR touches. So the comparison below captures **exit code + output-file hash + stdout hash** for every format, and runs the whole matrix again under the debug flags that actually reach those code paths.

Built master and this branch from identical source in the same image (`docker/Dockerfile`, `BUILD_TYPE=minimal`):

| Configuration | Result |
|---|---|
| plain | **13/13 identical** |
| `--debug` | **13/13 identical** |
| `--parsedebug` | **13/13 identical** |
| `--debug --parsedebug --xdsdebug --debugdvbsub` | **13/13 identical** |

Formats: webvtt, srt, sami, ttxt, txt, scc, ssa, smptett, g608, mcc, spupng, simple_xml, rcwt.

That is the expected result on Linux and is the point: `PRId64` expands to `"ld"` on LP64, i.e. the exact specifier that was already there, so these changes are provably inert on Linux and only alter the Windows builds — from wrong to right.

One thing worth flagging so nobody repeats it: my first pass reported a `spupng` difference. That was an artefact of my harness merging stdout and stderr with `2>&1` — the same error line landed at a different position between runs. With the streams separated, master and this branch produce identical stdout **and** identical stderr, deterministic across repeated runs. Not a real difference.

The Windows-side evidence (MSVC 2022 `%ld` truncating a 2^33 PTS to `MPEGTS:0`, and the `-Wformat` diagnostic on the real source line) is unchanged from the description above.

PR #2339, opened 2026-09-09
[FIX] webvtt: don't leak basefilename or skip the header flag when CSS creation fails

### LINKED ISSUE #2338: [BUG] write_webvtt_header(): --webvtt-create-css leaks basefilename, and re-emits the header before every cue when the CSS file can't be created
CCExtractor version: master (907be052)

# Necessary information

- Is this a regression (i.e. did it work before)? **NO** — present since `--webvtt-create-css` was added.
- What platform did you use? **All** (platform-independent; this is a logic bug, not a portability one)
- What were the used arguments? `--out=webvtt --webvtt-create-css`

# Video links

Not needed — any input that produces WebVTT output reproduces this. The trigger is the output *directory*, not the input file. Steps are below.

# Additional information

`write_webvtt_header()` in `src/lib_ccx/ccx_encoders_webvtt.c` has two defects in its `--webvtt-create-css` path.

## 1. Header re-emitted before every cue when the CSS file can't be created

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_webvtt.c#L255-L261

```c
FILE *f = fopen(css_file_name, "wb");
if (f == NULL)
{
        mprint("Warning: Error creating the file %s\n", css_file_name);
        free(css_file_name);
        return;                 // <-- skips wrote_webvtt_header = 1
}
```

The function guards its own re-entry with `if (context->wrote_webvtt_header) return;` at the top, and sets that flag on the **last line**:

```c
context->wrote_webvtt_header = 1; // Do it even if couldn't write the header, because it won't be possible anyway
```

That trailing comment states the intent explicitly: the flag should be set even when the header could not be fully written. The `return` above bypasses it.

`write_webvtt_header()` is called from `write_stringz_as_webvtt()`, `write_cc_bitmap_as_webvtt()` and `write_cc_buffer_as_webvtt()` — i.e. **once per cue**. So when the CSS file cannot be created, the flag never gets set and the whole header block (either the `X-TIMESTAMP-MAP` line or the blank line, depending on `--timestamp-map`) is written again **before every subsequent cue**, in the middle of the file. The resulting `.vtt` is not valid WebVTT.

**Repro:**
```bash
mkdir ro && cd ro
ccextractor ../any_input.ts --out=webvtt --webvtt-create-css -o out.vtt
# make the CSS creation fail, e.g. by making the CWD read-only first:
#   chmod a-w .        (Linux/macOS)
# then inspect out.vtt - the header block repeats before each cue
```
(`fopen` for the `.css` uses the *current working directory*, not the output path, so a read-only CWD is enough — which is itself a little surprising.)

## 2. `basefilename` is leaked on every path

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_webvtt.c#L247

```c
char *basefilename = get_basename(context->first_input_file);
```

`get_basename()` returns `malloc`ed memory (`utility.c`), but `basefilename` is never freed — not on the success path, not on the `fopen`-failure path. Every other caller in the tree frees it (`ccx_encoders_common.c:607`, `:636`, `:1369`, `:1437`).

Normally that is a single small leak per output file. Combined with defect 1 it becomes a leak **per cue**, since the function then re-runs for the whole recording.

## 3. Minor: unguarded `get_basename()` return

`get_basename()` returns `NULL` if its argument is `NULL` or on allocation failure, and `strlen(basefilename)` on the next line would then crash. `ccx_encoders_common.c:636-641` guards exactly this case:

```c
char *basefilename = get_basename(ctx->first_input_file);
...
if (basefilename == NULL)
{
        basefilename = get_basename("untitled");
}
```

To be straight about this one: I could **not** establish that `first_input_file` is ever actually `NULL` on a live run — the Rust parser leaves it as an empty string rather than `NULL` (`parser.rs:1618-1620`), so in practice `get_basename()` returns `""`. I'm flagging it as hardening consistent with the existing guard, not as a demonstrated crash. Happy to drop it from the patch if you'd rather keep the diff to the two real defects.

## Verification

I confirmed defects 1 and 2 by reading the code and compiling; both are static facts about the control flow (a `return` that skips an assignment, and an allocation with no matching `free`) rather than behavioural guesses, so they're checkable in about twenty lines. **I did not execute the `fopen`-failure path**, since I don't have a full local build — so I'd appreciate a sanity check on the repro above rather than taking my word for the runtime effect.

PR follows.

pranayr710:
Fixes #2338

Reason for this PR:

- [x] This PR fixes a bug that I have personally experienced and for which a sample exists (inline below, 153 bytes).

Sanity check:
- [x] Bug fix, so NOT added to the changelog.
- [x] Not adding new C code — this restructures an existing error path and adds the missing `free`.

---

### 1. The header block is re-emitted before every cue

`write_webvtt_header()` is called **once per cue** (from `write_stringz_as_webvtt`, `write_cc_bitmap_as_webvtt` and `write_cc_buffer_as_webvtt`) and guards re-entry with `context->wrote_webvtt_header`, which it sets on its final line:

```c
context->wrote_webvtt_header = 1; // Do it even if couldn't write the header, because it won't be possible anyway
```

The `--webvtt-create-css` path returned early when the `.css` file could not be opened, skipping that assignment:

```c
if (f == NULL)
{
        mprint("Warning: Error creating the file %s\n", css_file_name);
        free(css_file_name);
        return;              // <-- flag never set
}
```

The comment quoted above already says the flag was meant to be set regardless.

### 2. `basefilename` is never freed

It comes from `get_basename()`, which allocates, and was not freed on **any** path. Every other caller in the tree frees it (`ccx_encoders_common.c:607`, `:636`, `:1369`, `:1437`). Alone that is one small leak per output file; combined with the early return it became a leak per cue.

## The change

- `return` → `else` branch, so `wrote_webvtt_header = 1` is always reached and there is still only one assignment to it
- `free(css_file_name)` moved after the branch so both paths release it
- `free(basefilename)` added once the name has been formatted

## Repro

```bash
base64 -d > sample.rcwt <<'B64EOF'
B64EOF

mkdir -p ro out && mv sample.rcwt ro/
docker run --rm -v $PWD/ro:/ro:ro -v $PWD/out:/out ccx     /ro/sample.rcwt --out=webvtt --webvtt-create-css -o /out/o.vtt
```

sha256 `1e7e3f58151190124b113facc7543ac5ee6d7e230830cda88744de19cc17250c` — the same reproducer as #2341, three EIA-608 CC1 pop-on captions. The `.css` is written next to the **input** file (`get_basename()` keeps the path and only strips the extension), so a read-only input directory is what makes `fopen` fail.

## Measured

| | master | this branch |
|---|---|---|
| `"Error creating the file"` on a 3-cue file | **4x** | **1x** |
| blank lines in output | 7 | 4 |
| `valgrind` definitely lost | **96 bytes / 5 blocks** | **32 bytes / 1 block** |
| 13-format regression vs master | — | **13/13 identical** |

The 64 bytes in 4 blocks that disappear are exactly the records valgrind attributes to `get_basename() <- write_webvtt_header()`. **4 leaked blocks for 4 calls** — the leak count corroborates the re-entry count independently. The residual 32 bytes is present on master too and is not in this code path.

`clang-format` clean. Rebased on master including #2342.


pranayr710:
**Update: I've now executed the failure path, so the caveat in the PR description can go. It reproduces.**

I built master in Docker and ran it against a 153-byte RCWT input with three cues, with the `.css` target on a read-only mount so `fopen` fails:

```bash
EOF
docker run --rm -v $PWD/ro:/ro:ro -v $PWD/out:/out ccx:master     /ro/sample.rcwt --out=webvtt --webvtt-create-css -o /out/ev.vtt
```

```
Warning: Error creating the file /ro/sample.css
Warning: Error creating the file /ro/sample.css
Warning: Error creating the file /ro/sample.css
Warning: Error creating the file /ro/sample.css
```

**Four warnings for a three-cue file** — one per cue plus one from `write_subtitle_file_footer()`. `write_webvtt_header()` is re-entering every time, which is only possible because `wrote_webvtt_header` is never set on the early-return path. That is the defect this PR fixes.

The visible damage in the output, with the blank lines marked:

```
WEBVTT

00:00:01.001 --> 00:00:02.999 line:84.66%
FIRST CAPTION
                      <-- extra
                      <-- blank line
00:00:03.001 --> 00:00:04.999 line:84.66%
SECOND CAPTION
                      <-- extra
                      <-- blank line
00:00:05.001 --> 00:00:06.999 line:84.66%
```

Each re-entry runs the `else` branch ("Must have another newline if X-TIMESTAMP-MAP is not used") and injects an extra `CRLF` before the next cue. With `--webvtt-create-css --timestamp-map` it would be a repeated `X-TIMESTAMP-MAP` line instead of a blank one — I could not demonstrate that particular variant, because `--timestamp-map` on this input path segfaults for an unrelated reason, which I've filed separately as #2341.

One correction to my own report while I'm here: I originally described the corruption as the *header block* repeating. On the default path the header block is just a newline, so what a user actually sees is doubled blank lines rather than a repeated `WEBVTT`/`X-TIMESTAMP-MAP` line. The mechanism is exactly as described; the visible symptom is milder than my wording implied.

The `basefilename` leak in the same function is unchanged and unaffected by any of this — it is an allocation with no matching `free` on any path.

pranayr710:
Rebased onto master (now includes #2342), **dropped the speculative part**, and measured both defects instead of asserting them.

## Removed the NULL guard

I had included a `get_basename() == NULL` check and said in the description that I could not show it was reachable. That was the right thing to admit and the wrong thing to leave in a patch, so it is gone. What remains is only the two defects I can demonstrate:

+19/−13, down from +25/−13.

## 1. Header re-entry — measured

Same 153-byte RCWT reproducer as #2341, with the `.css` target on a read-only mount so `fopen` fails:

| | `"Error creating the file"` printed | blank lines in output |
|---|---|---|
| master | **4×** | 7 |
| this branch | **1×** | 4 |

Four warnings for a **three-cue** file: once per cue plus once from `write_subtitle_file_footer()`. That is `write_webvtt_header()` re-entering every time, which is only possible while `wrote_webvtt_header` stays unset. After the fix it runs once and the blank-line structure is correct.

## 2. The leak — measured

`valgrind --leak-check=full --show-leak-kinds=definite`, same invocation:

```
master        definitely lost: 96 bytes in 5 blocks
this branch   definitely lost: 32 bytes in 1 blocks
```

The 64 bytes in 4 blocks that disappear are exactly the records valgrind attributes to `get_basename() <- write_webvtt_header()`:

```
==1== 16 bytes in 1 blocks are definitely lost in loss record 4 of 10
==1==    by 0x1A296A: get_basename (in /ccextractor)
==1==    by 0x1CC392: write_webvtt_header (in /ccextractor)
==1== 48 bytes in 3 blocks are definitely lost in loss record 6 of 10
==1==    by 0x1A296A: get_basename (in /ccextractor)
==1==    by 0x1CC392: write_webvtt_header (in /ccextractor)
```

**4 leaked blocks for 4 calls** — the leak count corroborates the re-entry count independently.

The remaining 32 bytes in 1 block is present on master too, is not in this code path, and is out of scope here.

## 3. No regression

13 formats (webvtt, srt, sami, ttxt, txt, scc, ssa, smptett, g608, mcc, spupng, simple_xml, rcwt): **13/13 identical** exit codes, output hashes and stdout hashes versus master.

`clang-format` clean.

PR #2342, opened 2026-09-09
[FIX] Segfault on --out=webvtt --timestamp-map: zero encoder_ctx so timing starts out NULL

### LINKED ISSUE #2341: [BUG] Segfault in write_webvtt_header(): --out=webvtt --timestamp-map crashes on RCWT/raw input (uninitialised encoder_ctx->timing)
CCExtractor version: master (907be052), built from source

# Necessary information

- Is this a regression (i.e. did it work before)? **Unknown** — the uninitialised field predates the `--timestamp-map` option, so it has likely never worked on these input paths.
- What platform did you use? **Linux** (Debian bookworm in Docker, x86-64). Platform-independent — nothing here is OS-specific.
- What were the used arguments? `--out=webvtt --timestamp-map`

# Video links

The reproducer is 153 bytes, so it is inline below rather than on a file host. Save it with:

```bash
EOF
```
sha256: `1e7e3f58151190124b113facc7543ac5ee6d7e230830cda88744de19cc17250c`

It is a hand-built RCWT file (per `docs/BINARY_FILE_FORMAT.TXT`) holding three EIA-608 CC1 pop-on captions at 1 s, 3 s and 5 s. It decodes correctly on master (`--out=srt` etc. all work), so it is a valid input, not a fuzz artefact. Happy to have it in the regression suite.

# Additional information

## Symptom

```
$ ccextractor sample.rcwt --out=webvtt --timestamp-map -o out.vtt
...
Analyzing data in CCExtractor's binary format
Segmentation fault
$ echo $?
139
```

Without `--timestamp-map` the same file processes fine (exit 0). With `--out=srt --timestamp-map` it is also fine. The crash needs **webvtt + --timestamp-map**.

## Backtrace

```
Program received signal SIGSEGV, Segmentation fault.
#0  write_webvtt_header ()
#1  write_cc_buffer_as_webvtt ()
#2  encode_sub ()
#3  rcwt_loop ()
#4  start_ccx ()
#5  main ()
```

## Root cause

`init_encoder()` allocates with `malloc` and never assigns `ctx->timing`:

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_common.c#L765

```c
struct encoder_ctx *ctx = malloc(sizeof(struct encoder_ctx));
```

`encoder_ctx.timing` is assigned in only four places, and none of them covers this path:

| Location | Condition |
|---|---|
| `general_loop.c:1534` (`general_loop`) | always |
| `general_loop.c:697` (`raw_loop`) | **only when `write_format == CCX_OF_MCC`** |
| `general_loop.c:1357` | DVB path |
| `mp4.c:892` | MP4 path |

So `rcwt_loop()`, `process_hex()`, and `raw_loop()` for every non-MCC output format leave `ctx->timing` holding indeterminate heap contents.

`write_webvtt_header()` then does:

https://github.com/CCExtractor/ccextractor/blob/master/src/lib_ccx/ccx_encoders_webvtt.c#L215

```c
if (ccx_options.timestamp_map && context->timing != NULL && context->timing->sync_pts2fts_set)
```

The `!= NULL` guard is the right check, but an indeterminate value is almost never `NULL`, so it passes and `->sync_pts2fts_set` dereferences a wild pointer. Reading an indeterminate value is undefined behaviour in the first place (CWE-457), and the dereference is CWE-824.

This also explains why it is `--timestamp-map`-specific: that flag is the only thing that makes `timing` be read at all on this path.

## Scope

`init_encoder()` assigns 60 of the struct's 74 members. Nine are never assigned:

`timing`, `list`, `dtvcc_writers`, `dvb_lang`, `cdp_hdr_seq`, `next_caption_time`, `force_dropframe`, `header_printed_flag`, `write_previous`

Some are filled in by callers, but any that are not are indeterminate. `timing` is the one currently reachable as a crash.

## Suggested fix

Allocate with `calloc` instead of `malloc`. One line, no behaviour change for the 60 explicitly-assigned members, and it makes the existing `!= NULL` guard work as intended — with `timing == NULL`, RCWT input simply skips `X-TIMESTAMP-MAP` (correct, since `sync_pts2fts_set` would be 0 anyway).

## Verification

Built master and a `calloc` build from identical source in the same Docker image:

- crash: `exit 139` → `exit 0`
- output across **13 formats** (webvtt, srt, sami, ttxt, txt, scc, ssa, smptett, g608, mcc, spupng, simple_xml, rcwt): byte-identical exit codes and SHA-256 hashes before and after
- `--timestamp-map` output now matches plain `--out=webvtt`, i.e. the header is correctly omitted rather than crashing

PR follows.

pranayr710:
Fixes #2341

Reason for this PR:

- [x] This PR fixes a bug that I have personally experienced and for which a sample exists (inline below, 153 bytes).

Sanity check:
- [x] Bug fix, so NOT added to the changelog.
- [x] Not adding new C code — this changes one allocation.

---

## The crash

```
$ ccextractor sample.rcwt --out=webvtt --timestamp-map -o out.vtt
Analyzing data in CCExtractor's binary format
Segmentation fault      (exit 139)
```

```
#0  write_webvtt_header ()
#1  write_cc_buffer_as_webvtt ()
#2  encode_sub ()
#3  rcwt_loop ()
#4  start_ccx ()
```

## Root cause

`init_encoder()` allocates with `malloc` and assigns 60 of `encoder_ctx`'s 74 members. `timing` is one of the nine it does not touch, and it is only set later in four places — `general_loop()`, the DVB path, `mp4.c`, and `raw_loop()` **but only when the output format is MCC**. So `rcwt_loop()`, `process_hex()` and `raw_loop()` for every other format leave it holding indeterminate heap contents.

`write_webvtt_header()` guards it correctly:

```c
if (ccx_options.timestamp_map && context->timing != NULL && context->timing->sync_pts2fts_set)
```

but an indeterminate value is almost never `NULL`, so the guard passes and `->sync_pts2fts_set` dereferences a wild pointer. `--timestamp-map` is the only thing that causes `timing` to be read on this path, which is why the crash is specific to that flag.

## The change

`malloc` → `calloc`. The 60 explicitly-assigned members are unaffected; `timing` becomes `NULL` and the existing guard does what it was written to do, so RCWT input skips `X-TIMESTAMP-MAP` rather than crashing — the correct outcome, since `sync_pts2fts_set` would be 0 there anyway.

I chose `calloc` over `ctx->timing = NULL;` because the same nine members are all indeterminate and reading any of them is UB; this closes the class for the same one-line cost. The other eight are `list`, `dtvcc_writers`, `dvb_lang`, `cdp_hdr_seq`, `next_caption_time`, `force_dropframe`, `header_printed_flag`, `write_previous`. **If you'd rather keep the blast radius to one field, say so and I'll switch it to the single assignment.**

## Repro

```bash
EOF
ccextractor sample.rcwt --out=webvtt --timestamp-map -o out.vtt ; echo $?   # 139 before, 0 after
```
sha256 `1e7e3f58151190124b113facc7543ac5ee6d7e230830cda88744de19cc17250c`. Hand-built RCWT per `docs/BINARY_FILE_FORMAT.TXT`: three EIA-608 CC1 pop-on captions at 1 s / 3 s / 5 s. It decodes correctly on master with every other format, so it is a valid input rather than a fuzz artefact — reuse it in the suite if it's useful.

## Verification

Built master and this change from identical source in the same Docker image (`docker/Dockerfile`, `BUILD_TYPE=minimal`):

- crash: `exit 139` → `exit 0`
- **13 formats** — webvtt, srt, sami, ttxt, txt, scc, ssa, smptett, g608, mcc, spupng, simple_xml, rcwt — identical exit codes **and** identical SHA-256 output hashes before and after
- `--timestamp-map` output is now byte-identical to plain `--out=webvtt`, i.e. the header is correctly omitted

`clang-format` clean.


cfsmp3:
@canihavesomecoffee  Can you add this sample to the collection?

